PR #47, opened 2014-12-18
viewers: simplify notify and remove dead code in ImageJ viewer

jlmuir:
The ImageJ viewer newUniqueIdCallback inner class had a comment saying
it would prefer to do the synchronized notify in the monitorChanged
method.  This commit does that and removes the newUniqueId() method
which is no longer used.

The ImageJ viewer had three places in its updateImage() method
where it would assign to a variable the return value of
img.getProcessor().getPixels() but then immediately assign the return
value of epicsGet{Byte,Short,Float}Array() to the same variable.  The
value assigned to the variable from the img.getProcessor().getPixels()
call is overwritten and never used, therefore the first call and
assignment does nothing and can be removed.

The ImageJ viewer had dead code in its readProperties() method where it
assigned a value to the path variable and then checked whether the path
variable was null.  Since the propertyFile variable is never null, the
path variable can never be null, hence the null check can be removed.


MarkRivers:
Done, thanks.

PR #49, opened 2015-01-06
File writer create directory feature

ulrikpedersen:
This is a feature which was implemented April 2014 at DLS by @NicholasRees.

The NDFileCreateDir parameter is used to automatically create directories if they don't exist. It works in 3 modes:
* If it is zero (default), no directories are created.
* If it is negative, then the absolute value is the maximum of directories that will be created
 * a value of -1 will create a maximum of one directory to complete the path, -2 will create a maximum of 2 directories.
* If it is positive, then at least that many directories in the path must exist
 * a value of 1 will create all directories below the root directory and 2 will not create a directory in the root directory.

This feature is independent of the TempSuffix feature in pull request #48 . However, the source code of these two pull requests live in the same functions (neighbouring lines) so care should be taken when merging to ensure they do not overwrite each other...


ulrikpedersen:
I have just had a go at merging this and #48 together into master on my local clone. There are a few (benign) conflicts with #48 as these both add new parameters in the same location in the source files. Although we want to keep both additions, git thinks it is a conflict and so it has to be resolved manually.

It is not a big deal, but to avoid confusion and make life bearable; I propose to first merge #48 and then merge master into this CreateDir branch, resolving the conflicts here. Afterwards, this Pull Request can be automatically merged into master.

PR #50, opened 2015-01-08
Autosave tse

bhill-slac:
Adds support for initialization and autosave of TSE field for records that update w/ each new image.

PR #53, opened 2015-01-10
Added support for BitsPerPixel and BytesPerPixel in ADBase

bhill-slac:
These PV's are useful for determining how to view the images, particularly via edm where different video widgets need to be used depending on these values.

### DIFF
--- a/ADApp/ADSrc/NDAttribute.cpp
+++ b/ADApp/ADSrc/NDAttribute.cpp
@@ -388,4 +388,25 @@ int NDAttribute::report(FILE *fp, int details)
   return ND_SUCCESS;
 }
 
+/** Get the max number of bits per pixel for the specified data type
+  * \param[in] NDDataType_t  Data type
+  */
+int GetNDDataTypeBits( NDDataType_t tyData )
+{
+    int nBits;
+    switch ( tyData )
+    {
+    // Default to large number so conversions to
+    // integer data types will be clipped to max
+    default:        nBits   = 99;   break;
+    case NDInt8:    nBits   = 8;    break;
+    case NDUInt8:   nBits   = 8;    break;
+    case NDInt16:   nBits   = 16;   break;
+    case NDUInt16:  nBits   = 16;   break;
+    case NDInt32:   nBits   = 32;   break;
+    case NDUInt32:  nBits   = 32;   break;
+    }
+    return nBits;
+}
+
 
--- a/ADApp/ADSrc/NDArray.cpp
+++ b/ADApp/ADSrc/NDArray.cpp
@@ -30,7 +30,7 @@
 NDArray::NDArray()
   : referenceCount(0), pNDArrayPool(NULL),  
     uniqueId(0), timeStamp(0.0), ndims(0), dataType(NDInt8),
-    dataSize(0),  pData(NULL)
+    dataSize(0), bitsPerElement(0), pData(NULL)
 {
   this->epicsTS.secPastEpoch = 0;
   this->epicsTS.nsec = 0;
@@ -148,6 +148,7 @@ int NDArray::getInfo(NDArrayInfo_t *pInfo)
     pInfo->ySize     = this->dims[pInfo->yDim].size;
     pInfo->colorSize = this->dims[pInfo->colorDim].size;
   }
+  pInfo->bitsPerElement = this->bitsPerElement;
   return(ND_SUCCESS);
 }
 
@@ -201,8 +202,8 @@ int NDArray::report(FILE *fp, int details)
     this->ndims);
   for (dim=0; dim<this->ndims; dim++) fprintf(fp, "%d ", (int)this->dims[dim].size);
   fprintf(fp, "]\n");
-  fprintf(fp, "  dataType=%d, dataSize=%d, pData=%p\n",
-        this->dataType, (int)this->dataSize, this->pData);
+  fprintf(fp, "  dataType=%d, nBitsPerPixel=%d, dataSize=%d, pData=%p\n",
+        this->dataType, this->bitsPerElement, (int)this->dataSize, this->pData);
   fprintf(fp, "  uniqueId=%d, timeStamp=%f\n",
         this->uniqueId, this->timeStamp);
   fprintf(fp, "  number of attributes=%d\n", this->pAttributeList->count());
--- a/ADApp/ADSrc/NDArray.h
+++ b/ADApp/ADSrc/NDArray.h
@@ -68,7 +68,8 @@ typedef struct NDDimension {
 /** Structure returned by NDArray::getInfo */
 typedef struct NDArrayInfo {
     size_t nElements;       /**< The total number of elements in the array */
-    int bytesPerElement;    /**< The number of bytes per element in the array */
+    int    bitsPerElement;  /**< The number of bits   per element in the array */
+    int    bytesPerElement; /**< The number of bytes per element in the array */
     size_t totalBytes;      /**< The total number of bytes required to hold the array;
                               *  this may be less than NDArray::dataSize. */
                             /**< The following are mostly useful for color images (RGB1, RGB2, RGB3) */
@@ -115,6 +116,7 @@ public:
     NDDataType_t  dataType;     /**< Data type for this array. */
     size_t        dataSize;     /**< Data size for this array; actual amount of memory allocated for *pData, may be more than
                                   * required to hold the array*/
+    int           bitsPerElement; /**< The number of bits   per element in the array */
     void          *pData;       /**< Pointer to the array data.
                                   * The data is assumed to be stored in the order of dims[0] changing fastest, and 
                                   * dims[ndims-1] changing slowest. */
--- a/ADApp/ADSrc/NDArrayPool.cpp
+++ b/ADApp/ADSrc/NDArrayPool.cpp
@@ -89,6 +89,7 @@ NDArray* NDArrayPool::alloc(int ndims, size_t *dims, NDDataType_t dataType, size
     /* Initialize fields */
     pArray->pNDArrayPool = this;
     pArray->dataType = dataType;
+    pArray->bitsPerElement = GetNDDataTypeBits(dataType);
     pArray->ndims = ndims;
     memset(pArray->dims, 0, sizeof(pArray->dims));
     for (i=0; i<ndims && i<ND_ARRAY_MAX_DIMS; i++) {
@@ -191,6 +192,7 @@ NDArray* NDArrayPool::copy(NDArray *pIn, NDArray *pOut, int copyData)
     pOut = this->alloc(pIn->ndims, dimSizeOut, pIn->dataType, 0, NULL);
     if(NULL==pOut) return NULL;
   }
+  pOut->bitsPerElement = pIn->bitsPerElement;
   pOut->uniqueId = pIn->uniqueId;
   pOut->timeStamp = pIn->timeStamp;
   pOut->epicsTS = pIn->epicsTS;
@@ -479,7 +481,6 @@ int NDArrayPool::convert(NDArray *pIn,
   NDDimension_t dimsOutCopy[ND_ARRAY_MAX_DIMS];
   int i;
   NDArray *pOut;
-  NDArrayInfo_t arrayInfo;
   NDAttribute *pAttribute;
   int colorMode, colorModeMono = NDColorModeMono;
   const char *functionName = "convert";
@@ -516,20 +517,25 @@ int NDArrayPool::convert(NDArray *pIn,
     return(ND_ERROR);
   }
   /* Copy fields from input to output */
-  pOut->timeStamp = pIn->timeStamp;
-  pOut->epicsTS = pIn->epicsTS;
-  pOut->uniqueId = pIn->uniqueId;
+  pOut->bitsPerElement  = pIn->bitsPerElement;
+  if( pOut->bitsPerElement > GetNDDataTypeBits(pOut->dataType) )
+      pOut->bitsPerElement = GetNDDataTypeBits(pOut->dataType);
+  pOut->timeStamp       = pIn->timeStamp;
+  pOut->epicsTS         = pIn->epicsTS;
+  pOut->uniqueId        = pIn->uniqueId;
   /* Replace the dimensions with those passed to this function */
   memcpy(pOut->dims, dimsOutCopy, pIn->ndims*sizeof(NDDimension_t));
   pIn->pAttributeList->copy(pOut->pAttributeList);
 
-  pOut->getInfo(&arrayInfo);
-
+  NDArrayInfo_t arrayInfoIn;
+  NDArrayInfo_t arrayInfoOut;
+  pIn->getInfo( &arrayInfoIn );
+  pOut->getInfo(&arrayInfoOut);
   if (dimsUnchanged) {
     if (pIn->dataType == pOut->dataType) {
       /* The dimensions are the same and the data type is the same,
        * then just copy the input image to the output image */
-      memcpy(pOut->pData, pIn->pData, arrayInfo.totalBytes);
+      memcpy(pOut->pData, pIn->pData, arrayInfoOut.totalBytes);
       return ND_SUCCESS;
     } else {
       /* We need to convert data types */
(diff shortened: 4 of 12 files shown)


bhill-slac:
I realized I hadn't added the BitsPerPixel and BytesPerPixel parameters to the html docs.
I've done that now in my svn version, but am not sure the best way to submit them as this pull request isn't from my fork, and I don't have any way to create a clean branch on my fork till these pending pull requests are resolved.

Here's a patch that applies them:
Index: documentation/pluginDoc.html
===================================================================
--- documentation/pluginDoc.html        (revision 20310)
+++ documentation/pluginDoc.html        (working copy)
@@ -573,6 +573,42 @@
       </tr>
       <tr>
         <td>
+          NDBitsPerPixel</td>
+        <td>
+          asynInt32</td>
+        <td>
+          r/o</td>
+        <td>
+          Number of bits per pixel.
+                 Propagated by plugins if defined by driver.
+               </td>
+        <td>
+          BITS_PER_PIXEL</td>
+        <td>
+          $(P)$(R)BitsPerPixel_RBV<br />
+        <td>
+          longin</td>
+      </tr>
+      <tr>
+        <td>
+          NDBytesPerPixel</td>
+        <td>
+          asynInt32</td>
+        <td>
+          r/o</td>
+        <td>
+          Number of bytes per pixel.
+                 Propagated by plugins if defined by driver.
+               </td>
+        <td>
+          BYTES_PER_PIXEL</td>
+        <td>
+          $(P)$(R)BytesPerPixel_RBV<br />
+        <td>
+          longin</td>
+      </tr>
+      <tr>
+        <td>
           NDPluginDriver<br />
           DataType</td>
         <td>
Index: documentation/areaDetectorDoc.html
===================================================================
--- documentation/areaDetectorDoc.html  (revision 20310)
+++ documentation/areaDetectorDoc.html  (working copy)
@@ -1265,6 +1265,42 @@
       </tr>
       <tr>
         <td>
+          NDBitsPerPixel</td>
+        <td>
+          asynInt32</td>
+        <td>
+          r/o</td>
+        <td>
+          Number of bits per pixel.
+                 Only defined if set by driver.
+               </td>
+        <td>
+          BITS_PER_PIXEL</td>
+        <td>
+          $(P)$(R)BitsPerPixel_RBV<br />
+        <td>
+          longin</td>
+      </tr>
+      <tr>
+        <td>
+          NDBytesPerPixel</td>
+        <td>
+          asynInt32</td>
+        <td>
+          r/o</td>
+        <td>
+          Number of bytes per pixel.
+                 Only defined if set by driver.
+               </td>
+        <td>
+          BYTES_PER_PIXEL</td>
+        <td>
+          $(P)$(R)BytesPerPixel_RBV<br />
+        <td>
+          longin</td>
+      </tr>
+      <tr>
+        <td>
           ADTemperature</td>
         <td>
           asynFloat64</td>

ulrikpedersen:
I'm reviewing the changes now and I have a few comments and questions to discuss.

New asynNDArrayDriver parameters
----------------------------------------------
This PR adds two new parameters NDBitsPerPixel and NDBytesPerPixel. The data is stored as members of the NDArray class - and NDArrayPool sets the values based on the NDArray datatype. So it strikes me as another way to describe the datatype - is this correct? If true, this seems a bit redundant and extra work to maintain several parameters describing the same thing: datatype, bits and bytes per pixel.

Special datatype cases
-----------------------------
I can see how this would work in the basic case: a regular 8bit image has of course 1 byte per pixel and 8 bits. But we quickly get to a few more complex cases, where at least the bits per pixel could be useful, but should behave differently than your implementation:
* RGB images. This is an 8 bit image, but with 3 bytes (12bits) representing each pixel, by adding a 3rd dimension to a 2D image.
* 16 bit (mono) images with only 12 bits of data. Several firewire and GigE cameras have such a mode (possibly PCO as well).

Of course bitsPerElement should tell the truth - is it up to the driver to set this value if it cannot be directly deferred from the datatype?

Adding default values to TSE fields
--------------------------------------------
I am not entirely familiar with the TSE field, but I understand it relates to the timestamping of the record - and potentially has some relation to an external timing system. The use of timing systems vary wildly from site to site - will there be a side effect to setting TSE=-2 on sites/users which don't use timing systems with their areaDetector IOCs?

Autosave info fields
-------------------------
We have actually been discussing this in the areaDetector working group for a while and I believe some tags where already added. Perhaps @mp49 could take a look at these proposed tags and see if they agree with our discussions and his latest work on the same?

ADApp/ADSrc/NDAttribute.cpp
```diff
@@ -388,4 +388,25 @@ int NDAttribute::report(FILE *fp, int details)
...
+    // Default to large number so conversions to
+    // integer data types will be clipped to max
+    default:        nBits   = 99;   break;
+    case NDInt8:    nBits   = 8;    break;
+    case NDUInt8:   nBits   = 8;    break;
+    case NDInt16:   nBits   = 16;   break;
+    case NDUInt16:  nBits   = 16;   break;
+    case NDInt32:   nBits   = 32;   break;
+    case NDUInt32:  nBits   = 32;   break;
+    }
```
ulrikpedersen: We really must support the floating point datatypes as well...

I think it would be better to return an error code <0 as default - **and** add support for the floating point data types.

MarkRivers:
I understand the need for bitPerPixel.  I might have a 16-bit data type but the image data is only 12-bits, and some clients might want to know that.  However, I don't understand the need for BytesPerPixel.  That seems redundant with the NDDataType and bitsPerPixel.  We don't want to add redundant fields.  Are these fields intended for channel access clients, or for other plugins?  If they are intended for channel access clients then you need to remember that there is another data type involved, namely FTVL of the waveform record being used to transport the images.  What is NDDataType=NDUInt16, bitsPerPixel is 12, bytesPerPixel=2, and FTVL=LONG.  In this case the number of bytes per pixel in the waveform record is really 4, not 2.

MarkRivers:
There are plugins which produce new arrays but might or might not change the values of bitsPerPixel.  For example the NDPluginProcess plugin.  If this plugin was receiving NDUInt16 arrays with bitsPerPixel=12, and was performing only high-clipping, it would not be modifying bitsPerPixel.  But if it was performing scaling, which might then fill the entire 16-bit range, then it would be modifying bitsPerPixel.  Similarly the ROI plugin might be changing bitsPerPixel if is is doing binning, but might not if it is also doing scaling.  The current implementation appears to set bitsPerPixel to match the NDDataType in such cases.  But then that is just redundant information and bitsPerPixel has no added value.

mp49:
The info fields look fine.

However, shouldn't the default TSE be zero? I think setting TSE=-2 will change behavior won't it? It's been a while since I looked into that, so I'm not sure.

ADApp/ADSrc/NDAttribute.cpp
bhill-slac: GetNDDataTypeBits() is used to determine the default value for newly alloced NDArray's and for clipping in NDArrayPool::convert when converting to a smaller data type.   I agree it would be good to return -1 for invalid and specific values for NDFloat32 and NDFloat64.   I'm also open to returning a different value than 99 as long as it's large.   I don't think bitsPerPixel is particularly relevant for floating point NDArrays.

For color images, my searches seemed to indicate that discussions re the number of bits refer to the number of bits per pixel color, so 8 for typical RGB and 16 for RGB16.   I'm not aware of any color image standards that use anything but 8 or 16 except for digital video where you may see more bits for luminance, for example, 6 bits G, 5 bits R and 5 bits B.   For color, the
bitsPerPixel parameter is of less interest as the color mode tells you all you need to know.

Where bitsPerPixel is needed is for mono images, where there is no standard.   We have mono cameras with 8, 10, 12, and 16 bits and sometimes use the colorConvert plugin to reduce waveform size by converting to Uint8 images.

To be useful, BitsPerPixel must be set by the driver.   I've done this in one of my ADProsilica pull requests and in the ADEdtPdv driver I'll soon be posting for EDT framegrabber based cameras.

bhill-slac:
I'm OK with closing both this pull request and #50 as I can't do much with my current ADCore fork as long as they're open.   If you close them I'll delete my fork and resubmit any work you're interested in from a clean fork.

Re BitsPerPixel, I think it's useful and can resubmit w/ changes as discussed above.

Re BytesPerPixel, I added it thinking it would be useful but have found that BitsPerPixel is all I'm using so far in my viewers, so I'm OK w/ removing it.

Sounds like most of the autosave info changes are useful so I can do a new pull request for them.

Re the TSE default value, my understanding is that before registerTimeStampSource was recently added to asyn, it used epicsTimeGetCurrent() to set the timestamps, and if you don't register a TimeStampSource the timestamps still default to epicsTimeGetCurrent.

If you do register a TimeStampSource in your areaDetector driver, it's used to set the timestamp based on when the image was acquired.   For this use case I think the default should be to use
the TimeStampSource provided timestamp for those PV's that update once for each new image.

From this perspective, the behavior change would be limited to those sites which have added TimeStampSource support but didn't set TSE to -2 for those PV's.    I'm comfortable with defaulting those TSE fields to -2 at our site, but can submit w/ TSE defaulting to 0 if preferred.

ulrikpedersen:
git forks and pull requests
---------------------------------
Actually you can do anything you like to your own fork. It will not affect this PR as it is issued from my fork (ulrikpedersen:cherries) after I cherry-picked your two commits. You also can't make changes to the code in this PR for the same reason: it's on my fork.

So you can go ahead and reset (delete) your fork and create new branches if you want to submit some slightly different and more specific PRs than this one (sounds like that's where its heading :-))

The bitsPerPixel parameter
-----------------------------------
After thinking a bit more about it I think that the bitsPerPixel parameter could be quite useful for a number of things:
* Cameras which only use a subset of the bits available in a word.
* Plugins which produce NDArrays from calculations.
* It may be useful in writing compressed files (I'm thinking of a future extension to the HDF5 writer here).

I'm still not sure how to support the "RGB" case...

The bytesPerPixel is really not required as it does not have the potential to yield any new information (dataType covers it). 

I agree with Mark that it is not a good idea for the ROI and Process plugins to simply set the bitsPerPixels based on dataType - because adding just one of these plugins in the process chain before your viewer will circumvent the whole idea of having the field. These plugins should really calculate a true value for bitsPerPixel when the output of a calculation requires more bits (and clipping at the maximum allowed bitsPerPixel based on the output datatype). This feature would really be a big help for users, playing around with gains and calculations. They they often don't understand why or what plugin cause pixels to overflow and consequently waste a lot of time with trial-and-error configurations.

Floating point types must be supported as well. I guess that they can always be set to the maximum number of bits (32, 64), right?

TSE field
------------
Regarding the TSE field. Instead of specifying TSE=0, just don't specify the TSE field at all as it defaults to 0 anyway. Less lines of code to maintain :-) Here at our site we occasionally want a different default value on certain fields and we can do that by either doing dbpf in the startup script or more elegantly by creating another database (or template) with the same record definitions, but only specifying the fields that we want to change. This works if you load the database **after** the areaDetector ones.

MarkRivers:
> Re the TSE default value, my understanding is that before registerTimeStampSource was recently added to asyn,

> it used epicsTimeGetCurrent() to set the timestamps, and if you don't register a TimeStampSource the

> timestamps still default to epicsTimeGetCurrent.

> If you do register a TimeStampSource in your areaDetector driver, it's used to set the timestamp based on when

> the image was acquired.


> From this perspective, the behavior change would be limited to those sites which have added TimeStampSource

> support but didn't set TSE to -2 for those PV's. I'm comfortable with defaulting those TSE fields to -2 at our site,

> but can submit w/ TSE defaulting to 0 if preferred.


I don't think this analysis is correct.  If TSE=0 then the timestamp is computed at the time the record processes.  If TSE=-2 then the timestamp is assumed to have been set by device support.  Thus, even if one does not register a timestamp source the behavior would change if the TSE default is changed to -2.  Previously the time would have been when the corresponding record processed, but now it would be the time that the driver set in the pasynUser structure, whether that time came from epicsTimeGetCurrent() or a custom time source.


Am I wrong?


Mark


ADApp/ADSrc/NDAttribute.cpp
MarkRivers: > I'm not aware of any color image standards that use anything but 8 or 16 except for digital video where you may see more bits for

> luminance, for example, 6 bits G, 5 bits R and 5 bits B. For color, the bitsPerPixel parameter is of less interest as the color mode tells you all you need to know.

There may well be cameras that produce 16-bit RGB, but for which it only uses the low-order 12 bits for example.

There are also the YUV color modes that are supported in NDArray.h.  They can encode 2 RGB pixels in 4 bytes for example, or 4 pixels in 6 bytes.
typedef enum
{
    NDColorModeMono,    /**< Monochromatic image */
    NDColorModeBayer,   /**< Bayer pattern image, 1 value per pixel but with color filter on detector */
    NDColorModeRGB1,    /**< RGB image with pixel color interleave, data array is [3, NX, NY] */
    NDColorModeRGB2,    /**< RGB image with row color interleave, data array is [NX, 3, NY]  */
    NDColorModeRGB3,    /**< RGB image with plane color interleave, data array is [NX, NY, 3]  */
    NDColorModeYUV444,  /**< YUV image, 3 bytes encodes 1 RGB pixel */
    NDColorModeYUV422,  /**< YUV image, 4 bytes encodes 2 RGB pixel */
    NDColorModeYUV411   /**< YUV image, 6 bytes encodes 4 RGB pixels */
} NDColorMode_t;

Mark

bhill-slac:
Re BitsPerPixel
The pull request only uses the dataType to set bitsPerPixel on newly alloc'd NDArrays.
NDArray::copy sets it based on the input array.
NDArray::convert uses the input array but clips based on max number of bits in new type.
I agree that the ROI and Process plugins should consider scaling and binning factors and meant
to address those, but haven't gotten to that yet.    Should I pend this commit till fully supported?

Re TSE
I see Mark's point re the behavior difference.   On closer examination, the port::timeStamp field in asyManager.c only gets updated in response to a call to updateTimeStamp() from the driver, so even if you're using the default epicsTimeGetCurrent(), it only gets called once instead of each time asynPortDriver::getTimeStamp() is called.

It's a subtle difference as the PV's in question are all I/O Intr and processed once per image.
I could argue that the image timestamp is far more interesting than the process time, but I suppose that could vary for other sites.

I tried to figure out what happens if the driver doesn't call updateTimeStamp().   The AravisGigE driver on github doesn't appear to do so yet, and maybe others as well.    This may be a big drawback for TSE=-2, but I may need to try disabling the updateTimeStamp() code to see what happens.

bhill-slac:
More thoughts re bitsPerPixel:
My only reason originally for adding this was to allow viewers to determine how to convert mono pixel values appropriately for display including false color mappings.    From that perspective it allows one to compute the range between min and max luminance values.

Doing some more reading about RGB and YUV color standards I see that there are indeed many variations on the number of bits per component, including many w/ different numbers of bits per component.    When these articles referred to the bits per pixel, it was in regards to the total number of bits used to represent the RGB pixel.  As Mark pointed out, the YUV formats go a step further with 1, 2, or 4 RGB pixels packed using sub-sampling into a mega-pixel.

For each color standard, a fixed number of bits is used to represent each pixel or mega-pixel group for YUV standards.    The range of values for each component is thus determined by the standard.

Re floating point, I'm not sure who uses that.   Our color standards are all packed bit representations, so it would seem floating point would only be used for mono.   We can set them to the number of bits used to store the floating point value, 32 or 64, but the range is effectively infinite from the perspective of mapping pixel values into a color space.

So, how's this sound for a way forward re the bitsPerPixel value:
* Mono integer types:   Set by driver, propagated by NDArray::copy, add code to ROI and Process plugins to scale appropriately based on binning and scaling settings.
* Mono floating point:    32 for float, 64 for double
* Color standards:   Set to the fixed number of bits used to represent one pixel or group of pixels

If this sounds OK, I could wait to resubmit this till I strip out the unneeded bytesPerPixel code and implement the above including code for binning and scaling.

ulrikpedersen:
> So, how's this sound for a way forward re the bitsPerPixel value:
>
>   * Mono integer types: Set by driver, propagated by NDArray::copy, add code to ROI and Process plugins to scale appropriately based on binning and scaling settings.
>   * Mono floating point: 32 for float, 64 for double
>   * Color standards: Set to the fixed number of bits used to represent one pixel or group of pixels


Thanks Bruce for summarising this so clearly. Good plan - and I'm happy with the bitsPerPixel functionality as described.

Feel free to close this Pull Request if you're going to make another one.

ulrikpedersen:
Closing this as it's been hanging around for a while without any activity. The discussion on this PR lead to a conclusion that another implementation would be more appropriate...

Issue #65, opened 2015-03-02
NDFileHDF5: Attributes defined for datasets in the xml file do not always appear in the HDF5 file.


MarkRivers:
This problem is seen in the test2 files.  

test2_layout.xml contains the following lines:

            <dataset name="aequatorial_angle" source="ndattribute" ndattribute="sa">
                <attribute name="units"       source="constant"    value="degrees" type="string" />
            </dataset>

However, the HDF5 file contains the following (this is with recent changes to "automatic" attributes in the hdf5-bugfix branch)

         DATASET "aequatorial_angle" {
            DATATYPE  H5T_IEEE_F64LE
            DATASPACE  SIMPLE { ( 5 ) / ( 5 ) }
            DATA {
            (0): 0.001, 0, 0, 0, 0
            }
            ATTRIBUTE "NDAttrDescription" {
               DATATYPE  H5T_STRING {
                     STRSIZE 34;
                     STRPAD H5T_STR_NULLTERM;
                     CSET H5T_CSET_ASCII;
                     CTYPE H5T_C_S1;
                  }
               DATASPACE  SCALAR
               DATA {
               (0): "sample azimuthal rotation, degrees"
               }
            }
            ATTRIBUTE "NDAttrName" {
               DATATYPE  H5T_STRING {
                     STRSIZE 2;
                     STRPAD H5T_STR_NULLTERM;
                     CSET H5T_CSET_ASCII;
                     CTYPE H5T_C_S1;
                  }
               DATASPACE  SCALAR
               DATA {
               (0): "sa"
               }
            }
            ATTRIBUTE "NDAttrSource" {
               DATATYPE  H5T_STRING {
                     STRSIZE 8;
                     STRPAD H5T_STR_NULLTERM;
                     CSET H5T_CSET_ASCII;
                     CTYPE H5T_C_S1;
                  }
               DATASPACE  SCALAR
               DATA {
               (0): "ACQ_TIME"
               }
            }
            ATTRIBUTE "NDAttrSourceType" {
               DATATYPE  H5T_STRING {
                     STRSIZE 17;
                     STRPAD H5T_STR_NULLTERM;
                     CSET H5T_CSET_ASCII;
                     CTYPE H5T_C_S1;
                  }
               DATASPACE  SCALAR
               DATA {
               (0): "NDAttrSourceParam"
               }
            }
         }
         DATASET "thickness" {

[...]

Issue #66, opened 2015-03-02
NDFileHDF5: If the xml layout file does not define a group with “ndattr_default=true” it crashes the application.


MarkRivers:
If the NDFileHDF5 plugin is using a user-specific XML layout file, and if that file does not define a group with the ndattr_default=true attribute then it will crash the IOC application.

aglowacki:
What should the writer do if this field is missing. I believe I found the two spots in the code that we need to check for a NULL pointer. I can return an asynError but then these attributes are not saved to the hdf5 file. Should we put them in the root of the hdf5 if it doesn't exist?

prjemian:
Tells me that the layout author did not want them in the HDF5 file.

ulrikpedersen:
The trouble is when a user simply forgets to tag a group with ndattr_default=true and ends up missing a bunch of NDAttributes from the file... Do we want to support this unfortunate user (who don't know about or forgets setting ndattr_default=true) or the experienced user who deliberately leaves our ndattr_default=true because she does not want to store all NDAttributes?

I personally tend to agree with @prjemian on this one: if the user did not specify a ndattr_default group then it's probably because they're not required to be stored. It's a corner-case of course, because why would you define the NDAttributes in the first place if you don't wish to store them?

MarkRivers:
I would argue that this should always be created.  Users are much less happy with missing data in their files than with extra data.  If it is decided that the new datasets for UniqueID, timeStamp, etc. (see issue #68) should go in this group then I would strongly argue that is must always be created even if it was not present in the XML file.

prjemian:
One way to resolve this is to make writing of this group be optional 
with a default decision to always write.  An optional attribute on the 
root element of the layout XML file could be added to disable the 
automatic write if that is truly intended.  Here's an example:

<hdf5_layout auto_ndattr_default="false">
...
</hdf5_layout>

ulrikpedersen:
@prjemian yes that auto_ndattr_default=false would solve it: by default storing the NDAttributes in the root unless the user specifically disable this behaviour - or has tagged another group with the ndattr_default=true flag.

ulrikpedersen:
In response to @aglowacki's email:

> As I understand it we want the logic to be
> 
> 1) ndattr_default=true > // it saves the default to this group
> 2) if no ndattr_default, // save to root group
> 3) if new tag hdf5_layout auto_ndattr_default=“false // we don’t save default attr anywhere
> 
> Is this correct?

[...]

PR #71, opened 2015-03-04
Hdf5 bugfix

aglowacki:
This update fixes the crash for issue #66 “ndattr_default=true” is not defined. The current logic is that if it is not defined it will not save them anywhere. 

This partly fixes issue #65. It should now save attributes under constant datasets. It does not save attributes under ndattribute datasets.

### DIFF
--- a/ADApp/pluginSrc/NDFileHDF5.cpp
+++ b/ADApp/pluginSrc/NDFileHDF5.cpp
@@ -394,21 +394,25 @@ asynStatus NDFileHDF5::createTree(hdf5::Group* root, hid_t h5handle)
     // Write contant datasets to file
 //    this->writeHdfConstDatasets(new_group, root);
 
-    getIntegerParam(NDFileHDF5_storeAttributes, &storeAttributes);
-    if (storeAttributes == 1){
-      // Set some attributes on the group
-      this->writeHdfAttributes(new_group,  root);
-    }
-
     // Create all the datasets in this group
     hdf5::Group::MapDatasets_t::iterator it_dsets;
     hdf5::Group::MapDatasets_t& datasets = root->get_datasets();
     for (it_dsets = datasets.begin(); it_dsets != datasets.end(); ++it_dsets){
       hid_t new_dset = this->createDataset(new_group, it_dsets->second);
-      if (new_dset <= 0) continue; // failure to create the datasets so move on to next
+      if (new_dset <= 0) {
+//hdf5::Dataset *dset = it_dsets->second;
+//asynPrint(this->pasynUserSelf, ASYN_TRACE_ERROR, "%s::%s Failed to create dataset: %s\n",  driverName, functionName, dset->get_name().c_str());
+        continue; // failure to create the datasets so move on to next
+      }
       // Write the hdf attributes to the dataset
       this->writeHdfAttributes( new_dset,  it_dsets->second);
       // Datasets are closed after data has been written
+    }
+
+    getIntegerParam(NDFileHDF5_storeAttributes, &storeAttributes);
+    if (storeAttributes == 1){
+      // Set some attributes on the group
+      this->writeHdfAttributes(new_group,  root);
     }
 
     hdf5::Group::MapGroups_t::const_iterator it_group;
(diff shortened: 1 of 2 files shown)


ADApp/pluginSrc/NDFileHDF5.cpp
```diff
@@ -392,25 +392,29 @@ asynStatus NDFileHDF5::createTree(hdf5::Group* root, hid_t h5handle)
...
 
     // Write contant datasets to file
-    this->writeHdfConstDatasets(new_group, root);
-
-    getIntegerParam(NDFileHDF5_storeAttributes, &storeAttributes);
-    if (storeAttributes == 1){
-      // Set some attributes on the group
-      this->writeHdfAttributes(new_group,  root);
-    }
+//    this->writeHdfConstDatasets(new_group, root);
```
ulrikpedersen: With writeHdfConstDatasets() commented out at this point, this function is not used anywhere else. Is it no longer required and can be deleted? It should not be left in the code like this if it is not used...

I must admit I've lost track a bit in these many changes so I don't have the overview to determine if this function can go...

MarkRivers:
I just tested this fix.  There were several problems:

- It does not compile on Linux, and had a compiler warning due to no return value from function.  Please test that code compiles without error or warning on both Linux and Windows.  I have fixed the problems and pushed to github.

- When testing with test2_attributes.xml and test2_layout_no_ndattr_default.xml there are HDF5 error messages:
HDF5-DIAG: Error detected in HDF5 (1.8.7) thread 0:
  #000: H5G.c line 734 in H5Gclose(): not a group
    major: Invalid arguments to routine
    minor: Inappropriate type
When testing with test2_layout.xml there are no errors, so this must be related to the lack of the nd_attr_default tag.  So it no longer crashes, but it does produce HDF5 error messages.

I verified that attributes of constant datasets are now stored correctly.

aglowacki:
That is strange because I compiled on linux and didn't get any errors.

The writeHdfConstDatasets() function is not needed anymore, I forgot to remove it.

MarkRivers:
What version of asyn are you using on Linux?  In asyn R4-21 I added this:


************************

Changed the asynTrace print, printIO, vprint, vprintIO functions so they use EPICS_PRINTF_STYLE. This causes the GCC (version 3.0 and higher) and clang compilers to check the agreement of format strings with function arguments when using asynPrint() and asynPrintIO(), just as they do with printf(). This is very helpful in finding errors, and uncovered a number in asyn itself, which have been fixed.

************************

The error and fix were the following:
             asynPrint(this->pasynUserSelf, ASYN_TRACE_ERROR, "%s::writeHdfAttributes unknown type: unable to create attribute: %s\n",
-              driverName, attr.get_name());
+              driverName, attr.get_name().c_str());


So you were using something that was not a char* for a %s format.  gcc on Linux should have signaled that error.



aglowacki:
I have asyn-4-22 and didn't get the error when compiling for linux-x86_64-debug

MarkRivers:
I don't understand why you are not getting an error.  This is what I get with commit e95bd3e867b8293940d813d2c000fbbf8507414c of NDFileHDF5.cpp, gcc version 4.6.3, asyn R4-26.

corvette:ADCore/ADApp/pluginSrc>make
make -C O.linux-x86_64 -f ../Makefile TOP=../../.. T_A=linux-x86_64 install
make[1]: Entering directory `/home/epics/devel/areaDetector-2-1/ADCore/ADApp/pluginSrc/O.linux-x86_64'

/usr/bin/g++ -c  -D_POSIX_C_SOURCE=199506L -D_POSIX_THREADS -D_XOPEN_SOURCE=500           -D_X86_64_  -DUNIX  -D_DEFAULT_SOURCE -Dlinux  -D_REENTRANT   -O3 -g   -Wall      -m64     -fPIC -MMD -I. -I../O.Common -I. -I.. -I../../../include/os/Linux -I../../../include     -I/corvette/usr/local/epics/base-3.14.12.4/include/os/Linux -I/corvette/usr/local/epics/base-3.14.12.4/include  -I/corvette/home/epics/devel/asyn-4-26/include  -I/corvette/home/epics/devel/areaDetector-2-1/ADBinaries/include  -I/corvette/home/epics/devel/areaDetector-2-1/ADCore/include     -I/usr/local64/include  -I/usr/include/libxml2   ../NDFileHDF5.cpp
../NDFileHDF5.cpp: In member function 'void NDFileHDF5::writeHdfAttributes(hid_t, hdf5::Element*)':
../NDFileHDF5.cpp:510:13: error: cannot pass objects of non-trivially-copyable type 'std::string {aka struct std::basic_string<char>}' through '...'
../NDFileHDF5.cpp:510:13: warning: format '%s' expects argument of type 'char*', but argument 7 has type 'std::string {aka std::basic_string<char>}' [-Wformat]
../NDFileHDF5.cpp: In member function 'hid_t NDFileHDF5::writeHdfConstDataset(hid_t, hdf5::Dataset*)':
../NDFileHDF5.cpp:577:1: warning: control reaches end of non-void function [-Wreturn-type]
make[1]: *** [NDFileHDF5.o] Error 1
make[1]: Leaving directory `/home/epics/devel/areaDetector-2-1/ADCore/ADApp/pluginSrc/O.linux-x86_64'
make: *** [install.linux-x86_64] Error 2

PR #77, opened 2015-03-09
Db includes

MarkRivers:
This change simplifies and improves the handling of template files.

Previously there were many parameters that are defined in asynNDArrayDriver.template that had duplicate records in ADBase.template and NDPluginBase.template.  These common parameters have been moved to a new file, NDArrayBase.template.  This file defines all parameters from asynNDArrayDriver except those in NDFile.template.

ADDriver.template and NDPluginBase.template now have the line "include NDArrayBase.template", so they include the common records.

All plugin template files now contain the line "include NDPluginBase.template".  This means that startup scripts no longer need to include both NDPluginBase.template and the specific plugin template file, it is done automatically.

All file writing plugins now also contain the line "include NDFile.template", which means it is no longer necessary to load that file in the startup script.

Driver template files should now include the include "include ADBase.template", removing the need to load ADBase.template separately.

To simplify finding the include files, and indeed finding all template files the environment variables EPICS_DB_INCLUDE_PATH must now be defined in the startup script.  In the simDetector IOC this is defined to be $(ADCORE)/db".  This allowed removing the $(ADCORE)/db/ path from all dbLoadRecords commands.

Added a new file ADCoreVersion.h which defines ADCORE_RELEASE, ADCORE_REVISION, and ADCORE_MODIFICATION.  These are available for drivers and plugins that need to modify their behavior depending on the version of ADCore.

Created a new record in NDArrayBase.template, ADCoreVersion_RBV.  This record contains a string derived from the ADCoreVersion.h.  For ADCore R2-2 it will be the string "2.2.0".  This record will allow channel access clients to modify their behavior depending on the version of ADCore,


ulrikpedersen:
This looks like a substantial set of changes. Do you want to have all of this included in ADCore release 2-2?

MarkRivers:
Yes, I'd like to include it in 2-2.  While quite a few files have changed, the changes are pretty simple and straightforward.  The template files now have the same class hierarchy as the C++ code.  The top-level template file now "includes" the next level below, etc. down to the level of the new NDArrayBase.template file.  This eliminates the duplicate records that were previously present in ADBase.template and NDPluginDriver.template, which eases maintenance.  If the simDetector IOC boots without errors (which it does) then one can have high confidence that it is working OK.

I'd like you to look it over to make sure it will work with however DLS loads template files.

I also added the new ADCoreVersion record so that Channel Access clients can deal with future changes in the API, for example when we change NDPluginFile to eliminate the FileWriteMode and AutoSave records.

Mark

thomascobb:
This sounds like a good idea, anything that simplified database creation is a good thing! There will be a number of changes we will need to make internally to support this, but ultimately it should make it easier for our iocbuilder support too. The only thing that makes things difficult for us is if the end user ever needs to instantiate a bit of database that is also used via an include statement in another file. This is not done in this pull request so it not a problem here.

Tom

ulrikpedersen:
I trust Tom with his vast db experience :-) so I will have a go at merging and building it tomorrow.

ulrikpedersen:
Hmmm... Strangely my merge-comment does not appear in the history of this PR, so I'm just adding the comment again:

I have tested this in the DLS environment where we use MSI to expand template files into databases at build-time rather than at startup. It appears to work just fine. My tests was done by modifying the simDetectorIOC slightly. Just for for information; I have pushed the test up to my own fork on the [db-includes-msitest](https://github.com/ulrikpedersen/ADCore/tree/db-includes-msitest) branch.

I also added a commit bc51612 to this PR where I add the template filename at the top of every template. That makes it a bit easier to track where a record belongs to when looking at the expanded database.

PR #81, opened 2015-03-11
Update NDAttribute datasets to increase in dimension as required Fix #74

### LINKED ISSUE #74: HDF5: datasets that get their data from NDAttributes are initially created with the size of numArrays.
datasets that get their data from NDAttributes are initially created with the size of numArrays.  This is not correct, they should not be statically allocated they should grow in an unlimited dimension as new NDArrays arrive, just like the detector data does.  The user should be able to set numArrays to a very large value and then stop after the desired time without causing the dataset to be unnecessarily large.  This is how the netCDF plugin works.  This may also why it crashes when setting numArrays to 0.

ajgdls:
Introduced a new parameter NDAttributeChunk for setting chunking.  Updated
NDAttribute datasets to use unlimited maximum length and then increase the
size as required.  If the user stops the collection early then only
collected data is stored to the datasets.
This resolves issue #74.

### DIFF
--- a/ADApp/Db/NDFileHDF5.template
+++ b/ADApp/Db/NDFileHDF5.template
@@ -82,6 +82,22 @@ record(longin, "$(P)$(R)NumFramesChunks_RBV")
     field(SCAN, "I/O Intr")
 }
 
+record(longout, "$(P)$(R)NDAttributeChunk")
+{
+    field(DTYP, "asynInt32")
+    field(OUT, "@asyn($(PORT),0)HDF5_NDAttributeChunk")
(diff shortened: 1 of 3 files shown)


ulrikpedersen:
I've tested this and it works for me. 

Note that the default value of NDAttributeChunk=0 is probably the best option for 99% of use-cases. Except if you are writing really huge datasets (100.000+ frames). Because of that I think it is OK not to add this parameter to the screens - if someone accidentally (or without understanding the parameter) sets a small value, the performance of writing the attribute datasets may well suffer significantly.

Issue #72, opened 2015-03-05
HDF5: Starting file saving in Stream mode crashes IOC if NumCapture=0


MarkRivers:
The HDF5 plugin crashes if FileWriteMode=Stream, NumCapture=0, and Capture is set to 1.

The behavior of Stream mode when NumCapture=0 is not documented.  The netCDF plugin and the single-file writing plugins (TIFF, JPEG, etc.) treat NumCapture=0 as meaning infinite, i.e. continue streaming forever until it is stopped by setting Capture=0.  Perhaps we should document this as the "correct" behavior?

ulrikpedersen:
I agree that setting NumCapture=0 is a sensible API for allowing infinite number of frames to be written. It is easy to prevent this mode of operation from crashing the ioc, but we must be a bit careful as NumCapture is used in a number of places throughout the HDF5 plugin. For example to calculate the size of the internal hdf5 chunk index (a binary search tree). 

Also @ajgdls latest fix to #74 uses NumCapture to find a sensible default chunk size for NDAttribute datasets. 

I think I have some ideas on how to get around this and chose sensible (i.e. well performing) default values without knowing the final dataset size...

ulrikpedersen:
OK, I have had a go at writing support for NumCapture=0 (i.e. infinite number of frames). The feature basically works - but I have not yet tested it with large acquisitions, nor have I tested for performance on large file systems yet...

ulrikpedersen:
Fix in hdf5-issues branch PR #85

Issue #68, opened 2015-03-02
NDFileHDF5: Important default attributes are lacking


MarkRivers:
All NDArrays have certain important attributes, including UniqueID, timeStamp, epicsTS, and the dimension information for each dimension.  This critical information should be stored for all NDArrays in the HDF5 file even if it is not specifically requested in the XML layout file.

The following is an example of the output of ndump on a netCDF file that was written using no attributes.xml file for the driver, so there were no user-specific NDAttributes on the NDArray.

corvette:simDetectorIOC/iocBoot/iocHDF5Test>ncdump test_netcdf_002.nc
netcdf test_netcdf_002 {
dimensions:
        numArrays = 1 ;
        dim0 = 10 ;
        dim1 = 10 ;
        attrStringSize = 256 ;
variables:
        int uniqueId(numArrays) ;
        double timeStamp(numArrays) ;
        int epicsTSSec(numArrays) ;
        int epicsTSNsec(numArrays) ;
        byte array_data(numArrays, dim0, dim1) ;
        int Attr_ColorMode(numArrays) ;

// global attributes:
                :dataType = 1 ;
                :NDNetCDFFileVersion = 3. ;
                :numArrayDims = 2 ;
                :dimSize = 10, 10 ;
                :dimOffset = 0, 0 ;
                :dimBinning = 1, 1 ;
                :dimReverse = 0, 0 ;
                :Attr_ColorMode_DataType = "Int32" ;
                :Attr_ColorMode_Description = "Color mode" ;
                :Attr_ColorMode_Source = "Driver" ;
                :Attr_ColorMode_SourceType = "NDAttrSourceDriver" ;
data:

 uniqueId = 9 ;

 timeStamp = 794167395.448596 ;

 epicsTSSec = 794167395 ;

 epicsTSNsec = 449894139 ;

 array_data =
  8, 9, 10, 11, 12, 13, 14, 15, 16, 17,
  9, 10, 11, 12, 13, 14, 15, 16, 17, 18,
  10, 11, 12, 13, 14, 15, 16, 17, 18, 19,
  11, 12, 13, 14, 15, 16, 17, 18, 19, 20,
  12, 13, 14, 15, 16, 17, 18, 19, 20, 21,
  13, 14, 15, 16, 17, 18, 19, 20, 21, 22,
  14, 15, 16, 17, 18, 19, 20, 21, 22, 23,
  15, 16, 17, 18, 19, 20, 21, 22, 23, 24,
[...]

Issue #64, opened 2015-03-02
NDFileHDF5: Requested datasets whose values come from NDAttributes of type NDAttrString result in HDF5 attributes rather than HDF5 datasets


MarkRivers:
This was a design decision made in the original design of the HDF5 plugin.  However, it means that string attributes that change from one NDArray to the next cannot be stored in the file.  It also means that hardlinks cannot be created to datasets of this type, because they are not really datasets, they are attributes.  
The original restriction was because of the problem with string lengths varying from one NDArray to another.  In the netCDF plugin this is handled by storing string attributes as 2-D datasets, [MAX_ATTRIBUTE_STRING_SIZE, numArrays], i.e. using a fixed string size.  It is proposed to use a similar scheme in the HDF5 plugin.

ulrikpedersen:
Mark commented by email:

"I think it would be good to create a new parameter in asynNDArrayDriver which is NDAttrMaxStringSize.  It would replace MAX_ATTRIBUTE_STRING_SIZE and would allow the user to control the maximum allowed size of NDAttributeStrings in certain places, i.e. in the paramAttribute.cpp class, and the TIFF, netCDF, and HDF5 file plugins.  The default value of this parameter would be 256 for backwards compatibility."

ulrikpedersen:
It would not be difficult to create such a NDAttrMaxStringSize parameter, but I have a concern about how to deal with users changing this value (in particularly if it happens dynamically during acquisition). If this is allowed then we really have to support any-length strings (which we could probably do with STL strings anyway!)

For HDF5 datasets we would probably want it to be constant during acquisition - in fact we would only read it once at open-time in order to determine the max. string-length of string type NDAttributes. If it changes during acquisition that would be ignored by the file writer.

BTW, I'm listed as assigned, but really Alan Greer @ajgdls  will work on this - I just can't assign him because he is not part of this github organization...

MarkRivers:
I don't think it is a problem with it changing dynamically during acquisition.  As you said it would only be read during openFile(), not after that.  I'm happy postponing this and just using the current MAX_ATTRIBUTE_STRING_SIZE for now.

ulrikpedersen:
OK, I am most comfortable to stick with the MAX_ATTRIBUTE_STRING_SIZE for now as well.

We will at some point later have to work on multi-dimension NDAttributes - and variable length strings may fall naturally under that work.

ulrikpedersen:
Fix in hdf5-issues branch PR #85

PR #85, opened 2015-03-13
Hdf5 issues

ulrikpedersen:
A number of fixes to the following issues (which I am now closing):

* #64 NDFileHDF5: Requested datasets whose values come from NDAttributes of type NDAttrString result in HDF5 attributes rather than HDF5 datasets 
* #65 NDFileHDF5: Attributes defined for datasets in the xml file do not always appear in the HDF5 file. 
* #66 NDFileHDF5: If the xml layout file does not define a group with “ndattr_default=true” it crashes the application. 
* #68 NDFileHDF5: Important default attributes are lacking 
* #72 HDF5: Starting file saving in Stream mode crashes IOC if NumCapture=0 
* #74 HDF5: datasets that get their data from NDAttributes are initially created with the size of numArrays. 

During fixing we have tested each fix locally and I have now built and tested the full hdf5-issues branch for Linux and Windows. I made the tests using the supplied test1 and test2 XML files in the simDetectorIOC.

The following tasks should be done before doing this merge into master:
* Build and test by original creators of the tickets @MarkRivers and @prjemian 
 * If you find that it does not work as expected, please create a new issue ticket (or re-open a relevant one) and reference this PR...
* Decide whether to keep or delete the iocHDF5Test dir with the reproducing XML files(?)

Thanks @prjemian and @MarkRivers for tracking down and reproducing these issues for us and also thanks for the help with the fixing @aglowacki and @ajgdls :smiley:


ulrikpedersen:
Great, thanks for adding these tests @MarkRivers. Feel free to merge whenever you're happy with the state of the branch.

MarkRivers:
Merged hdf5-issues into master.  NDFileHDF5.html documentation still needs some work, but I would like others to begin testing the entire master branch.

Issue #86, opened 2015-03-18
HDF5 plugin fails to write files when additional dimensions are specified


ajgdls:
Whilst testing the HDF5 writer with additional virtual dimensions specified I received the error message
...
HDF5-DIAG: Error detected in HDF5 (1.9.178-swmr_chksum5) thread 140631045666560:
  #000: /dls_sw/prod/tools/RHEL6-x86_64/hdf5-swmr/1-9-178/src/hdf5-1.9.178-swmr_chksum5/src/H5Pdcpl.c line 1391 in H5Pset_chunk(): number of elements in chunk must be < 4GB
    major: Invalid arguments to routine
    minor: Out of range
...
Upon closer inspection it looks like the chunking parameters are not being correctly defined when additional dimensions are requested.

I'm investigating this problem now.

ajgdls:
I can repeat this with the following steps:
1) Start an acquisition with a detector (simulated in my case).
2) On the HDF5 plugin edm screen set the extra dims PV (TEST-86:HDF:NumExtraDims) to 2.
3) Set the extra dim X and extra dim Y PVs (TEST-86:HDF:ExtraDimSizeX and TEST-86:HDF:ExtraDimSizeY) to values other than the defaults.
4) Start the HDF5 plugin writing to file.
The error above along with a set of further errors is displayed.  If the write completes successfully on the first attempt then carry out 3 and 4 again, setting them to slightly larger values.  It looks like the chunking calculation is reading past the end of an array.

ajgdls:
Hi, I have just pushed an additional change to this branch and also a script that I was able to use to get the error that was described by Pete.  However, I could only guess at how Pete was getting this crash so I would be grateful if this could be tested by others (with or without the script).  I basically repeated it by setting the mode to single and then turning on autosaving, but perhaps it is possible to get the same situation with a 1 dimensional array passed in.  I discussed this with @ulrikpedersen  and we believe we are now covering all of the cases.
Cheers, Alan

PR #87, opened 2015-03-18
Only loop over 3 user chunking array elements to fix issue #86. 

ajgdls:
Within the configureDims method a loop is performed to calculate chunking
parameters.  There are three user chunking parameters declared but the
loop is looping over the number of dims, this has been changed with a
comment to explain what the loop is doing.

This fixes issue #86


MarkRivers:
hdf5-issues was merged into master yesterday.  Can you make this change to master instead and then issue a pull request for master?

ulrikpedersen:
Closing this PR. @ajgdls will create another one for the master branch.

PR #100, opened 2015-04-27
File locks

MarkRivers:
I reverted NDPluginFile.cpp on master to the R2-2 version, and created a new branch, file_locks, with the changes to NDPluginFile.cpp and all of the derived classes.

### DIFF
--- a/ADApp/pluginSrc/NDFileHDF5.cpp
+++ b/ADApp/pluginSrc/NDFileHDF5.cpp
@@ -62,37 +62,45 @@ asynStatus NDFileHDF5::openFile(const char *fileName, NDFileOpenMode_t openMode,
 {
   int storeAttributes, storePerformance;
   static const char *functionName = "openFile";
+  int numCapture;
+  asynStatus status = asynSuccess;
 
   asynPrint(this->pasynUserSelf, ASYN_TRACE_FLOW, "%s::%s Filename: %s\n", driverName, functionName, fileName);
+
+  /* These operations are accessing parameter library, must take lock */
+  this->lock();
+  getIntegerParam(NDFileNumCapture, &numCapture);
+  getIntegerParam(NDFileHDF5_storeAttributes, &storeAttributes);
+  getIntegerParam(NDFileHDF5_storePerformance, &storePerformance);
 
   // We don't support reading yet
   if (openMode & NDFileModeRead) {
     setIntegerParam(NDFileCapture, 0);
     setIntegerParam(NDWriteFile, 0);
-    return asynError;
+    status = asynError;
   }
   
   // We don't support opening an existing file for appending yet
   if (openMode & NDFileModeAppend) {
     setIntegerParam(NDFileCapture, 0);
     setIntegerParam(NDWriteFile, 0);
-    return asynError;
+    status = asynError;
   }
 
   // Check if an invalid (<0) number of frames has been configured for capture
-  int numCapture;
-  getIntegerParam(NDFileNumCapture, &numCapture);
   if (numCapture < 0) {
     asynPrint(this->pasynUserSelf, ASYN_TRACE_ERROR,
               "%s::%s Invalid number of frames to capture: %d. Please specify a number >= 0\n",
               driverName, functionName, numCapture);
-    return asynError;
+    status = asynError;
   }
 
-  // Verify the XML path and filename
+  // Verify the XML path and filename. Must be called with lock held.
   if (this->verifyLayoutXMLFile()){
-    return asynError;
+    status = asynError;
   }
+  this->unlock();
+  if (status != asynSuccess) return status;
 
 
   // Check to see if a file is already open and close it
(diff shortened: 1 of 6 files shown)


ulrikpedersen:
Thanks for doing this Mark - I couldn't find time for this right now. I'm off work tomorrow, but can review this on Wednesday and Thursday (probably with Alan as he is also in the office then).

MarkRivers:
Ulrik,
Any estimate on when you might merge this?

ulrikpedersen:
Hi Mark,

Sorry, bank holiday weekend - just back to work today. I'm reviewing it today and if I find no issues then I can probably hit the merge button by the end of today (UK time).

Cheers,
Ulrik

ulrikpedersen:
Hmmm... there is an issue here - a race condition. When the simDetector is running flat out - on my machine with the following settings:

* dimensions: 200 x 1024 x UInt32
* acquire period/time: 0.001s
* ~195fps (so simulator is CPU bound)
* HDF5 writer in Stream mode
* HDF5 writer is non-blocking mode

If the simDetector is already running, the HDF5 NumCaptured_RBV pv (NUM_CAPTURED parameter) is not reset to zero when starting HDF5 capture... So it starts counting from the number of frames stored in the previous run.  This has 2 (at least) implications:
 * Offsets are most likely not correct and data may be written in unexpected places in the file (although at least the writer didn't crash so it's not out of bounds!)
 * If the previous HDF5 capture acquisition was longer (or the same lenght as the current one) the NUM_CAPTURED parameter can already have increased beyond the desired value - and the file continue to grow (presumably forever). This is how I spotted the issue as my disk nearly filled up...

However, if the simDetector is **not** running (so CPU is idle) when starting HDF5 capture, the NUM_CAPTURED parameter **is** reset to 0 as it should be. It is because of this behaviour that I expect this is a race-condition.

I have not yet tracked down what causes this race condition. I am working on it, but feel free to pitch in if anyone spots it before me! :smiley:

ulrikpedersen:
Still haven't found it, but there are only a few places where the "NumCaptured" parameter is reset. That should narrow it down:

    [up45@pc0009 ADCore]$ grep -n -r setIntegerParam.*NDFileNumCaptured .    
    ./ADApp/ADSrc/asynNDArrayDriver.cpp:692:    setIntegerParam(NDFileNumCaptured, 0);
    ./ADApp/pluginSrc/NDPluginFile.cpp:229:     setIntegerParam(NDFileNumCaptured, 1);
    ./ADApp/pluginSrc/NDPluginFile.cpp:301:     setIntegerParam(NDFileNumCaptured, 0);
    ./ADApp/pluginSrc/NDPluginFile.cpp:425:     setIntegerParam(NDFileNumCaptured, 0);
    ./ADApp/pluginSrc/NDPluginFile.cpp:473:     setIntegerParam(NDFileNumCaptured, 0);
    ./ADApp/pluginSrc/NDPluginFile.cpp:537:     setIntegerParam(NDFileNumCaptured, 1);
    ./ADApp/pluginSrc/NDPluginFile.cpp:725:     setIntegerParam(NDFileNumCaptured, numCaptured);
    ./ADApp/pluginSrc/NDPluginFile.cpp:742:     setIntegerParam(NDFileNumCaptured, numCaptured);

ulrikpedersen:
I think this is a fundamental problem with unlocking inside around openFile():

If the simDetector is running flat out (200Hz or more) the [NDPluginFile::processCallbacks()](https://github.com/areaDetector/ADCore/blob/file_locks/ADApp/pluginSrc/NDPluginFile.cpp#L681) is called virtually *all the time* and one of the first things it does it take a copy of NDFileNumCaptured [here](https://github.com/areaDetector/ADCore/blob/file_locks/ADApp/pluginSrc/NDPluginFile.cpp#L705)

If the user then suddenly clicks on "Capture" we call writeInt32() the NDFileCapture parameter gets set [here](https://github.com/areaDetector/ADCore/blob/file_locks/ADApp/pluginSrc/NDPluginFile.cpp#L792)

However, before doCapture() has time to finish calling the virtual openFile() method (around which the mutex is unlocked) and setup all relevant parameters (like resetting NDFileNumCaptured) the really fast simDetector has caused another call to NDPluginFile::processCallbacks() - which now see the NDFileCapture parameter be true (or 1 or whatever) and promptly increment NDFileNumCaptured before calling writeFileBase() - and the inherited writeFile() method.

So the trouble is really that NDFileCapture is set in the parameter library too early - before the inherited openFile() method has finished and the file writer is truly ready to write frames.

What I really don't understand is why this doesn't cause a more visible crash!!!

ulrikpedersen:
So I think I've got it now (commit b6d18f5e998c5268f961a6387efdd52245a18b69)... The issue was really that in the writeInt32() call to start capture, the mutex is unlocked before the openFile() call has completed. This allowed starting to write frames before the writer was truly in a good state to write them!

Please, check if this works for you @MarkRivers and @ajgdls

MarkRivers:
I reproduced the problem with numCaptured not being reset to 0 in StreamMode when Capture was started with a fast simeDetector  in my existing version of the file_locks branch.  I then fetched the latest version from github and the problem went away.

I also switched to the master branch and reproduced the problem with Single mode creating 23MB HDF5 files if NumCapture was 0, but 1.1MB files if NumCapture was > 0.  This problem also went away with the latest file_locks branch.

PR #106, opened 2015-05-27
Changed sseq_settings.req to sseqRecord_settings.req

kmpeters:
Changed sseq_settings.req to sseqRecord_settings.req, which is the preferred filename in new versions of the calc module. This is also required for autosaveBuild calls to succeed.

### DIFF
--- a/ADApp/Db/NDStats_settings.req
+++ b/ADApp/Db/NDStats_settings.req
@@ -12,5 +12,5 @@ $(P)$(R)HistMax
 $(P)$(R)TSNumPoints
 $(P)$(R)TSRead.SCAN
 file "NDPluginBase_settings.req", P=$(P), R=$(R)
-file "sseq_settings.req", P=$(P), S=$(R)Reset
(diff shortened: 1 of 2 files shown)


kmpeters:
Missed a relevant file.

MarkRivers:
I just looked at the trunk version of the calc module.  Not only is sseqRecord_settings.req the preferred filename as you said, it is the ONLY filename.  sseq_settings.req no longer exists.  Is this the plan?  But then why has swait_settings.req not be renamed the same way for consistency.  It seems that deleting sseq_settings.req may cause a lot of grief because people will reboot an IOC with the next release of calc without changing their auto_settings.req file, and they will no longer be saving their sseq record settings at all.

MarkRivers:
I just check the released versions of the calc module.  There is no released version of the calc module that contains the file sseqRecord_settings.req.  I definitely do not want to make ADCore depend on an unreleased version of the calc module.  I am closing this pull request.  You can re-open it when there is a released version of the calc module that contains sseqRecord_settings.req.

Issue #108, opened 2015-06-29
Make optional dependencies optional


kmpeters:
There are a number of EPICS modules that are useful but not required for areaDetector IOCs.  An example of this is devIocStats, which was proposed in issue #103.  These optional modules should be added to ADCore in such a way that that they are easily included if available.  The omission of these modules should not break the ADCore build.

MarkRivers:
This has been done in R2-3.

PR #109, opened 2015-06-29
Optionally include useful modules

kmpeters:
Fixes issue #108


MarkRivers:
I think perhaps sscan and calc belong in the same category as devIocStats and alive.  calc is only needed for sseq, and sscan is only needed if scanning capability is desired.  commonDriverMakefile could be changed accordingly.

I don't like creating a new "extra.db".  It means that all IOC st.cmd files would need to be edited to include it if I want to use devIocStats, which I do. I think commonPlugins.cmd should be changed to add comment lines saying that the optional SSCAN module is required to use that command.  Users are going to need to edit extra.cmd anyway because they won't have "alive", for example, make it simpler and edit commonPlugins.cmd instead.

MarkRivers:
Your indentation in commonDriverMakefile is not consistent with the rest of the file.  It looks like perhaps you used tab characters?  Please do not use tab characters in any files (except Makefile rules, where they are required).

ulrikpedersen:
I'm all in favour of a "lean" basic build - with a couple of options to enable a more complex build if this is desired. However, when the number of options grow, the test-matrix for the maintainers also grow as we (ideally) need to ensure every conceivable combination of options will work (build and run). We can expand our build matrix of course, but I feel we may need to consider alternatives...

Perhaps this is a topic which is a bit beyond this particular PR (and it's just a crazy idea), but how about moving all IOC related stuff out of ADCore and into a separate repo? Then we can keep ADCore as a lean software library and not littered with various IOC configuration files. That way we only have a single (or maybe two) build configuration of the core libraries - and a separate "ADIOCs" repo with the more complex configurations for building IOCs with a whole host of optional dependencies... This idea sort of follows on with what we started when splitting areaDetector drivers apart.

As mentioned, this is just a crazy thought and is not intended to hold up the work ongoing with this PR... :smiley:

kmpeters:
> Your indentation in commonDriverMakefile is not consistent with the rest of the file. It looks like perhaps you used tab characters?

I corrected the indentation.  The lines were copied from xxx/xxxApp/src/Makefile, which uses tabs for indentation.  I failed to notice the mismatch.

> I don't like creating a new "extra.db". It means that all IOC st.cmd files would need to be edited to include it if I want to use devIocStats, which I do.

I moved the content of extra.cmd to commonPlugins.cmd

> I think commonPlugins.cmd should be changed to add comment lines saying that the optional SSCAN module is required to use that command.

I modified the comments.

kmpeters:
> Perhaps this is a topic which is a bit beyond this particular PR (and it's just a crazy idea), but how about moving all IOC related stuff out of ADCore and into a separate repo? Then we can keep ADCore as a lean software library and not littered with various IOC configuration files. That way we only have a single (or maybe two) build configuration of the core libraries - and a separate "ADIOCs" repo with the more complex configurations for building IOCs with a whole host of optional dependencies... This idea sort of follows on with what we started when splitting areaDetector drivers apart.

I think this is an excellent idea.  It would make it much easier to fork the IOC code and include site-specific modules.

MarkRivers:
> and a separate "ADIOCs" repo with the more complex configurations

I have thought of this. It could just be ADSimDetector, just another detector repository. 

The question then becomes where do we put commonDriverMakefile and commonPlugins.cmd?  In ADSimDetector?  In ADCore?  

Mark

ulrikpedersen:
Great! :smile: I was halfway expecting to be booed out with this suggestion!

> I have thought of this. It could just be ADSimDetector, just another detector repository.

That sounds fine to me. I don't actually mind whether the simDetector driver goes in either ADCore or the new IOC repository. My main concern is just that all the IOC-related configurations live in a separate repo.

> The question then becomes where do we put commonDriverMakefile and commonPlugins.cmd?  In ADSimDetector?  In ADCore?

Well, as the commonDriverMakefile is really only used to building IOCs, I would suggest putting it in the new IOC module as well. The idea is to keep ADCore as lean as possible so that the only interface it publish to other developers are include files, libraries and plugin screens (as the basic plugins are part of ADCore).

prjemian:
+1 for the separate IOC repo

Can this separate IOC repo also simplify the user question I hear: 
"Where are the XML files I need to modify for the HDF5 plugin?"

Pete

MarkRivers:
Your pull request still has extra.cmd being loaded from st.cmd.  That should be removed?

MarkRivers:
I manually made most of these changes.  The only change I did not do was the one to st.cmd.

Issue #110, opened 2015-07-06
HDF5 File Writer is not setting istorek parameter correctly.


ajgdls:
This bug was introduced during commits for #100 which altered the ordering of some variable definitions.  The result here is that a counter jumps off the end of an array and the istorek parameter is then set to an uninitialised value.

ulrikpedersen:
This is fixed in PR #111 
Closing issue.

PR #111, opened 2015-07-06
Reorder the code for setting extradim to fix issue #110.

ajgdls:
This is a small fix to ensure the HDF5 istorek parameter is set correctly.  Fixes issue #110.

### DIFF
--- a/ADApp/pluginSrc/NDFileHDF5.cpp
+++ b/ADApp/pluginSrc/NDFileHDF5.cpp
@@ -2899,7 +2899,7 @@ asynStatus NDFileHDF5::createNewFile(const char *fileName)
   hid_t create_plist = H5Pcreate(H5P_FILE_CREATE);
   unsigned int istorek = this->calcIstorek();
   asynPrint(this->pasynUserSelf, ASYN_TRACE_FLOW,
-            "%s::%s Setting istorek=%u\n",
+            "%s::%s Calculated istorek=%u\n",
             driverName, functionName, istorek);
   // Check if the calculated value of istorek is greater than the maximum allowed
   if (istorek > MAX_ISTOREK){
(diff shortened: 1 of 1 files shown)


ulrikpedersen:
I can think of a couple of improvements that we can do which relates to this rather troublesome algorithm:

1) Lets get rid of the bit of code in line 1904, so if the algorithm returns an invalid value we deal with it at a higher level as an error/warning condition (see 2):
    
    if (retval <= 0) retval = 1; // delete me!

2) Lets check the calculated istore_k value, just before setting it in line 2899 - if the value is <=0; print an asyn WARNING - and skip the call to H5Pset_istore_k(). Most likely the writer will continue with the default istore_k; although not at optimal performance when writing huge files. I think this is better than accidentally producing invalid files ;-)

ulrikpedersen:
Actually I have two more improvements, regarding this istore_k parameter:

3) There is an undocumented higher limit of 32767 to the istore_k parameter, which has caused us some pain in the past. The HDF Group is aware of this issue. So I suggest inserting a high (32767) and low (1?) check on the calculated istorek parameter, just before calling H5Pset_istore_k() - print an asyn WARNING if the value is invalid.

4) Arrrggg - I've just noticed... Let's check the status return code from H5Pset_istore_k()!!! :smile:

ajgdls:
The problem we suffered here was that the calculated value was multiplied by nonsense.  We are checking for a value less than 1 and then imposing a limit of 1.  This check can be carried out at the higher level, along with the upper limit check but if we catch these error conditions we should probably not try to use them in the HDF5 call as we don't know the outcome in that case.  I would go for catching the error values and not setting istore_k, instead printing the warning and accepting a default.  We should still check the outcome of the HDF5 call also, in case of other errors that we are not aware of.

MarkRivers:
Alan and Ulrik, are you happy with this pull-request and ready for me to do it?

ajgdls:
I'm happy with the change, over to you Ulrik...

ulrikpedersen:
Yes, I think this one is ready for merge. Please do go ahead @MarkRivers

PR #118, opened 2015-08-18
Correct NDROIStats Screens

mdmoo1978:
Corrected ROI9-32 related displays. They were 8-31.


MarkRivers:
The change in NDROIStat.adl is fine.  However, there are 2 problems with NDROIStat.ui.  First, this should always be autoconverted, not manually converted.  Second, there are errors in the file because the extension .adl is used where .ui should be used.  Please remove NDROIStat.ui from the  pull request.

mdmoo1978:
Actually I did autoconvert NDROIStat.ui from the .adl file and caQtDM is
smart enough to handle .adl to .ui conversion.

PR #119, opened 2015-08-20
Support for ADPluginEdge which also requires some opencv libraries.

kebrister:
This optionally adds ADPluginEdge support.

### DIFF
--- a/ADApp/commonDriverMakefile
+++ b/ADApp/commonDriverMakefile
@@ -4,6 +4,14 @@
 $(PROD_NAME)_DBD += commonDriverSupport.dbd
 
 PROD_LIBS               += NDPlugin ADBase netCDF
+
+
+ifdef ADPLUGINEDGE
+  PROD_LIBS             += NDPluginEdge
+  PROD_SYS_LIBS_Linux   += opencv_core opencv_imgproc opencv_imgcodecs
+  $(PROD_NAME)_DBD      += NDPluginEdge.dbd
+endif
+
 ifneq (vxWorks, $(findstring vxWorks, $(T_A)))
   PROD_LIBS             += NeXus
 endif


ulrikpedersen:
I don't really think this belongs in ADCore as the plugin itself is not part of ADCore. If you want to help your IOC builder, then the ADPluginEdge module could provide a similar makefile to be included in the IOCs src/Makefile...

kebrister:
Hi Ulrik,

Any suggestions are welcome as I'm a bit beyond my level of competence with
the EPICS build system.

-Keith


On Fri, Aug 21, 2015 at 3:14 AM, Ulrik Kofoed Pedersen <
notifications@github.com> wrote:

> I don't really think this belongs in ADCore as the plugin itself is not
> part of ADCore. If you want to help your IOC builder, then the ADPluginEdge
> module could provide a similar makefile to be included in the IOCs
> src/Makefile...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/areaDetector/ADCore/pull/119#issuecomment-133326449>.
>



-- 
Dr. Keith Brister
LS-CAT Operations Manager
Research Associate Professor
Synchrotron Research Center
Northwestern University
9700 S. Cass Ave, Bldg 436A
Argonne, IL 60439
1-630-343-9532 (V)
1-630-252-4664 (F)

ulrikpedersen:
Hi Keith,

There is not really anything wrong with the code you've added - it's just not really in the appropriate repository as ADCore essentially does not know anything about this Edge plugin...

Instead I would suggest creating a new "commonPluginEdgeMakefile" somewhere in your ADPluginEdge repo - which you can then include in IOCs src/Makefile to automatically pull in these library dependencies.

I'll close this PR then.

Cheers,
Ulrik

MarkRivers:
This actually would be appropriate once we move the simDetector into a new ADExample repository.  That is then where commonDriverMakefile will reside.

The build order in the top-level areaDetector Makefile will then be:

ADBinaries
ADCore
ADPluginEdge
(other optional plugins)
ADExample
(other detectors)

commonDriverMakefile can then contain references to optional plugins like ADPluginEdge.

kebrister:
I'm a little confused at how to proceed: should I really modify the IOC
makefiles, such as prosilicaApp, to include commonPluginEdgeMakefile?  That
seems less good than just modifying commonDriverMakefile but I'm probably
missing some key step here.

-Keith


On Fri, Aug 21, 2015 at 12:20 PM, Mark Rivers <notifications@github.com>
wrote:

> This actually would be appropriate once we move the simDetector into a new
> ADExample repository. That is then where commonDriverMakefile will reside.
>
> The build order in the top-level areaDetector Makefile will then be:
>
> ADBinaries
> ADCore
> ADPluginEdge
> (other optional plugins)
> ADExample
> (other detectors)
>
> commonDriverMakefile can then contain references to optional plugins like
> ADPluginEdge.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/areaDetector/ADCore/pull/119#issuecomment-133501708>.
>



-- 
Dr. Keith Brister
LS-CAT Operations Manager
Research Associate Professor
Synchrotron Research Center
Northwestern University
9700 S. Cass Ave, Bldg 436A
Argonne, IL 60439
1-630-343-9532 (V)
1-630-252-4664 (F)

MarkRivers:
Your are correct, modifying commonDriverMakefile is the better approach.  However, that won't work until we move the simDetector IOC into its own repository.  The problem is that ADCore needs to be built before ADPluginEdge, since it needs to install the header files and libraries.  But if we want to include ADPluginEdge in simDetector, which we definitely do for testing purposes, then ADPluginEdge needs to be built before ADCore, since simDetector is in ADCore.  So we have a conflict.  The solution is to move simDetector into a new repository called ADSimDetector or ADExample or some other name.  That is where we will then place commonDriverMakefile and commonPlugins.cmd.  ADCore will then also not require any EPICS dependencies other than EPICS base and asyn.

I plan to work on this change soon.  I just have some questions about how to do it correctly.

kebrister:
That sounds like the right approach.  Thanks.

On Fri, Aug 21, 2015 at 1:47 PM, Mark Rivers <notifications@github.com>
wrote:

> Your are correct, modifying commonDriverMakefile is the better approach.
> However, that won't work until we move the simDetector IOC into its own
> repository. The problem is that ADCore needs to be built before
> ADPluginEdge, since it needs to install the header files and libraries. But
> if we want to include ADPluginEdge in simDetector, which we definitely do
> for testing purposes, then ADPluginEdge needs to be built before ADCore,
> since simDetector is in ADCore. So we have a conflict. The solution is to
> move simDetector into a new repository called ADSimDetector or ADExample or
> some other name. That is where we will then place commonDriverMakefile and
> commonPlugins.cmd. ADCore will then also not require any EPICS dependencies
> other than EPICS base and asyn.
>
> I plan to work on this change soon. I just have some questions about how
> to do it correctly.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/areaDetector/ADCore/pull/119#issuecomment-133528541>.
>



-- 
Dr. Keith Brister
LS-CAT Operations Manager
Research Associate Professor
Synchrotron Research Center
Northwestern University
9700 S. Cass Ave, Bldg 436A
Argonne, IL 60439
1-630-343-9532 (V)
1-630-252-4664 (F)

ulrikpedersen:
I agree - that is a more complete solution. Thanks Mark.

PR #128, opened 2015-09-09
Allow boost, hdf5, and sz to be SYS_LIBS 

MarkRivers:
The main problem we discussed earlier is when library A should use a system library in the normal default location, but library B exists in the system location but that version system should not be used for library B.  The problem is that if A_DIR is specified then the default system location can be searched first, and it will find the wrong version of library B.  This version of the Makefile allows defining BOOST but not BOOST_LIB, not defining HDF5_LIB, and defining SZIP but not SZIP_LIB.  If the _LIB definition is present it defines the _DIR and adds the library to PROD_LIBS.  If the _LIB definition is not present then it does not define the _DIR and adds the library to PROD_SYS_LIBS.  This prevents the system libraries from being added early in the library search path, and should fix the most common issues.

### DIFF
--- a/ADApp/pluginTests/Makefile
+++ b/ADApp/pluginTests/Makefile
@@ -28,8 +28,12 @@ ifdef BOOST
   #plugin-test_SRCS += test_<plugin name>.cpp
   
   plugin-test_LIBS += ADTestUtility
-  plugin-test_SYS_LIBS += boost_unit_test_framework
-  boost_unit_test_framework_DIR=$(BOOST_LIB)
+  ifdef BOOST_LIB
+    boost_unit_test_framework_DIR=$(BOOST_LIB)
+    plugin-test_LIBS += boost_unit_test_framework
+  else
+    plugin-test_SYS_LIBS += boost_unit_test_framework
+  endif
 endif
 
 USR_INCLUDES += $(HDF5_INCLUDE)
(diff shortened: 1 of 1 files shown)


ulrikpedersen:
This is elegant and works for both my possible configurations (and on Travis):

Boost                  | HDF5              | status  
-------------------|----------------|--------
standard system | non-standard | OK 
non-standard      | non-standard | OK 

First case:

```
[up45@pc0009 ADCore]$ ldd bin/linux-x86_64/plugin-test | grep -E '(boost)|(hdf5)'
	libhdf5.so.9 => /dls_sw/prod/tools/RHEL6-x86_64/hdf5/1-8-14/prefix/lib/libhdf5.so.9 (0x00007fb0e9a2a000)
	libboost_unit_test_framework.so.5 => /usr/lib64/libboost_unit_test_framework.so.5 (0x00007fb0e7db5000)
```

Second case:

```
[up45@pc0009 ADCore]$ ldd bin/linux-x86_64/plugin-test | grep -E '(boost)|(hdf5)'
	libboost_unit_test_framework.so.1.48.0 => /dls_sw/prod/tools/RHEL6-x86_64/boost/1-48-0/prefix/lib/libboost_unit_test_framework.so.1.48.0 (0x00007ffb5fa31000)
	libhdf5.so.9 => /dls_sw/prod/tools/RHEL6-x86_64/hdf5/1-8-14/prefix/lib/libhdf5.so.9 (0x00007ffb5ed3e000)
```

Issue #130, opened 2015-09-10
NDAttribute sourceType is not initialised.


ajgdls:
I think this is a problem (sorry if I misunderstand what should be happening in the NDAttribute class)...

In the NDAttribute constructor that takes a NDAttrSource_t sourceType parameter, the source type is not stored as a member of the class.  The string representation of the source is correctly allocated but there is no line that looks like

this->sourceType = sourceType;

or 

sourceType(sourceType)

in the constructor.

If the getSourceInfo method is later called then the pointer to NDAttrSource_t appears to be filled with junk as the sourceType member variable is uninitialised.

MarkRivers:
Fixed in pull request #131.

PR #131, opened 2015-09-10
NDAttribute::sourceType was never being set

MarkRivers:
NDAttribute::sourceType was never being set. Set it in constructor.  Add new NDAttrSourceUndefined enum; print NDAttribute::sourceType in NDAttribute::report().


MarkRivers:
This fixes issue #130.  I am merging it myself since I think it is pretty straightforward.

PR #133, opened 2015-09-15
Fixes to Attributes handling

brunoseivam:
A couple of issues on the conversion code between NDArrays and NTNDArrays regarding Attributes handling were fixed. Namely:

* Undefined attributes from NDArray to NTNDArray are now parsed correctly instead of throwing an exception.
* Source and source type of Attributes are now correctly parsed from NTNDArray to NDArray instead of being ignored.

### DIFF
--- a/ADApp/ADSrc/NDAttribute.cpp
+++ b/ADApp/ADSrc/NDAttribute.cpp
@@ -45,6 +45,7 @@ const char *NDAttribute::attrSourceString(NDAttrSource_t type)
 
   this->pName = pName? epicsStrDup(pName): epicsStrDup("");
   this->pDescription = pDescription? epicsStrDup(pDescription): epicsStrDup("");
+  this->sourceType = sourceType;
   switch (sourceType) {
     case NDAttrSourceDriver:
       this->pSourceTypeString = epicsStrDup("NDAttrSourceDriver");
@@ -59,6 +60,7 @@ const char *NDAttribute::attrSourceString(NDAttrSource_t type)
       this->pSourceTypeString = epicsStrDup("NDAttrSourceFunct");
       break;
     default:
+      this->sourceType = NDAttrSourceUndefined;
       this->pSourceTypeString = epicsStrDup("Undefined");
   }
   this->pSource = pSource? epicsStrDup(pSource): epicsStrDup("");
@@ -385,7 +387,8 @@ int NDAttribute::report(FILE *fp, int details)
   fprintf(fp, "NDAttribute, address=%p:\n", this);
   fprintf(fp, "  name=%s\n", this->pName);
   fprintf(fp, "  description=%s\n", this->pDescription);
-  fprintf(fp, "  source type=%s\n", this->pSourceTypeString);
+  fprintf(fp, "  source type=%d\n", this->sourceType);
+  fprintf(fp, "  source type string=%s\n", this->pSourceTypeString);
   fprintf(fp, "  source=%s\n", this->pSource);
   switch (this->dataType) {
     case NDAttrInt8:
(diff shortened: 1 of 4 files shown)


ulrikpedersen:
Hi @brunoseivam 

Did you cherry-pick that commit from Mark? "NDAttribute::sourceType was never being set; set in constructor...". Do you know whether that will cause a conflict when we eventually merge the "v4-plugin" branch into "master" where that commit is already present?

I'm hoping (I think) the answer is "don't worry git will work it out and do the right thing..." ;-)

brunoseivam:
Yes, I cherry picked it. It seems it won't cause problems (according to this [answer](http://stackoverflow.com/questions/14486122/how-does-git-merge-after-cherry-pick-work/14486425#14486425)). The commit hashes are different. Nevertheless I will be careful when merging master back into v4-plugin after you guys finish the separation between ADCore and the iocs. 

BTW, do you know when you guys will want to merge v4-plugin into master?

ulrikpedersen:
Thanks @brunoseivam 

> BTW, do you know when you guys will want to merge v4-plugin into master?

We are basically ready for it - but want to have some confidence that it works in real-world cases first. That's how we came across this issue.
Right now @MarkRivers is working on preparing release 2-4 with the ADCore/ADExample split. I don't expect we can get enough testing done on the v4-plugin in time for 2-4. But if we don't stumble on other issues then I expect we can merge into master shortly after 2-4. 

On our side @ajgdls is doing the testing as part of another project where we want to use this plugin.

MarkRivers:
Some things could be done before the merge.

- Merge master back into the v4-plugin branch, make sure it still works.

- Make a v4-plugin branch in areaDetector/areaDetector that adds what is need to .gitmodules, EXAMPLE_RELEASE_PATHS.local, EXAMPLE_CONFIG_SITE.local, Makefile, INSTALL.md, documentation/Doxyfile, etc.

- Make a v4-plugin branch in ADExample that allows testing the plugin and driver with simDetector.

Mark

ulrikpedersen:
> Merge master back into the v4-plugin branch, make sure it still works

Absolutely. I would just wait until we're certain that the ADCore/ADExample split has been completed - so we don't have to do it several times. Do you reckon we are there now with your changes from yesterday, @MarkRivers ?

> Make a v4-plugin branch in areaDetector/areaDetector that adds what is need to .gitmodules, EXAMPLE_RELEASE_PATHS.local, EXAMPLE_CONFIG_SITE.local, Makefile, INSTALL.md, documentation/Doxyfile, etc.
> Make a v4-plugin branch in ADExample that allows testing the plugin and driver with simDetector.

OK that is all the non-development stuff which makes plugins actually useful... Volunteers?  @brunoseivam as you're the original author?

ajgdls:
Hi Bruno,

Although this update has fixed this specific Attribute handling issue, during the test of this change the IOC seg faulted, also due to Attribute handling.

To recreate the problem...
1) Start a detector IOC, enable the V4 plugin.
2) Take a single frame (with no attribute file loaded).
3) Now load an attribute file.
4) Take another frame, crash occurs.

Some initial digging shows that in the method NTNDArrayConverter::fromAttributes after the attributes have been loaded the destVec has a size greater than the attribute list (srcList->count()), and so the loop ends up running off the end of the attribute list.

I'm not familiar enough with the code to be certain, but it looks like we are adding attributes to destVec even though they have already been added previously (see createAttributes method, called whenever the source list size differs from the current dest size).

I guess it would be good to get that fixed as part of this pull request too.

Cheers, Alan

MarkRivers:
I think it is fine to merge ADCore master into v4-plugin.  I don’t plan to make any more changes to ADCore before R2-4.

brunoseivam:
@ulrikpedersen Sure, sign me up. 

@ajgdls Thanks for the debugging! Yes, it was just appending stuff to the previous vector instead of building a new one from scratch. 

However, I noted that the code only rebuilds the list of attributes when the current size is different from the previous one. So, this won't work if an attribute is removed and another one is added between frames. I am working on that.

brunoseivam:
I think there won't be any issues anymore *fingers crossed*

ajgdls:
Hi @brunoseivam the tests are now working, thanks!

PR #139, opened 2015-09-29
functional replacement

prjemian:
yySseq.db has been removed from synApps CALC, sseqRecord.db is identical replacement


MarkRivers:
This requires calc R3-5.  However, the calc home page:
http://www.aps.anl.gov/bcda/synApps/calc/calc.html
does not contain a link to R3-5.  Please add that link and also add a pull request for the areaDetector/configure/EXAMPLE_RELEASE_PRODS.local that uses the new required version of calc.  Then I will merge this pull request.

MarkRivers:
calc 3-6-1 is now on the BCDA synApps Web page.  I changed EXAMPLE_RELEASE_PRODS.local to reference this version of calc, which does have sseqRecord.db available. I also changed iocBoot/commonPlugin_settings.req to use sseqRecord_settings.req rather than sseq_settings.req because the latter is deprecated.

PR #140, opened 2015-10-25
Add timestamps waveform to timeseries

stuwilkins:
Hi,

I am submitting this as a suggestion to add a new waveform to the stats plugin to record the timestamp of each image in the timeseries buffer along with the actual values. This means that the data can be used later to match up with other values. This of course means the pImage->timeStamp value needs to be correct and hopefully set by hardware.

Any comments? Is there a better way to do this? 

S

### DIFF
--- a/ADApp/Db/NDStats.template
+++ b/ADApp/Db/NDStats.template
@@ -481,6 +481,14 @@ record(waveform, "$(P)$(R)TSSigmaXY")
    field(SCAN, "I/O Intr")
 }
 
+record(waveform, "$(P)$(R)TSTimestamp")
+{
+   field(DTYP, "asynFloat64ArrayIn")
+   field(INP,  "@asyn($(PORT),$(ADDR),$(TIMEOUT))TS_TIMESTAMP_VALUE")
+   field(NELM, "$(NCHANS)")
+   field(FTVL, "DOUBLE")
+   field(SCAN, "I/O Intr")
+}
 
 ###################################################################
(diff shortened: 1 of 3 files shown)


ulrikpedersen:
Hi @stuwilkins thanks for the PR :+1: The changes look fine to me at a quick glance.

One minor practical comment: one of your two commits don't have a comment to describe the change (it just has "[ENH]" as a comment). When merging this PR we should add a suitable commit message. 

Alternatively, and to make life easier for the maintainers - if you're up for some git magic you can edit (reword) the commit message on your repository in an interactive rebase (git rebase -i HEAD~2). Warning: this will essentially rewrite your git history (the latest 2 commits on "add_timestamps") and may cause trouble if others have cloned you repository in the mean time. Github will automatically update this Pull Request to match the state of your commits on the "add_timestamps" branch after you push (with --force).

stuwilkins:
HI @ulrikpedersen. Thanks ! I made the change you suggested and added the commit message. I obviously should have added it. I wonder if people agree that this should be added to the NDPluginROIStat module as well?

ulrikpedersen:
Cool - that was quick! :smile: 

> I wonder if people agree that this should be added to the NDPluginROIStat module as well?

I think that is a good idea. Perhaps the authors @mp49 or @MarkRivers has an opinion on this?

MarkRivers:
That seems like a good idea to me.

Mark

PR #142, opened 2015-11-18
Don't include HDF5, SZIP, nor XML2 on vxWorks

jlmuir:


MarkRivers:
I'm not this is the best way to do this.  Perhaps a better way is to create a file areaDetector/configure/CONFIG_SITE.local.vxWorks-ppc32 (for example) and to define HDF5_INCLUDE, etc. to be empty strings there.  

Were you getting errors, or what prompted you to do this?

jlmuir:
> I'm not sure this is the best way to do this. Perhaps a better way is to create a file areaDetector/configure/CONFIG_SITE.local.vxWorks-ppc32 (for example) and to define HDF5_INCLUDE, etc. to be empty strings there.

I'm open on how best to solve the problem, but I don't think I should have to create a CONFIG_SITE.local.* file to get it to build; I think it should build correctly out of the box.

> Were you getting errors, or what prompted you to do this?

Yes, the build fails as shown in [areaDetector-2-4-build-fail.txt](https://github.com/areaDetector/ADCore/files/38323/areaDetector-2-4-build-fail.txt).

MarkRivers:
I have changed areaDetector/configure/EXAMPLE_CONFIG_SITE.local so that HDF5, SZIP, and XML2 defines are not done on vxWorks.  This fixes the warnings I was getting for XML2 on vxWorks 6.9.4.1, and will presumably fix the errors that Lewis was getting on vxWorks 5.5.

PR #143, opened 2015-12-13
Cleaning conditional directives that break statements.

lzmths:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

    https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892
    https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/ADApp/netCDFSrc/libsrc/ffio.c
+++ b/ADApp/netCDFSrc/libsrc/ffio.c
@@ -392,12 +392,14 @@ ncio_ffio_global_test(const char *ControlString)
 static int
 ncio_ffio_sync(ncio *const nciop)
 {
+	int test;
 #ifdef __crayx1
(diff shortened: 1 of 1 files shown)


MarkRivers:
In general I would agree with making changes like this.  However, in this specific case the files in this directory are "vendor" files from the Unidata netCDF library.  I want to make as few changes as possible to these files to facilitate future updates.  The only changes I have made are those required to get it to build on all of our target architectures (Linux, Darwin, Windows, vxWorks, etc.)

lzmths:
I understand. Thank you for your honest opinion.

PR #146, opened 2016-02-05
Set EnableCallbacks from db, and set ArrayCallbacks to 1 by default

thomascobb:
We've had this local patch at DLS for a while, it means that:
* all areaDetector drivers produce ArrayCallbacks by default. This is a change from the current behaviour, but we've never had a reason to turn ArrayCallbacks off apart from debugging.
* all areaDetector plugins can optionally be turned on by setting Enabled=1 in the substitution file. If this is not specified, the behaviour is the same as it is currently.

We've found this very useful, as we can deploy an IOC without having to go and enable manually every plugin and driver we need, and then rely on autosave to keep those changes for us.

Would this be acceptable?


MarkRivers:
Those changes look good.  I agree it will prevent a lot of first-time user confusion.

MarkRivers:
I changed Enabled=0 to ENABLED=0.  All database macro parameters in areaDetector are upper-case.

PR #147, opened 2016-02-18
HDF5 SWMR

ajgdls:
This pull request is to add support for SWMR (Single Writer Multiple Reader) when writing HDF5 files, and to add support for selective positional placement of frames within a dataset.  The changes have been limited to the HDF5 classes but those files have been changed substantially.  There is also available a new positional placement plugin that can be used to specify exactly where each frame is placed within a dataset.  The HDF5 writer now also supports up to 10 extra dimensions, instead of the previous 3.

I have implemented the changes such that running the software without turning on any of the features should produce HDF5 files exactly as they would have been produced prior to these changes, and if SWMR support is available but not turned on then the HDF5 files should still be compatible with standard HDF5 tools (prior to the change of file structure introduced by HDF5 group during SWMR development).

I would really appreciate it if everyone could try out their own specific setup with this new version of the writer and check that it does still work in the same way for you if additional features are not switched on.  Please report any problems that you notice.  After that please go ahead and try out the new features too.

I've updated the documentation so let me know if it is not clear.


ajgdls:
OK so we just discovered that Travis is building with an old version of the HDF5 library that doesn't have the version checking macro that is used to check whether the HDF5 version supports SWMR.  I'm going to check for the existence of the macro in the HDF5 public header and then define it for the file if it is not already defined.

I will push when I have the solution and hopefully the Travis build will no longer fail.

MarkRivers:
In order to use the SWMR feature one needs to use HDF5-1.10-alpha0 or alpha1?  Is there an estimate of when a stable HDF5-1.10 release will occur?

tacaswell:
Official word is end of March https://github.com/h5py/h5py/issues/670#issuecomment-175000734

ulrikpedersen:
For testing you can use any of tge alphas 0, 1 (or the upcoming 2). However, the HDF Group discourage use of alpha in production as the file format is not set in stone - so you could potentially end up with files that can not be read with the final 1.10 library... 

Thus I suggest that we only update ADBinaries once the final 1.10 release is out (end of March). 

By all means please download, build and test the alphas as much as you can! We have done a significant amount of testing at Diamond throughout the development of SWMR and are fairly confident in it. We will switch to use it in production almost immediately after the 1.10 release.

MarkRivers:
If I understand correctly, files written with SWMR will not be readable with programs built with HDF5 versions prior to 1.10?

tacaswell:
Correct, the new libraries can read old files, but old libraries can not
read new files.

ajgdls:
I would appreciate any additional testing under varying environments.  The HDF5 alphas that support SWMR contain a command line tool for monitoring (h5watch).  I've also used a lot of python scripts with h5py for monitoring datasets.

ulrikpedersen:
> If I understand correctly, files written with SWMR will not be readable with programs built with HDF5 versions prior to 1.10?

Correct. However, the SWMR option is *not* switched ON by default on the HDF5 writer. So by default you will write non-swmr files (even with library v1.10) and they *can* be opened by HDF5 1.8.x

Also, once you have written a SWMR file you can convert it from the v1.10 format back down to v1.8 using the supplied h5format_convert tool. This is a meta-data operation only which operates directly on the file - thus the file is converted without copying all the data. Then the file can be opened by HDF5 v1.8.x

stuwilkins:
Hi, I can test here at NSLS-II at CSX with the FastCCD we won't do this for user data but we can do this for some tests where we take large data sets and see how it behaves. I will aim to get any comments back before the new HDF5 library comes out and we are hoping once that happens to migrate over to this at CSX. Not having SWMR has been a pain for XPCS....

Thanks @ajgdls and all at Diamond for this work.

Attn: @cmazolli @ambarb

MarkRivers:
> Thus I suggest that we only update ADBinaries once the final 1.10 release is out (end of March).

Do we want to test on Windows before the final 1.10 release?

MarkRivers:
Matt Newville made me aware of the following item in the HDF5 SWMR documentation:

>     The SWMR feature does not work on network file systems 
    (NFS, SMB/Windows file shares, and so on) since write 
    ordering may not be preserved on those systems.

This would make it difficult to use for us, since our detector IOCs are typically mounting the data storage systems either with NFS or SMB.  I take it Diamond does not have this problem?

stuwilkins:
We have just gone into shutdown now at NSLS-II and I was planning on testing here next week. However, we use GPFS for detector data so AFIK that does preserve write ordering so we should be good. I would be interested in what other people are using

ulrikpedersen:
That's correct @MarkRivers: in order for SWMR to work correctly you need to use a file system which preserves write ordering. The [SWMR user guide](https://www.hdfgroup.org/HDF5/docNewFeatures/SWMR/HDF5_SWMR_Users_Guide.pdf) explain this in section 1.1 and 2. [Further docs are available](https://www.hdfgroup.org/HDF5/docNewFeatures/NewFeaturesSwmrDocs.html).

At Diamond we use GPFS and Lustre - and have tested SWMR successfully on both of these systems. In addition we have also run tests on local disks with NTFS and ext4. 

The SWMR feature is a run-time optional in the NDFileHDF5 plugin (and the default is to have it disabled). You can still carry on using the plugin as before even without access to a GPFS or Lustre file system...

MarkRivers:
Can you please change this pull request to pull into the swmr_support branch in areaDetector/ADCore, rather than directly into master?  You could also create a new branch rather than using swmr_support if you want.  That will make it easier for us to test and then merge into master manually.

ajgdls:
Hi @MarkRivers I have created a branch called swmr_integration and raised the pull request #179 as you suggested.  Let me know if you need me to do anything else (I notice the branch didn't automatically merge as our hdf5_swmr branch has fallen behind master).

stuwilkins:
We have been testing this for the past weeks at CSX at NSLS-II with the fast CCD. So far no problems (on a GPFS system)

Thanks for this work! Is there any testing or other things that you would like us to contribute? We have a few weeks more before the shutdown ends and can easily run some tests.

ajgdls:
Hi @stuwilkins thanks for the testing.  Can you just provide a couple of lines to let me know the details of your tests:
Have you been turning on and off the SWMR feature, or simply leaving it on?
Have you always been monitoring the files "live"?
Have you used any other features available in the SWMR branch (shaped attribute datasets etc)?
Are you acquiring specific numbers of frames, or do you run until the stop button is pressed?
Do you acquire with additional dimensions, and attributes stored?

If you have time to test some more then that would be great.  I don't have any specific tests for you to try, I would be happy for you to try out the various options on the writer any way you see fit and then report any problems that you do find.

MarkRivers:
There is a new branch, swmr_integration, that contains the merge of the current master branch and the dls-controls:hdf5_swmr branch.  Please use that branch for testing. It allows specifying how NDAttributes of type NDAttrString are to be saved: as H5T_NATIVE_CHAR (the previous type, and still the default), or H5T_C_S1 (HDF5 strings).  I am closing this pull request and will open a new one for the swmr_integration branch once some testing has been done.

PR #149, opened 2016-03-04
exclude irrelevant header

xiaoqiangwang:
netcdf.h is irrelevant here.


ulrikpedersen:
Thanks @xiaoqiangwang - seems fine to me :+1:

PR #153, opened 2016-03-24
ADApp/Db: ran dos2unix on ADPrefixes.template, NDFileHDF5.template, N…

mp49:
…DFileNexus.template and NDTransform.template. The windows carriage returns was causing MSI to fail in the database expansion when building an IOC.


ulrikpedersen:
That's odd; I haven't had trouble with MSI on these templates...

Could these windows CR have been sneaking in through git? https://help.github.com/articles/dealing-with-line-endings/

prjemian:
that's possible if the file has a mix of EOL types: some \n, some \r\n

mp49:
Hi,

I only noticed this because I updated one of our IOCs from areaDetector 1-9 to ADCore 2-2 (so this has been there since at least 2-2). I found that the NDPluginBase template wasn’t expanded when I included the NDTransform template in the IOC substitution file. Removing the ^M fixed it, then I found a few other files with the same issue.

Cheers,
Matt

PR #156, opened 2016-03-30
adjust text boxes so font is not tiny in autoconverted .ui files

prjemian:
* [x] .adl files updated
* [x] .ui (caQtDM) files autoconverted
* [ ] .edl files autoconverted
* [x] .opi files autoconverted

note: ADExample simDetector.ui also needs this change (uses most but not all of ADCollect screen)


prjemian:
Can someone convert these two screens to .edl?

PR #158, opened 2016-03-31
Change queue size

MarkRivers:
Added the ability to change the QueueSize of a plugin at run-time.  This can be very useful, particularly for file plugins where an acquisition of N frames is overflowing the queue, but increasing the queue can fix the problem.  This will be even more useful in ADCore R3-0 where we plan to eliminate Capture mode in NDPluginFile.  Being able to increase the queue does everything that Capture mode did, but has the additional advantage that in Capture mode the NDArray memory is not allocated from the NDArrayPool, so there is no check on allocating too many arrays or too much memory.  Using the queue means that arrays are allocated from the pool, so the limits on total number of arrays and total memory defined in the constructor will be obeyed.  This can be really important in preventing system freezes if the user tries to allocate all the system memory.

### DIFF
--- a/ADApp/Db/NDPluginBase.template
+++ b/ADApp/Db/NDPluginBase.template
@@ -140,12 +140,19 @@ record(longin, "$(P)$(R)DroppedArrays_RBV")
     field(SCAN, "I/O Intr")
 }
 
-record(longin, "$(P)$(R)QueueSize")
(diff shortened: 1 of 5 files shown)


MarkRivers:
This needs more work, not ready to be pulled.

PR #159, opened 2016-04-01
Add lock in constructor before starting thread; thread takes lock before reading queue

MarkRivers:
NDPluginDriver had a basic mistake.  The constructor was accessing class data without taking the lock.  The taskThread also did not take the lock before trying to read from the queue.  I have added the appropriate calls to lock() and unlock().  
I suspect this is what was causing the crashes that Ulrik observed.  Please test this branch and see if it fixes the problem, and if so then merge the pull request.


MarkRivers:
This change, although technically correct, was not sufficient to stop the crash in the thread created by the NDPluginDriver constructor.

PR #160, opened 2016-04-01
Change queue size

MarkRivers:
Fixed logic, I think it is ready to be tested and merged.

### DIFF
--- a/ADApp/Db/NDPluginBase.template
+++ b/ADApp/Db/NDPluginBase.template
@@ -140,12 +140,19 @@ record(longin, "$(P)$(R)DroppedArrays_RBV")
     field(SCAN, "I/O Intr")
 }
 
-record(longin, "$(P)$(R)QueueSize")
+record(longout, "$(P)$(R)QueueSize")
+{
+    field(DTYP, "asynInt32")
+    field(OUT,  "@asyn($(PORT),$(ADDR),$(TIMEOUT))QUEUE_SIZE")
+    field(PINI, "YES")
+}
+
+record(longin, "$(P)$(R)QueueSize_RBV")
 {
     field(DTYP, "asynInt32")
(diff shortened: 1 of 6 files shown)


ulrikpedersen:
Starting testing.

ulrikpedersen:
Functionality question: what happens when a plugin queue is resized during processing? As it is destroyed it seems to me that all NDArrays currently waiting in the queue are silently dropped?

MarkRivers:
Yes, that is the design and it will be documented.

ulrikpedersen:
I've tested it with simdetector and it appears to work. When resizing (many many times) during acquisition it seems that the queue just gets dropped and you start over with a fresh/empty queue (as @MarkRivers confirmed above). Even if the user don't change the value but just hits enter the queue is scrapped and re-created.

I also played around with making the queue huge - beyond the available memory on my PC. I used a processing plugin with all calculations enabled and then started the simdetector runnnig at full speed (~198 fps). The simDetector NDArrayPool has been created with no max limit on the amount of memory it can allocate. It took a few minutes to run out of both RAM and swap - then Linux automatically killed the process. :grinning: Surprisingly the PC did not grind to a halt while this was happening.

In summary: it worked for me even when pushing the boundaries. Documentation should state that this is potentially dangerous:

 - input queues are dropped when resizing.
 - too large input queues can cause the application to be killed or crash  (or just really slow down when swapping) when a driver or producer plugin runs out of memory.

ulrikpedersen:
ram-test
ram-test-after

ulrikpedersen:
Ugh. Conflicts... I bet that's in NDPluginDriver.cpp with the recent merge of #163. I think the branch need to be updated with the latest changes on master.

MarkRivers:
I can fix that.  I did it yesterday on a local merge branch.

Issue #157, opened 2016-03-31
Race condition in NDPluginDriver constructor


ulrikpedersen:
Once in a while our unittests will crash with a segfault. 

A couple of examples: [build no. 229](https://travis-ci.org/areaDetector/ADCore/builds/119205876) and [build no. 218](https://travis-ci.org/areaDetector/ADCore/builds/118113158). Re-running the same travis build job will usually succeed.

The issue is not easily reproducible - at least I cannot reproduce it on RHEL6 (EPICS R3.14.12.3), but I have been able to reliably reproduce it on Ubuntu 14.04 with EPICS R3.14.12.5 and get the following stack-trace from GDB:

```
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff2788b80 (LWP 63130)]
0x00000000006888c0 in ?? ()
(gdb) bt
#0  0x00000000006888c0 in ?? ()
#1  0x00007ffff759c070 in processTask (drvPvt=0x6868d0) at ../NDPluginDriver.cpp:158
#2  0x00007ffff5dfca6f in start_routine (arg=0x68af60) at ../../../src/libCom/osi/os/posix/osdThread.c:389
#3  0x00007ffff651b182 in start_thread (arg=0x7ffff2788b80) at pthread_create.c:312
#4  0x00007ffff682b47d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111

(gdb) up
#1  0x00007ffff759c070 in processTask (drvPvt=0x6868d0) at ../NDPluginDriver.cpp:158
158	    pPvt->processTask();
(gdb) print pPvt
$1 = (NDPluginDriver *) 0x6868d0
(gdb) print pPvt->processTask
$2 = {void (NDPluginDriver * const)} 0x7ffff759c084 <NDPluginDriver::processTask()>
(gdb) print pPvt->msgQId
$3 = (epicsMessageQueueId) 0x0
```

So in [NDPluginDriver.cpp:158](https://github.com/areaDetector/ADCore/blob/master/ADApp/pluginSrc/NDPluginDriver.cpp#L158) it segfaults when dereferencing the plugin's `processTask()` function.

[...]

PR #161, opened 2016-04-03
Plugin fix clean

MarkRivers:
This seems like the correct solution.  Even if we don't fully understand why it is crashing on some systems and not on others, the fact is that the processTask thread should not be accessing the plugin object until we can be sure that all constructors have finished.  One easy way to do that is to create the thread after the final derived class is instantiated.  Changes still need to be make to ADPluginEdge (a plugin not in ADCore) and simDetectorNoIOC which instantiates objects outside the context of an IOC.


ulrikpedersen:
I've added some more complete thread startup and shutdown signalling to help fix #157.

However, I'm not entirely happy with this solution just yet so don't merge yet. I'll carry on tomorrow.

ulrikpedersen:
and travis is not happy either! :grimacing: 

Carry on tomorrow!

ulrikpedersen:
Closing this PR as #163 include the relevant commits from here.

PR #163, opened 2016-04-05
Destructor race condition issue

### LINKED ISSUE #157: Race condition in NDPluginDriver constructor

ulrikpedersen:
This fixes #157 which was originally an issue with a race condition between the NDPluginDriver constructor and a thread being started in the constructor. Eventually we also found that there was a similar issue with the destructor. The two are somewhat related and this PR include fixes for both.

- Adding a locking mechanism to the NDPluginDriver constructor and thread so that a thread does not start running until the object is fully initialized.
- Switching NDPluginDriver from using the epics thread C interface to the epicsThreadRunable C++ class. This class has a feature to wait for a thread to exit which is required in the object destructor.
- Added NDPluginDriver::start() method which starts up the thread - and wait for it to signal that it is running. This method must be called outside of the NDPluginDriver constructor.

Part of the code in this PR was developed on branch "plugin_fix_clean" and PR #161 which I will now close as deprecated.

### DIFF
--- a/ADApp/pluginSrc/NDFileHDF5.cpp
+++ b/ADApp/pluginSrc/NDFileHDF5.cpp
@@ -2738,9 +2738,9 @@ extern "C" int NDFileHDF5Configure(const char *portName, int queueSize, int bloc
 {
   // Stack Size must be a minimum of 2MB
   if (stackSize < 2097152) stackSize = 2097152;
-  new NDFileHDF5(portName, queueSize, blockingCallbacks, NDArrayPort, NDArrayAddr,
-                 priority, stackSize);
-  return(asynSuccess);
+  NDFileHDF5 *pPlugin = new NDFileHDF5(portName, queueSize, blockingCallbacks, NDArrayPort, NDArrayAddr,
+                                       priority, stackSize);
+  return pPlugin->start();
 }
 
 void NDFileHDF5::checkForOpenFile()
--- a/ADApp/pluginSrc/NDFileJPEG.cpp
+++ b/ADApp/pluginSrc/NDFileJPEG.cpp
@@ -337,9 +337,9 @@ extern "C" int NDFileJPEGConfigure(const char *portName, int queueSize, int bloc
                                    const char *NDArrayPort, int NDArrayAddr,
                                    int priority, int stackSize)
 {
-    new NDFileJPEG(portName, queueSize, blockingCallbacks, NDArrayPort, NDArrayAddr,
-                   priority, stackSize);
-    return(asynSuccess);
+    NDFileJPEG *pPlugin = new NDFileJPEG(portName, queueSize, blockingCallbacks, NDArrayPort, NDArrayAddr,
+                                         priority, stackSize);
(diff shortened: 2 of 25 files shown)


MarkRivers:
This looks OK to me.  Ready for me to merge?

ulrikpedersen:
> This looks OK to me. Ready for me to merge?

Great! Yes, please.

Issue #165, opened 2016-04-06
segmentation fault in NDPluginTimeSeries


ulrikpedersen:
As I was testing the imagej 1D change in PR #162 I stumbled upon a segmentation fault.

Setup as follows:

 - TimeSeries plugin connected to ADC simdetector. No other plugins enabled.
 - Start ADC simdetector running continuously.
 - Start TimeSeries acquisition into circular buffer with number of time points set to 20.
 - Increase number of timepoints from 20 to 1024. Crash.

I'm on my RHEL6 desktop. The ADCore code version is the current tip of the "imagej_1d! branch at rev. e29eba5. As this has just been updated with the master branch it should have all fixes for the TimeSeries plugin. 

This crash feels a little similar to the segfault I discovered in reviewing #148 - but that was fixed...

Here is a GDB stacktrace:

```
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff690df700 (LWP 20178)]
0x00007ffff7b4c82e in NDPluginTimeSeries::doTimeSeriesCallbacksT<float> (this=0x741000) at ../NDPluginTimeSeries.cpp:282
282	        signalData_[timeOut] = timeCircular[signal*numTimePoints_ + timeIn];
Missing separate debuginfos, use: debuginfo-install glibc-2.12-1.166.el6_7.7.x86_64 libgcc-4.4.7-16.el6.x86_64 libjpeg-turbo-1.2.1-3.el6_5.x86_64 libstdc++-4.4.7-16.el6.x86_64 libtiff-3.9.4-10.el6_5.x86_64 libxml2-2.7.6-20.el6_7.1.x86_64 ncurses-libs-5.7-4.20090207.el6.x86_64 nss-softokn-freebl-3.14.3-23.el6_7.x86_64 readline-6.0-4.el6.x86_64 zlib-1.2.3-29.el6.x86_64
(gdb) bt
#0  0x00007ffff7b4c82e in NDPluginTimeSeries::doTimeSeriesCallbacksT<float> (this=0x741000) at ../NDPluginTimeSeries.cpp:282
#1  0x00007ffff7b47f7e in NDPluginTimeSeries::doTimeSeriesCallbacks (this=0x741000) at ../NDPluginTimeSeries.cpp:324
#2  0x00007ffff7b48c08 in NDPluginTimeSeries::writeInt32 (this=0x741000, pasynUser=0x133ca28, value=1)
    at ../NDPluginTimeSeries.cpp:465
[...]

PR #166, opened 2016-04-07
Time series fix

MarkRivers:
I was unable to reproduce a crash by changing the number of point in the time series plugin reported in #165.  However, I was able to crash it if the number of signals (first array dimension) in the driver callback was different from MaxSignals used in the array constructor.  This pull request fixes this problem.  I tested by changing the ADCSimDetector st.cmd to have MaxSignals be 1, 4, 8, 9.   The previous value was 8, which matches the size of the array from ADCSimDetector.  It now passes the tests, while previously it crashed.


ulrikpedersen:
Works for me. Fixes #165 

The travis build failure is unrelated to this change (an issue with installing dependencies). 

Merging.

Issue #176, opened 2016-05-01
Attributes of type NDAttrString are not written as H5T_STRING in HDF5 files


MarkRivers:
NDFileHDF5 does not treat attributes of type string consistently.  Constant attributes are written with datatype H5T_C_S1 and appear as type H5T_STRING.  However, NDArray attributes of type NDAttrString are written as type H5T_NATIVE_CHAR, and appear as 2-D arrays of type H5T_STD_I8LE (on my Linux system).  This seems to me to be in violation of the HDF5 documentation, which states that H5T_NATIVE_CHAR is a numeric datatype, not intended for storing string data.

This is an example of using h5dump to display the CameraModel attribute.  Note that h5dump is displaying integer values.  This is not very user-friendly to say the least.  Note also that the NDAttrDescription _is_ an H5T_STRING, and is nicely legible.

<pre>corvette:~/scratch>h5dump --dataset=/entry/instrument/NDAttributes/CameraModel test_004.h5
HDF5 "test_004.h5" {
DATASET "/entry/instrument/NDAttributes/CameraModel" {
   DATATYPE  H5T_STD_I8LE
   DATASPACE  SIMPLE { ( 1, 256 ) / ( H5S_UNLIMITED, H5S_UNLIMITED ) }
   DATA {
   (0,0): 66, 97, 115, 105, 99, 32, 115, 105, 109, 117, 108, 97, 116, 111,
   (0,14): 114, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
   (0,35): 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
   (0,57): 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
   (0,79): 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
   (0,101): 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
[...]

PR #177, opened 2016-05-01
Allow writing NDAttributes of type NDAttrString as strings rather than char arrays

MarkRivers:
Allow writing NDAttributes of type NDAttrString as as 1-D array of type H5_C_S1[NumArrays] and size MAX_ATTRIBUTE_STRING_SIZE rather than a 2-D array of type H5T_NATIVE_CHAR[NumArrays, MAX_ATTRIBUTE_STRING_SIZE].  The default behavior is backwards compatbile, and there is a new EPICS PV that allows changing the behavior at run-time.  Fixes issue #176.


MarkRivers:
I merged this myself because I want to get ADCore R2-5 released with this enhancement

PR #179, opened 2016-05-16
Hdf5 swmr

ajgdls:
I have created this pull request from dls-controls/ADCore hdf5_swmr branch to a newly created (16th May 2016) ADCore swmr_integration branch as requested by @MarkRivers for testing prior to acceptance into ADCore.


MarkRivers:
I am working on merging this pull request.  It has gotten somewhat complex because of the recent addition to master of the new NDFileHDF5_stringAttributeDataType parameter I added to master recently.  This allows one to specify the HDF5 data type for NDAttributes of type NDAttrString.  The previous HDF5 data type for these was HDF5_NATIVE_CHAR.  This is really an 8-bit integer data type in HDF5, not a string data type.  The NDFileHDF5_stringAttributeDataType parameter keeps HDF5_NATIVE_CHAR data type as the default, but allows one to select H5T_C_S1 instead, with a fixed string length of MAX_ATTRIBUTE_STRING_SIZE.  This is much more intuitive and user-friendly, because h5dump displays the string attributes as strings, rather than integer arrays.  

The complex arises because DLS have refactored NDFileHDF5AttributeDataset into a separate class from NDFileHDF5, so from that class one cannot call getIntegerParam, for example.  So the value of the NDFileHDF5_stringAttributeDataType parameter must be passed to the constructor for the NDFileHDF5AttributeDataset class.  

This is not a big deal, and I can easily code it, but I would like to have a decision from the ADCore community on whether the NDFileHDF5_stringAttributeDataType parameter is actually needed.  Would it be acceptable to simply always write NDAttributes of type NDAttrString as H5T_C_S1 rather than HDF5_NATIVE_CHAR?  This would mean that new files would not be exactly the same as old files.  I suspect that APS and NSLS-II would be OK with this, since they don't have a large investment in the current HDF5 file format.  What about Diamond?  Is backwards compatibility required, or can we make a non-backwards compatible change for the sake of simplicity and doing it more correctly?

I am proceeding assuming backwards compatibility and the extra parameter is required, but I can simplify it if that is the consensus of the group.

MarkRivers:
I have done this merge and resolved the conflicts.  I have kept the NDFileHDF5_stringAttributeDataType parameter for now.

It compiles without warning, but I have not yet tested it.  Everyone interested in HDF5 should test this branch before it is merged into master.

ajgdls:
We had the discussion at DLS regarding the backwards compatibility and decided that it is perhaps better to not retain this backwards compatibility.  We cannot find a good reason to keep the old type and having the parameter to switch only adds more code to the already heavyweight plugin.  We accept that if removing the backwards compatibility breaks something DLS use then it is our responsibility to fix/update that ourselves.

MarkRivers:
Thanks for the quick response.  Since I have not heard a requirement for backwards compatibility from NSLS-II or APS I am assuming it is OK to unconditionally change to using H5T_C_S1.  I have committed the changes for this and pushed to the swmr_integration branch on github.  Please test ASAP.  I will plan to merge swmr_integration into master in the next few days.

stuwilkins:
Thanks, sorry @MarkRivers got sidetracked with our users meeting this week. AFIK backward compatibility is not an issue at NSLS-II because we use handlers to read files and therefore are not married to one file format. @danielballan and @tacaswell can speak much better for the facility on this. 

As for testing, @ajgdls we have been running in the SWMR mode and reading as the file is being written to. We haven't played with turning SWMR on and off for example. So far we have found no problems at all. 

I did notice that the SWMR settings are not in the ``NDFileHDF5_settings.req`` file so they are not restored on restart of the IOC. Was this intentional.

ajgdls:
Hi @stuwilkins I'll need to come back to you regarding the settings, we have always planned here to default to the original non SWMR writing of files so that people cannot accidentally write incompatible files.  Auto-saving the setting means that an IOC could be rebooted with SWMR left on.  Of course this is desirable once writing SWMR files is the norm, I'm just not sure if we should be doing it during this initial phase of roll out.
If you have any testing time left I would ask that you test the branch that @MarkRivers mentions here, and I will do the same.

danielballan:
> AFIK backward compatibility is not an issue at NSLS-II because we use handlers to read files and therefore are not married to one file format

Yes. When our users request data, they are directly given contiguous array in memory; we control how it is read, so moving between the formats will be seamless from the user POV.

stuwilkins:
@ajgdls Thanks! I hear you as to why the default is not SWMR. However I had the opposite experience where I had selected SWMR as an option then restarted the IOC and wrote files not in SWMR not realizing that it wasn't in save-restore. I think this is a policy question: If the settings of the HDF5 plugin are in save/restore then I would personally prefer if they are all in. After all, someone would still have to turn on SWMR to get it to write SWMR files in the first place. Thoughts?

decarlof:
@MarkRivers from my side there is no need to maintain the back compatibility and is perfectly fine to always write NDAttrString as H5T_C_S1. I will check at the APS and get back if other have issues but  I doubt since as far as I know I am the only one using this plugin in production.

PR #185, opened 2016-06-09
copy hdf5 dlls if dynamically built

xiaoqiangwang:
For a multiple detector system, I have to use the dynamical library on windows-x64. I found it to be convenient to copy the dlls to the IOC bin directory. 

But I also realize one can set the PATH to include $(ADBINARIES)/bin/windows-x64. I create this pull request in case you agree on this idea.


MarkRivers:
EPICS base automatically builds a Windows command script, dllPath.bat, to set the PATH to include all of the required directories for the DLLs.  For example for ADLightField the iocLightField/Makefile creates the following dllPath.bat file:

> PATH %PATH%;J:/epics/support/areaDetector-2-5/ADLightField/iocs/lightFieldIOC/bin/windows-x64;J:/epics/support/areaDetector-2-5/ADLightField/iocs/lightFieldIOC/../../bin/wind
ows-x64;H:/epics/base-3.14.12.5/bin/windows-x64;J:/epics/support/asyn-4-29/bin/windows-x64;J:/epics/support/areaDetector-2-5/ADBinaries/bin/windows-x64;J:/epics/support/areaD
etector-2-5/ADCore/bin/windows-x64;J:/epics/support/autosave-5-7-1/bin/windows-x64;J:/epics/support/busy-1-6-1/bin/windows-x64;J:/epics/support/calc-3-6-1/bin/windows-x64;J:/
epics/support/seq-2-2-3/bin/windows-x64;J:/epics/support/sscan-2-10-2/bin/windows-x64;J:/epics/support/devIocStats-3-1-14/bin/windows-x64

This seems preferable to me to copying the DLLs to each IOC bin directory.  One of the advantages of DLLs is that the application does not need to be rebuilt if a backwards-compatible change is made to a DLL.  But this change would require running make in every IOC directory if the ADBinaries DLLs changed in a backwards-compatible way.

xiaoqiangwang:
Thanks for the info about dllPath.bat. That fits perfectly. I withdraw this request.

PR #205, opened 2016-10-10
Fixes v2

hdante:


MarkRivers:
Implemented these fixes.  Slightly different format for some files.

PR #209, opened 2016-10-17
NDPluginDriver: Avoid creating the callback thread if blocking-callbacks are enabled at initialization.

ambrop72:
If blocking-callbacks are enabled in the port initialization, the callback thread will not be created, saving memory. In this case it is not allowed to disable blocking callbacks at runtime later (to allow runtime selection, blocking-callbacks must be set to 0 at initialization, so that the thread is created).

### DIFF
--- a/ADApp/pluginSrc/NDPluginDriver.cpp
+++ b/ADApp/pluginSrc/NDPluginDriver.cpp
@@ -18,6 +18,7 @@
 #include <epicsThread.h>
 #include <epicsEvent.h>
 #include <epicsTime.h>
+#include <cantProceed.h>
 
 #include <asynDriver.h>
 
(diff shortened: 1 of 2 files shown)


ambrop72:
Hi Mark,
Did you forget about this?
I see the CI is saying "check failed" but it seems to be unrelated to the change.

brunoseivam:
Is the callback thread creation causing excessive memory consumption? Otherwise we have to weight losing flexibility in favor of a small gain

MarkRivers:
I would prefer a slightly different solution.
- If the constructor is called with blocking callbacks enabled then don't create the thread
- If the user later selects blocking callbacks disabled then create the thread
- This should just involve moving the code for the thread creation into a separate method that can either be called in the constructor or in writeInt32.
If you are worried about memory consumption then it is your responsibility to ensure that the control to disable blocking callbacks is hidden from the user.

Have you actually analysed to see why each thread is consuming 70MB?  This seems like it must be due to something that can be changed, like a stack size or something.  Certainly on vxWorks each of these plugin threads is only using a few kB.

ambrop72:
Okay I will change it to start the thread on demand if you think this is best.

We've tried to start the thread with a very small stack size (16K) but it only produced a small improvement (several MB less) in the use of virtual memory (using Linux). Do you have any idea what else could be tried?

MarkRivers:
I don't know anything about locking memory and virtual memory size of threads.  But I suspect there are people on tech-talk who do so I would post your problem there.

ambrop72:
Hey Mark, I've modified the change to start the thread on demand as you recommended.

PR #215, opened 2016-11-04
Allow datatype to be set at iocInit

alex-soderqvist:
I think it is good practice to allow user to set reasonable default values at iocInit. I know that most people probably use autosave and if available this fulfill their needs. However when writing startup scripts without autosave infrastructure this can greatly speed up the process of testing.

In principal I think that all setters should be macrofied, and default values should be removed from the cpp code. Although this a much bigger chunk of work and you have to know about implications this can have. In my opinion datatype is an essential setting which is not likely to change during run-time, hence it makes sense having a macro to allow it to be set once at iocInit. Currently this is the only one I am missing, but it's likely I will run in to more.

What is your take on this?

### DIFF
--- a/ADApp/ADSrc/asynNDArrayDriver.cpp
+++ b/ADApp/ADSrc/asynNDArrayDriver.cpp
@@ -675,7 +675,6 @@ asynNDArrayDriver::asynNDArrayDriver(const char *portName, int maxAddr, int numP
     setIntegerParam(NDArraySizeZ,   0);
     setIntegerParam(NDArraySize,    0);
     setIntegerParam(NDNDimensions,  0);
-    setIntegerParam(NDDataType,     NDUInt8);
     setIntegerParam(NDColorMode,    NDColorModeMono);
     setIntegerParam(NDUniqueId,     0);
     setDoubleParam (NDTimeStamp,    0.);
--- a/ADApp/Db/NDArrayBase.template
+++ b/ADApp/Db/NDArrayBase.template
@@ -168,6 +168,7 @@ record(mbbo, "$(P)$(R)DataType")
    field(SXVL, "6")
    field(SVST, "Float64")
    field(SVVL, "7")
+   field(VAL,  "$(DATATYPE=0)")
    info(autosaveFields, "VAL")
 }
 


ulrikpedersen:
> I know that most people probably use autosave and if available this fulfill their needs.

The initialisation of parameters is a discussion that have come up a few times here at Diamond... We don't use autosave on **all** IOCs. We have had some complaints from areaDetector users that the startup can be very noisy as drivers/plugins load NDAttribute XML definitions at startup - but many parameters are uninitialised and the result is an error message for each uninitialised parameter that is defined in the XML.

In the small example case this is not a big deal. But in some of our diagnostic camera IOCs (aravis GigE) we have maybe 10 drivers instantiated and each driver has a chain of multiple plugins: at least one of each of the common ones plus a handful of extra ROI plugins... Then the error messages are a significant problem and slows down the startup of the IOC.

> In principal I think that all setters should be macrofied, and default values should be removed from the cpp code. Although this a much bigger chunk of work and you have to know about implications this can have. In my opinion datatype is an essential setting which is not likely to change during run-time, hence it makes sense having a macro to allow it to be set once at iocInit. Currently this is the only one I am missing, but it's likely I will run in to more.

> What is your take on this?

I would also like to find a solution where parameter default values can be initialised without a dependency on autosave. It still have to work with autosave of course as this is a big use-case for many.

alex-soderqvist:
I have another slightly different example taken from adprosilica driver. Where for example binning parameters are read from hardware and used as startup parameters. I assume that this is a systematic approach covering all parameters that are supported by the hardware, and covering many drivers.
 
The problem as I see is that there is no alignment in this behavior, you basically have to check the code for each parameter to know the behavior.

The hardware example here raises a wider, more of a conceptual question. Where do you want your reference settings to be stored? My opinion is that the camera hardware itself **shouldn't** be the reference point.

I would also in this case prefer if the settings were user settable with a macro at iocInit. The overall reason for this is that, in my opinion, it should be possible to replicate exactly the same hardware setup run exactly the same IOC with startup script and get exactly the same settings at startup.

MarkRivers:
> The hardware example here raises a wider, more of a conceptual question. Where do you want your reference settings to be stored? My opinion is that the camera hardware itself shouldn't be the reference point.

This is not an easy question.  Sometimes the user may want to be able to restart the IOC without changing the current camera settings at all.  In this case reading the current settings from the hardware may be desired.  On the other hand sometimes the user may want the saved EPICS state to be pushed to the camera every time the IOC starts.

Note that if autosave is used then the fact that the Prosilica driver reads from the hardware has no effect.  Those readback settings will be overridden by the autosave values.

alex-soderqvist:
> Note that if autosave is used then the fact that the Prosilica driver reads from the hardware has no effect. Those readback settings will be overridden by the autosave values.
In this case wouldn't hardware values and autosave values be the same? Effectively eliminating the need for reading from hardware. Although I must admit I don't know how autosave works, this may toggle the parameters between different values.

ulrikpedersen:
> Sometimes the user may want to be able to restart the IOC without changing the current camera settings at all. In this case reading the current settings from the hardware may be desired. On the other hand sometimes the user may want the saved EPICS state to be pushed to the camera every time the IOC starts.

That's right. The most basic functionality of a control system must be to reflect the state of the hardware devices. As an additional feature you can enable autosave/restore (or burt or similar) to restore a device to a certain previous state. It is important that we can do both (or either).

My comment is more directed at an issue with uninitialized values - i.e. (asyn) parameters that have no direct hardware parameters to read from at startup and thus are left uninitialized and causing errors... I would like to see a mechanism for initializing these parameters, in particular in Plugins as these are typically software-only parameters with no direct hardware equivalent.

MarkRivers:
>My comment is more directed at an issue with uninitialized values - i.e. (asyn) parameters that have no direct hardware parameters to read from at startup and thus are left uninitialized and causing errors... I would like to see a mechanism for initializing these parameters, in particular in Plugins as these are typically software-only parameters with no direct hardware equivalent.

I agree this is important.  There a several ways it can be done.  The first is to have the constructor initialize the value.  This will result in the EPICS output record obtaining this value at iocInit, via the readback mechanism in asyn device support.  It has the disadvantage that any value that is assigned to the record in the database before iocInit will be lost, because of the readback mechanism in asyn device support.  So I think it is probably better for the driver to do almost no initialization in the constructor, and have the values come either from the database (first), autosave (second), or dbpf after iocInit (third).

MarkRivers:
>>Note that if autosave is used then the fact that the Prosilica driver reads from the hardware has no effect. Those readback settings will be overridden by the autosave values.

> In this case wouldn't hardware values and autosave values be the same? Effectively eliminating the need for reading from hardware. Although I must admit I don't know how autosave works, this may toggle the parameters between different values.

No, that is definitely not always the case.  What if the camera had been power-cycled, or had been controlled by some other software like the vendor's control program? Then the autosave file and the camera values would be different. It is also possible to use an autosave file that was not the one written the last time the IOC shut down, but rather one from some earlier period in time.    This allows users to save their settings from one run, and restore them at some future time, even if the IOC has been run with other settings in the mean time.

MarkRivers:
> Although I must admit I don't know how autosave works, this may toggle the parameters between different values.

Output records with PINI=YES will be written only once at iocInit.  The value written will be determined by the following algorithm in this order.

- The value defined in the record's .dbd file.
- The value defined in the database
- The value read from the hardware, assuming that the initial read operation to the driver returned asynSuccess for this parameter
- The value from autosave

Finally, a dbpf from the startup script after iocInit.

alex-soderqvist:
> So I think it is probably better for the driver to do almost no initialization in the constructor, and have the values come either from the database (first), autosave (second), or dbpf after iocInit (third)

Great, so we agree on the on the main topic and reason for this pull request.

> What if the camera had been power-cycled, or had been controlled by some other software like the vendor's control program?

True, I buy the argument if a user changed the camera settings from another piece of software. Although, if the camera is power cycled I don't see why you would like the default settings. At least I wouldn't like that, although maybe other people use this as a feature so that they effectively can reset the settings by powercycling the camera. Anyhow the first argument is strong enough to shup me up in this matter.

ulrikpedersen:
>> So I think it is probably better for the driver to do almost no initialization in the constructor, and have the values come either from the database (first), autosave (second), or dbpf after iocInit (third)

> Great, so we agree on the on the main topic and reason for this pull request.

So with this in mind, could we begin to extend this and put default values in databases (I think that's missing in a lot of places right now) and start removing constructor initializations?

Should we treat drivers and plugins differently? Drivers typically control hardware with state - plugins do not...

MarkRivers:
One way to handle this would be to implement a new option in asynPortDriver.  It could have a flag that controls whether any read of the parameter library before interruptAccept=true returns asynError, even if the parameter has been previously written with writeXXXParam by the driver.  

This has the advantage that it will allow the constructor to initialize to reasonable defaults, but prevent those values from getting pushed into the records, overriding database defaults.  In parallel one can add defaults and initialization with optional macros to the databases.  

The advantage of having the constructor initialize to reasonable defaults is making things easier for unit tests and standalone C++ programs not running in the context of an IOC.

alex-soderqvist:
> The advantage of having the constructor initialize to reasonable defaults is making things easier for unit tests and standalone C++ programs not running in the context of an IOC.

In my opinion, making tests depend on default values in the piece of code you are actually testing is conceptually wrong. Changing the default values then potentially breaks the test. I would have designed the tests so that the are completely independent of any default values. Which also makes sense if you take into account the drivers, which takes startup values from hardware. It would be more coherent and safe to specify all parameters in the test. Also if you just look at the test code, and what parameters is set you can't actually foresee the outcome, because they depend on the code that they are testing. In principal you should be able to write the tests before the actual code you are testing are written.

But our view on tests may differ.

PR #218, opened 2016-11-08
ROI collapse dims

### LINKED ISSUE #217: NDPluginROI should have the ability to collapse dimensions of size 1
If the NDPluginROI plugin is configured to select just one element on a particular dimension then the resulting output arrays will have a superfluous dimension of 1.  For example if the input array to the ROI plugin has dimensions [4, 256, 256] and the plugin is configured with MinX=1, MaxX=1, MinY=0, MaxY=255, MinZ=0, MaxZ=255 then the output arrays will be of size [1, 256, 256].  This is very inconvenient for some purposes.  For example file plugins like JPEG or TIFF will not work with arrays of this size, nor will the ImageJ display plugin.  They all require that the array have dimensions [256, 256].

What is needed is the ability of the ROI plugin to optionally collapse dimensions of size 1.  When the collapse is done the dimension attributes (size, offset, binning, reverse) must also be copied.

MarkRivers:
Implements new CollapseDims logic in NDPluginROI.  Fixes #217 .


ulrikpedersen:
That's a good idea 👍 

The code *looks* right to me but I wish you'd add a unittest case to demonstrate :-)

MarkRivers:
I do have a Python unit test in [NDDriverStdArrays](https://github.com/areaDetector/NDDriverStdArrays/tree/master/NDDriverStdArraysApp/src/Python_tests) which I tested with yesterday.  However, I need to learn how to build a unit test in ADApp/pluginTests, so I will work on that today.  Please wait to merge this until I finish the unit test.

MarkRivers:
I have finished the unit tests for the ROI plugin.  That was a good idea, because it uncovered bugs not only my new code for CollapseDims, but also in the old code when the input array had ndims=3 and dims[2]=1.  It was always collapsing that dimension, which is not correct.  It should only collapse that dimension if the input array has ColorMode=RGB3.  This is then consistent with how RGB1 and RGB2 are handled.

MarkRivers:
@ulrikpedersen any chance you could look at this today?  I'd like to get it into master for someone who is testing NDDriverStdArrays.

PR #220, opened 2016-11-09
Add support for detector serial number.

hinxx:
When working with several detectors of the same model in single IOC it is useful to see a serial number of the detector in the OPI.
See ADExample, ADProsilica and ADAndor pull requests that show the use of this feature.

Note that ADCSimDetector can not use this since it derives asynNDArrayDriver instead of ADDriver. Would it be better to move serial number support to asynNDArrayDriver? If so, maybe manufacturer and model are also candidates for migration to asynNDArrayDriver.

### DIFF
--- a/ADApp/ADSrc/ADDriver.cpp
+++ b/ADApp/ADSrc/ADDriver.cpp
@@ -143,6 +143,7 @@ ADDriver::ADDriver(const char *portName, int maxAddr, int numParams, int maxBuff
     createParam(ADStatusMessageString,       asynParamOctet, &ADStatusMessage);
     createParam(ADStringToServerString,      asynParamOctet, &ADStringToServer);
     createParam(ADStringFromServerString,    asynParamOctet, &ADStringFromServer);    
+    createParam(ADSerialNumberString,        asynParamOctet, &ADSerialNumber);
 
     /* Here we set the values of read-only parameters and of read/write parameters that cannot
      * or should not get their values from the database.  Note that values set here will override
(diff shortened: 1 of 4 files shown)


MarkRivers:
I have been thinking for some time that we need to add something like the following to the ADDriver base class:

SDKVersion The version number of the vendor's SDK
FirmwareVersion The firmware version in the detector
SerialNumber What you have added here
ADDriverVersion The version of this AD driver

This should go in the asynNDArrayDriver base class
ADCoreVersion The version of ADCore that the driver or plugin was built with

These have been added on an ad-hoc basis to a number of drivers in their own classes. It seems like a good idea to put them in the base classes, and into the OPI screens. They can be initialized to reasonable defaults.  

I think we should try to assemble a list like the above and do this once, rather than in a piecemeal fashion to minimize work on both the base class and derived class drivers.

MarkRivers:
I notice that the OPI file that you modified is the one in the op/opi/autoconvert/ directory.  This is not the right place to modify it: that directory is intended only for files that are autoconverted from medm, so they should not be manually edited.  If you have a manually edited file then it should be placed in the overlying op/opi/ directory.

hinxx:
I can do this and test it for the ADAndor, ADProsilica and aravisGigE and submit new pull request, in a separate branch.
I'll also move the manually changed OPI to proper location.

MarkRivers:
I am closing this pull request in anticipation of a new one from @hinxx that includes more parameters.

PR #221, opened 2016-11-09
startPVAServer will be called only once. Fixes #219

brunoseivam:
A new static member in `NDPluginPva` keeps track if `startPVAServer` was already called by another instance and prevents calling it more than once for the process.

PR #222, opened 2016-11-09
startPVAServer won't be called by the plugin anymore. Fixes #219

### LINKED ISSUE #219: Multiple instances of NDPluginPva call startPVAServer too many times
`startPVAServer` should be called just once in each process. From tech-talk:

Hi all,

I am at ESS running an IOC with 2 pva plugins both connected to the simDetector:

```
# Create a simDetector driver
# simDetectorConfig(const char *portName, int maxSizeX, int maxSizeY, int dataType,
#                   int maxBuffers, int maxMemory, int priority, int stackSize)
simDetectorConfig("DMSCSIMDET", 500, 500, 1, 0, 0)
dbLoadRecords("simDetector.template", "P=DMSC:, R=CAM-, PORT=DMSCSIMDET, ADDR=0, TIMEOUT=1")
#create pva plugin
NDPvaConfigure(DMSCPVA, 20, 0, "DMSCSIMDET", 0, "DMSC:IMAGE", 0)
VERSION : pvAccess Server v4.1.2
PROVIDER_NAMES : local
BEACON_ADDR_LIST :
AUTO_BEACON_ADDR_LIST : 1
BEACON_PERIOD : 15
BROADCAST_PORT : 5076
SERVER_PORT : 5075
RCV_BUFFER_SIZE : 16384
IGNORE_ADDR_LIST:
STATE : INITIALIZED
dbLoadRecords("NDPva.template", "P=DMSC:, R=PVA-, PORT=DMSCPVA, ADDR=0, TIMEOUT=1, NDARRAY_PORT=DMSCSIMDET")
#create pva plugin
NDPvaConfigure(DMSCPVA2, 20, 0, "DMSCSIMDET", 0, "DMSC:IMAGE2", 0)
2016-11-08T16:33:54.789 Using dynamically assigned TCP port 52363.
VERSION : pvAccess Server v4.1.2
PROVIDER_NAMES : local
BEACON_ADDR_LIST :
AUTO_BEACON_ADDR_LIST : 1
BEACON_PERIOD : 15
BROADCAST_PORT : 5076
SERVER_PORT : 52363
RCV_BUFFER_SIZE : 16384
IGNORE_ADDR_LIST:
STATE : INITIALIZED
dbLoadRecords("NDPva.template", "P=DMSC:, R=PVA2-, PORT=DMSCPVA2, ADDR=0, TIMEOUT=1, NDARRAY_PORT=DMSCSIMDET")
```

As you can see it start 2 pvAccess servers.

Monitoring one array sees multiple instance of the pv:
```
pvget -m DMSC:IMAGE2
[ChannelRequesterImpl] message(More than one channel with name 'DMSC:IMAGE2' detected, connected to: 194.47.240.7:52834, ignored: 194.47.240.7:5075, warning)
DMSC:IMAGE2
structure
    union value
        (none)
```

Listing the servers shows the 2 pvAccess servers (Which yo):
```
pvlist
GUID 0x7EE8215800000000799BA728, version 1: tcp@[x.x.x.x:5075]
GUID 0x7EE82158000000008AD62829, version 1: tcp@[x.x.x.x:52834]
```

Listing pvs on each server shows that both are publishing the same PVs.
```
pvlist 0x7EE8215800000000799BA728
DMSC:IMAGE
DMSC:IMAGE2

pvlist 0x7EE82158000000008AD62829
DMSC:IMAGE
DMSC:IMAGE2

```
Has anyone tried running 2 NDPluginPvas in the same IOC before? Are you supposed to start multiple pvAccess servers in the same IOC?

Note that we are using our own build system, and it might be an issue there.

Regards,
Alexander Söderqvist

brunoseivam:
This is an alternative solution to #221. With this change the plugin never calls `startPVAServer`, and it is left to the user to actually call it after `iocInit()` in the startup script. The documentation was updated to reflect this change. The `dbd` file that registers `startPVAServer` is included in the `commonDriverMakefile` file.

### DIFF
--- a/ADApp/commonDriverMakefile
+++ b/ADApp/commonDriverMakefile
@@ -13,6 +13,7 @@ $(PROD_NAME)_DBD += NDFileNull.dbd
 
 ifeq ($(WITH_EPICS_V4),YES)
   $(PROD_NAME)_DBD += NDPluginPva.dbd
+  $(PROD_NAME)_DBD += PVAServerRegister.dbd
   PROD_LIBS += ntndArrayConverter
   PROD_LIBS += nt
   PROD_LIBS += pvDatabase
--- a/ADApp/pluginSrc/NDPluginPva.cpp
+++ b/ADApp/pluginSrc/NDPluginPva.cpp
@@ -164,8 +164,6 @@ NDPluginPva::NDPluginPva(const char *portName, int queueSize,
 
     if(!master->addRecord(m_record))
         throw runtime_error("couldn't add record to master database");
-
-    m_server = startPVAServer(PVACCESS_ALL_PROVIDERS, 0, true, true);
 }
 
 /* Configuration routine.  Called directly, or from the iocsh function */
--- a/ADApp/pluginSrc/NDPluginPva.h
+++ b/ADApp/pluginSrc/NDPluginPva.h
@@ -34,7 +34,6 @@ protected:
     int NDPluginPvaPvName;
 
 private:
-    epics::pvAccess::ServerContext::shared_pointer m_server;
     NTNDArrayRecordPtr m_record;
 };
 
--- a/documentation/NDPluginPva.html
+++ b/documentation/NDPluginPva.html
@@ -101,7 +101,7 @@
   <h2 id="PVAServer">Starting the pvAccess server</h2>
   <p>
     In order to actually serve the EPICSv4 PV created by this plugin it is necessary to call 
-    <pre>startPVAServer</pre> after <pre>iocInit</pre> in the IOC startup script.
+    <pre>startPVAServer</pre>. 
   </p>
 </body>
 </html>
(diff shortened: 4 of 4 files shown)


MarkRivers:
Why is it required that the call to startPVAServer be after the call to iocInit. What if we want to test outside the context of an IOC where iocInit is never called?

brunoseivam:
I have been assuming it is required since that's what is in the [example](http://epics-pvdata.sourceforge.net/gettingStarted.html) mentioned by Alexander Söderqvist. Come to think of it, `startPVAServer` was being called before `iocInit` by the plugin before, so I see no harm. I will edit the documentation accordingly.

alex-soderqvist:
It seems that it doesn't matter when you actually start the PVAServer, although you should probably ask a V4 developer. I only pointed to the only documentation I know.

MarkRivers:
I added a commented out line to call startPVAServer in ADCore/iocBoot/commonPlugins.cmd.  This should be uncommented if the NDPva plugin is started there.

Issue #211, opened 2016-11-01
Add ellipsoid shape to NDPluginOverlay


MarkRivers:
NDPluginOverlay would benefit from the addition of an ellipsoid shape option.

PR #226, opened 2016-11-26
Overlay enhancements

MarkRivers:
This addresses issue #211.


ADApp/pluginTests/test_NDPluginOverlay.cpp
```diff
@@ -0,0 +1,255 @@
...
+    BOOST_REQUIRE_EQUAL(pStr->arrayDims.size(), pStr->pArrays[0]->ndims);
+  }
+}
+
+
+BOOST_AUTO_TEST_CASE(basic_overlay_operation)
+{
+
+  overlayTestCaseStr *pStr = &overlayTestCaseStrs[0];
+  for (size_t i=0; i<overlayTestCaseStrs.size(); i++, pStr++)  {
```
ulrikpedersen: I would probably use an iterator here to loop through the vector as a general rule of best practice when using a std container... In this case however the result is the same.

ADApp/pluginTests/test_NDPluginOverlay.cpp
```diff
@@ -0,0 +1,255 @@
...
+    BOOST_CHECK_NO_THROW(Overlay->write(NDPluginOverlayDrawModeString,  pStr->drawMode));
+    BOOST_CHECK_NO_THROW(Overlay->write(NDPluginOverlayRedString,       pStr->red));
+    BOOST_CHECK_NO_THROW(Overlay->write(NDPluginOverlayGreenString,     pStr->green));
+    BOOST_CHECK_NO_THROW(Overlay->write(NDPluginOverlayBlueString,      pStr->blue));
+
+    BOOST_CHECK_NO_THROW(Overlay->write(NDArrayCallbacksString, 1));
+    BOOST_CHECK_EQUAL(Overlay->readInt(NDArrayCallbacksString), 1);
+
+    Overlay->lock();
+    BOOST_CHECK_NO_THROW(Overlay->processCallbacks(pStr->pArrays[0]));
```
ulrikpedersen: Am I right to understand that the only real test here is whether the plugin processes the frame and the downstream plugin input counter increments?

I suppose generating test images and checking the output pixel-by-pixel would be a bit of an overhead... It might be made easier with an extension to the test utilities but that's another development!

ADApp/pluginTests/test_NDPluginOverlay.cpp
ulrikpedersen: I get why you're doing the loop over the test cases, but I can't help but feel a bit uneasy about cramming all logical test cases into a single boost test case. As the test cases are essentially the same but with different input sets I wonder if there is a more appropriate way to create multiple boost test cases from a list of input.

ADApp/pluginTests/test_NDPluginOverlay.cpp
MarkRivers: Yes, I am not checking the actual output image, but only that corner cases like sizes of 0 or larger than the image dimensions don't cause it to crash.  I wanted to add at least a basic unit test for the overlay plugin.  More thorough testing is for the future ...

ADApp/pluginTests/test_NDPluginOverlay.cpp
MarkRivers: Regarding the indents in NDPluginOverlay.cpp it was getting deeply nested code so I changed the indents from 4 to 2.  It would be nice if github let one set the -w -b options for git diff.  You can checkout and do it on the command line of course.

ADApp/pluginTests/test_NDPluginOverlay.cpp
ulrikpedersen: Hi Mark! Yeah for future, I think we could consider adding a few utilities to aid checking/testing output images. For the one-off plugin-test it's too much work to setup. 

For example an NDArray equal operator (pixel-by-pixel) or equivalent function could be useful.

ADApp/pluginTests/test_NDPluginOverlay.cpp
ulrikpedersen: A quick google yields some result: http://www.boost.org/doc/libs/1_62_0/libs/test/doc/html/boost_test/tests_organization/test_cases/test_case_generation.html

I have not tried to use the BOOST_DATA_TEST_CASE yet but at a glance this looks like it might help with the case of testing a set of different input variables. I wouldn't hold up this particular PR with a change like that though!

MarkRivers: > See the few comments I added inline in the test code. Mostly I would like to find a better way to    create multiple boost unittest cases from a list of varying input data.

Yes, I agree.  I found I could not initialize structs with std::vector conveniently, so I needed to create both a different struct with conventional arrays to do that and then copy to the one with std::vector.  I do like being able to add a new test case with 1 line of code, I use that for the ROI plugin testing as well.

I'm happy for you to do the merge.

ulrikpedersen: > I found I could not initialize structs with std::vector conveniently, so I needed to create both a different struct with conventional arrays to do that and then copy to the one with std::vector. 

That is a bit of a nuisance with std vectors and containers but improved in C++11 so will eventually get easier :-)

> I do like being able to add a new test case with 1 line of code, I use that for the ROI plugin testing as well.

I agree. A combination of a clever initialization and BOOST_DATA_TEST_CASE I think we we should be able to get to that point at least for tests that just check for no throw.

> I'm happy for you to do the merge.

OK. Will do.

PR #229, opened 2016-12-06
Remove max attribute string size

MarkRivers:
NDAttribute.h previously defined the constant MAX_ATTRIBUTE_STRING_SIZE to be 256.

This is not desirable because it implies that NDAttributes of type NDAttrString have a maximum size.  They do not, and this constant should not be defined in this file.

What is true is that some plugins and drivers may want or need to impose a maximum size on attribute strings.  If so they should define these sizes locally, since there is no need for all plugins and drivers to use the same limits.

I have changed all files in areaDetector to locally define MAX_ATTRIBUTE_STRING_SIZE (or a constant of another name) rather than to use the value from NDAttribute.h.

This was motivated by the following:
- The TIFF files that the Pilatus writes have a very useful TIFF tag called TIFFTAG_IMAGEDESCRIPTION.  camserver puts a lot of important information as a very long string in that tag.  It can be viewed with tiffinfo.  The string is much longer than the previous limit of 256 characters.

- I changed the ADPilatus driver to add that string as an NDAttribute to the image.  I then want file plugins to be able to save that string in other file formats, including new TIFF files written by areaDetector.  This required removing the 256 character limitation in the TIFF plugin.

I have also modified the NDFileTIFF plugin to look for the NDAttribute TIFFImageDescription.  If it is found it writes that string into the tag TIFFTAG_IMAGEDESCRIPTION in the new tiff file.


mp49:
Would it also be useful to add a new parameter to ADBase to be able to set an image description, that can either be written into a camera driver or set as an attribute for a file writing plugin? Or just let a camera driver implement this if it needs to?

ulrikpedersen:
Just to let you know that I am aware of this one but haven't been able to get to review it properly yet. Hopefully next week... I hope I'm not holding up progress!

MarkRivers:
If you could do it this week that would be great, thanks,

ulrikpedersen:
The changes give some extra flexibility at the cost of slightly more complexity... 

It would be nice to move to std::string types and not worry about length of strings (although some plugins may impose limits for their internal use). I think this is what you're referring to in you comment on epics-modules/asyn#35

prjemian:
This level of unit test is a great example of "start small and expand".  I recommend adding it.  In my limited experience, payback on unit testing comes much faster than one expects.

ulrikpedersen:
I agree @prjemian - I'm just not pushing it hard as I know that I don't personally have the time to sort it out right now ;-)

PR #234, opened 2017-01-04
Change NDAttributes to use std::string

MarkRivers:
Changed NDAttribute files to use std::string for most of the string variables rather than char *.  Added new methods to set and get attributes of type NDAttrString using std::string rather than char *.

Removed the length limitation of MAX_ATTRIBUTE_STRING_SIZE in paramAttribute.cpp.  

Added trailing underscores to all private member variables for clarity.

These changes requiring using the master branch of asyn until R4-31 is released.  I have changed the files in ci/ to use git to get that branch and call it R4-31.

### DIFF
--- a/ADApp/ADSrc/NDAttribute.h
+++ b/ADApp/ADSrc/NDAttribute.h
@@ -9,6 +9,8 @@
 #ifndef NDAttribute_H
 #define NDAttribute_H
 
+#include <string>
+
 #include <stdio.h>
 #include <string.h>
 
@@ -98,8 +100,10 @@ public:
     virtual NDAttrDataType_t getDataType();
     virtual int getValueInfo(NDAttrDataType_t *pDataType, size_t *pDataSize);
     virtual int getValue(NDAttrDataType_t dataType, void *pValue, size_t dataSize=0);
+    virtual int getValue(std::string& value);
     virtual int setDataType(NDAttrDataType_t dataType);
-    virtual int setValue(void *pValue);
+    virtual int setValue(const void *pValue);
+    virtual int setValue(const std::string&);
     virtual int updateValue();
     virtual int report(FILE *fp, int details);
     friend class NDArray;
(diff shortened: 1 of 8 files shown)


ADApp/ADSrc/NDAttribute.h
```diff
@@ -98,8 +100,10 @@ class epicsShareClass NDAttribute {
     virtual NDAttrDataType_t getDataType();
     virtual int getValueInfo(NDAttrDataType_t *pDataType, size_t *pDataSize);
     virtual int getValue(NDAttrDataType_t dataType, void *pValue, size_t dataSize=0);
+    virtual int getValue(std::string& value);
     virtual int setDataType(NDAttrDataType_t dataType);
-    virtual int setValue(void *pValue);
+    virtual int setValue(const void *pValue);
```
ulrikpedersen: Constness :+1:

ci/build-synapps-deps.sh
```diff
@@ -5,11 +5,12 @@ set -e
...
 cd external
 
-wget -nv https://github.com/epics-modules/asyn/archive/R4-26.tar.gz
-tar -zxf R4-26.tar.gz
-echo "EPICS_BASE=/usr/lib/epics" > asyn-R4-26/configure/RELEASE
-#echo "EPICS_LIBCOM_ONLY=YES" >> asyn-R4-26/configure/CONFIG_SITE
-make -C asyn-R4-26/
+#wget -nv https://github.com/epics-modules/asyn/archive/R4-26.tar.gz
+#tar -zxf R4-26.tar.gz
+git clone https://github.com/epics-modules/asyn asyn-R4-31
```
mdavidsaver: A simpler, and less github specific, alternative is to clone with ```--depth 10``` (or a similarly small number).  This can be combined with ```--branch <name>```.

wrt. travis-ci, this leds naturally to setting the branch name with an environment variable.  As for example https://github.com/epics-modules/devlib2/blob/master/.travis.yml where I build against several different Base versions.

PR #235, opened 2017-01-16
Extend centroid calculations with image moments up to the fourth order

alex-soderqvist:
I have implemented 3rd and 4th order image moments to calculate skewness and kurtosis according to:
https://en.wikipedia.org/wiki/Image_moment
I have also added eccentricity and orientation according to: https://imagej.nih.gov/ij/plugins/download/Moment_Calculator.java
Specifically these lines:
```
// Compute Orientation and Eccentricity
// source: Awcock, G.J., 1995, "Applied Image Processing", pp. 162-165
orientation = 0.5*Math.atan2((2.0*m11),(m20-m02));
orientation = orientation*180./Math.PI; //convert from radians to degrees
eccentricity = (Math.pow((m20-m02),2.0)+(4.0*m11*m11))/m00;
```

This is requirements for a profile monitor here at ESS, to avoid forking or code duplication, by putting it in a separate plugin, it would be nice to merge this.

### DIFF
--- a/ADApp/Db/NDStats.template
+++ b/ADApp/Db/NDStats.template
@@ -211,11 +211,19 @@ record(ao, "$(P)$(R)CentroidThreshold")
     field(VAL,  "1")
     info(autosaveFields, "VAL")
 }
+
 record(ai, "$(P)$(R)CentroidThreshold_RBV")
 {
     field(DTYP, "asynFloat64")
     field(INP,  "@asyn($(PORT),$(ADDR),$(TIMEOUT))CENTROID_THRESHOLD")
     field(VAL,  "0")
+    field(SCAN, "I/O Intr")
+}
+
+record(ai, "$(P)$(R)CentroidTotal_RBV")
+{
+    field(DTYP, "asynFloat64")
+    field(INP,  "@asyn($(PORT),$(ADDR),$(TIMEOUT))CENTROID_TOTAL")
     field(SCAN, "I/O Intr")
 }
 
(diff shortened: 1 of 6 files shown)


MarkRivers:
This looks very useful, and could be incorporated in R2-6.  Can you update the RELEASE.md and NDPluginStats.html?  Do you have any OPI screens that are updated with the new PVs?

alex-soderqvist:
> This looks very useful, and could be incorporated in R2-6.

That would be very nice.

> Can you update the RELEASE.md and NDPluginStats.html?

Yes.

>  Do you have any OPI screens that are updated with the new PVs?

Not yet, but I will add the PVs to the autoconverted opi and add that.

I also noticed some setter records for the former parameters available, starting here:
https://github.com/areaDetector/ADCore/blob/master/ADApp/Db/NDStats.template#L222
Can I remove these? I guess they are copy paste mistake.

If you have any other aesthetics that need fixing in the stats plugin let me know I might as well fix it while I am at it.

alex-soderqvist:
I have made additions regarding this to release notes, documentation and OPI. I also remove the abundant record in the database template file.
I have not added the new parameters to the timeseries, I guess i should do that as well?

alex-soderqvist:
> I have not added the new parameters to the timeseries, I guess i should do that as well?

Done

MarkRivers:
I am testing this and some things don't quite seem correct.
I am using the simDetector in Peaks mode.  I have a single peak with a 1:2 aspect ratio in X:Y and 2% noise as shown in the following screen shots.
simdetector
peaksetup

The problem is that the Orientation is rapidly flipping between +90 and -90, and the Eccentricity is a huge value as shown in the following screen shot.  @alex-soderqvist is this correct?
stats

MarkRivers:
I have modifed NDStats.adl to have the new records both on the main screen and in the centroid time series plots.  

I have also replaced NDTimeSeriesAll.adl with 2 new files: NDStatsTimeSeriesBasic.adl which contains the plots for the basic statistics, and NDStatsTimeSeriesCentroidAll.adl which contains the plots for the centroid statistics.  I added the word Stats to the file name because these files are specific to the NDPluginStats plugin, while NDTimeSeries.adl is generic, it can be used for any time series data.

alex-soderqvist:
I am working on it, I made a big mistake trusting that imageJ plugin.

alex-soderqvist:
The Eccentricity computation, or elongation which the author of the imageJ plugin also calls it, definitely is wrong. I will replace the calculation with eccentricity as interpreted for an ellipse. 
http://answers.opencv.org/question/59955/program-that-check-for-the-roundness-of-ellipse/

The orientation is in relation to x axis. Since you generate an ellipse in Y direction with noise. I would say that flipping between 90 and -90 is correct. 
If you remove the noise so that image is deterministically generated each time you should see it stable in one extreme.
If you have a centered ellipse spanning from first to third quadrant you would get 45 degrees, and from second to forth quadrant -45 degrees. Also if central moments mu20-mu02 = 0 the orientation is invalid, but in that case you have a circle.
Maybe there is a better representation of this that I am unaware of.

I did double check skew and kurtosis calculations, and they should be correct. However I can't explain the kurtosis being negative. Which would indicate that the distribution is slightly flatter than normal gaussian. 

I will make a new pull request: fixing the eccentricity, documentation and some cosmetics.

alex-soderqvist:
I discussed the orientation problem with a co-worker. Since this is arctangens problem we will never get rid of the oscillations completely for certain scenarios. However if you want one stable value I can add orientation in relation Y axis as well as X axis. This way at least one of them will be stable.

Let me know if you think this is an improvement. In that case I can add for 2.6 release.

MarkRivers:
I think the oscillations are OK.  It probably won't happen in real data, only the simulation has an ellipse so close to horizontal.

MarkRivers:
I have reverted commit 4b9b1e0e2f2c96b7a6f28e29098be18811017b43.  It turns out those ao records were being used for the Reset functionality in the database.  Look at the end of the .template file.

PR #237, opened 2017-01-26
Implemented check for FilePluginClose attribute.

ajgdls:
Added check for FilePluginClose attribute and close file if the value is 1.  This feature is used at DLS (@thomascobb) and we think it is a generally useful feature.


ajgdls:
OK sorry this change obviously needs to be looked at further to ensure it builds on Travis...

ajgdls:
I had missed a later commit when cherry picking.

PR #241, opened 2017-01-27
Build NDPluginSupport.dbd in Makefile

### LINKED ISSUE #239: Should NDPosPlugin.dbd be in NDPluginSupport.dbd?
This breaks IOC builds, unless you've rebuilt ADCore with WITH_XML2=YES.
Other plugins that rely on a build variable are not in NDPluginSupport.dbd.

MarkRivers:
This fixes #239 and simplifies things.  NDPluginSupport.dbd is removed from source and is built in Makefile.  NDPosPlugin.dbd is not put in NDPluginSupport.dbd.


mp49:
This should fix it

Issue #249, opened 2017-02-23
Allow plugins to reprocess previous input array on command 


MarkRivers:
Currently plugins only process when they receive a new NDArray from a driver or another plugin.

It would be nice to be able to force a plugin to process again with the same input NDArray it received the last time it processed.  This would allow changing plugin parameters and seeing the affect without having to run the underlying detector again.  

There are many use cases, for example:

- Changing an ROI definition and looking at the change in statistics
- Repositioning an overlay
- Changing the processing plugin parameters and looking at the result

MarkRivers:
Closed via #250

PR #250, opened 2017-02-23
Process plugin

MarkRivers:
This pull request addresses issue #249 .

The NDPluginDriver base class now stores a pointer to the last NDArray it received.  There is a new ProcessPlugin parameter and record.  When the plugin receives the ProcessPlugin command it executes again with the previous NDArray input.  

All plugin medm screens are changed to be 25 pixels taller to display the new ProcessPlugin PV.

When a plugin reprocesses it does not change the UniqueId of the NDArray.  This resulted in the ImageJ viewer not displaying the image, because it was trigggered by the UniqueID PV in the NDPluginStdArrays plugin.  This was fixed by changing the ImageJ viewer to trigger on the ArrayCounter PV in the NDPluginStdArrays plugin instead of UniqueID.

Removed the parameter counting in NDPluginDriver, which is a nice feature of ADCore R2-6, but one which  has not be used yet.  This required a change to the NDPluginFile PV, which was really misusing the counter from NDPluginDriver.

### DIFF
--- a/ADApp/pluginSrc/NDPluginDriver.cpp
+++ b/ADApp/pluginSrc/NDPluginDriver.cpp
@@ -72,6 +72,10 @@ static const char *driverName="NDPluginDriver";
     if (dimsChanged) {
         doCallbacksInt32Array(this->dimsPrev, ND_ARRAY_MAX_DIMS, NDDimensions, 0);
     }
+    // Save a pointer to the input array for use by ProcessPlugin
+    if (pInputArray_) pInputArray_->release();
+    pInputArray_ = pArray;
+    pArray->reserve();
 }
 
 extern "C" {static void driverCallback(void *drvPvt, asynUser *pasynUser, void *genericPointer)
@@ -357,7 +361,11 @@ asynStatus NDPluginDriver::writeInt32(asynUser *pasynUser, epicsInt32 value)
         this->lock();
     } else if (function == NDPluginDriverQueueSize) {
         newQueueSize_ = value;
-     } else {
+    } else if (function == NDPluginDriverProcessPlugin) {
+        if (pInputArray_) {
+            driverCallback(pasynUserSelf, pInputArray_);
+        }
+    } else {
         /* If this parameter belongs to a base class call its method */
         if (function < FIRST_NDPLUGIN_PARAM) 
             status = asynNDArrayDriver::writeInt32(pasynUser, value);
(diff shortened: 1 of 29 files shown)


ADApp/pluginSrc/NDPluginDriver.cpp
```diff
@@ -72,6 +72,10 @@ static const char *driverName="NDPluginDriver";
     if (dimsChanged) {
         doCallbacksInt32Array(this->dimsPrev, ND_ARRAY_MAX_DIMS, NDDimensions, 0);
     }
+    // Save a pointer to the input array for use by ProcessPlugin
+    if (pInputArray_) pInputArray_->release();
+    pInputArray_ = pArray;
+    pArray->reserve();
```
ulrikpedersen: Should these 2 lines be the other way around, strictly speaking?

ADApp/pluginSrc/NDPluginDriver.cpp
```diff
@@ -357,7 +361,11 @@ asynStatus NDPluginDriver::writeInt32(asynUser *pasynUser, epicsInt32 value)
         this->lock();
     } else if (function == NDPluginDriverQueueSize) {
         newQueueSize_ = value;
-     } else {
+    } else if (function == NDPluginDriverProcessPlugin) {
+        if (pInputArray_) {
+            driverCallback(pasynUserSelf, pInputArray_);
+        }
```
ulrikpedersen: Should we return an error status here if a pInputArray_ is not available instead of silently carrying on?

ADApp/pluginSrc/NDPluginDriver.cpp
```diff
@@ -72,6 +72,10 @@ static const char *driverName="NDPluginDriver";
     if (dimsChanged) {
         doCallbacksInt32Array(this->dimsPrev, ND_ARRAY_MAX_DIMS, NDDimensions, 0);
     }
+    // Save a pointer to the input array for use by ProcessPlugin
```
ulrikpedersen: The plugins now hang on to the last received NDArray - potentially forever. Would it be useful to have feature to force the plugin to drop it? 

It could for example be dropped when the plugin is disabled?

ADApp/pluginSrc/NDPluginDriver.cpp
MarkRivers: I don't think there was a danger the other way, since the pArray has at least one reference count, but I've changed since it looks better.

ADApp/pluginSrc/NDPluginDriver.cpp
MarkRivers: I have now printed an error message and returned an error status.  I improved the error handling in general in the NDPluginDriver::writeInt32() method.

MarkRivers:
I have updated ADCore/RELEASE.md and ADCore/documentation/pluginDoc.html to document this new feature.

ADApp/pluginSrc/NDPluginDriver.cpp
MarkRivers: I changed the code to release the NDArray when the plugin is disabled.

PR #257, opened 2017-03-01
Fixed a leak in NDPosPlugin.

blaz-kranjc:
Fixed the memory leak in the NDPosPlugin. It leaked the filename on every write.

### DIFF
--- a/ADApp/pluginSrc/NDPosPlugin.cpp
+++ b/ADApp/pluginSrc/NDPosPlugin.cpp
@@ -12,6 +12,7 @@
 #include <string>
 #include <list>
 #include <map>
+#include <vector>
 
 #include <epicsTypes.h>
 #include <epicsThread.h>
@@ -296,8 +297,6 @@ asynStatus NDPosPlugin::writeOctet(asynUser *pasynUser, const char *value, size_
   int addr=0;
   int function = pasynUser->reason;
   asynStatus status = asynSuccess;
-  char *fileName = new char[MAX_POS_STRING_LEN];
-  fileName[MAX_POS_STRING_LEN - 1] = '\0';
   const char *functionName = "writeOctet";
 
   status = getAddress(pasynUser, &addr); if (status != asynSuccess) return(status);
@@ -307,15 +306,18 @@ asynStatus NDPosPlugin::writeOctet(asynUser *pasynUser, const char *value, size_
 
   if (function == NDPos_Filename){
     // Read the filename parameter
-    getStringParam(NDPos_Filename, MAX_POS_STRING_LEN-1, fileName);
+    std::vector<char> fileName(MAX_POS_STRING_LEN);
+    getStringParam(NDPos_Filename, MAX_POS_STRING_LEN-1, &fileName[0]);
+    fileName[MAX_POS_STRING_LEN - 1] = '\0';
     // Now validate the XML
     NDPosPluginFileReader fr;
-    if (fr.validateXML(fileName) == asynSuccess){
+    if (fr.validateXML(&fileName[0]) == asynSuccess){
       setIntegerParam(NDPos_FileValid, 1);
     } else {
       setIntegerParam(NDPos_FileValid, 0);
       status = asynError;
     }
+
     // If the status of validation is OK then load the file
     if (status == asynSuccess){
       // Call the loadFile function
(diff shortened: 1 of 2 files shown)


MarkRivers:
Presumably this could have been fixed by just changing

char *fileName = new char[MAX_POS_STRING_LEN];
to
char *fileName[MAX_POS_STRING_LEN];

so it was a stack variable rather than being allocated with new?

ulrikpedersen:
> Presumably this could have been fixed by just changing
>
> char *fileName = new char[MAX_POS_STRING_LEN];
> to
> char *fileName[MAX_POS_STRING_LEN];
>
> so it was a stack variable rather than being allocated with new?

I think so. That would be an even more obvious solution than std::string...

ajgdls:
Hi @ulrikpedersen @MarkRivers  actually I think it should be 
char fileName[MAX_POS_STRING_LEN];
instead of 
char *fileName[MAX_POS_STRING_LEN];

I just cannot believe that I wrote it that way in the first place, using a new for char array?.?.?.

ajgdls:
Oh actually I think I do know, it is because the array is large and I didn't want that on the stack. 1,000,000 elements, because rather than just a filename the parameter can actually take the xml contents or the filename.

blaz-kranjc:
Sorry I missed the new methods with string params. Without them the usage of string is not too practical as according to standard (pre-c++11) the string is not required to be stored contiguously (http://en.cppreference.com/w/cpp/string/basic_string). This means that it's practically impossible to safely use the strings underlying buffer as an input char * (c_str() returns a const char * but it's not portable according to standard to cast the const away). I'll rewrite it to use a string.

Alan is there a better way to name this parameter then? Or at least the variable in the code? It's really unexpected to see something named filename being 1MB in size. I don't know the plugin well enough to name it properly, I was only concerned with the memory leak (with malcom writes we were leaking 5GB in a day).

ajgdls:
Hi yes that name is not very intuitive, there was a requirement creep I'm afraid.  Originally it really was just a file name, but then after some use the extra ability of sending the xml directly through the same parameter was requested.  I'm happy if it is changed to something more descriptive, but be aware that changing the parameter name would be changing the API to the plugin (you could of course leave the record with the same name, but I bet it is called filename or something similar).  As far as I know it is only used by @thomascobb currently for the DLS mapping project.

thomascobb:
A change of the parameter name and record name is quite reasonable, and as long as it happens within the next few weeks is not a problem.

MarkRivers:
I think this commit misses the point of the original use of "new", i.e. to avoid creating a very large memory allocation on the stack.  I think this is still necessary.  areaDetector now supports HDF5 files on vxWorks which creates threads with only about 10 KB of stack space.

The simplest fix would be to simply call "delete fileName" at the end of the original code, both in writeOctet and in loadFile().

ulrikpedersen:
> I think this commit misses the point of the original use of "new", i.e. to avoid creating a very large memory allocation on the stack. I think this is still necessary. areaDetector now supports HDF5 files on vxWorks which creates threads with only about 10 KB of stack space.

This is a rather complex discussion actually. The C++ standard does not specify how the std::string stores the actual character array internally. Sure, the std::string object is created on the stack with a smallish footprint. But internally it may store the actual data on heap or stack - and this is where things get complex. As far as I can tell most modern std::string implementations use a combination of heap/stack depending on the required size. See for example: http://info.prelert.com/blog/cpp-stdstring-implementations 

And in this case we actually now only allocate as much memory as we need - rather than allocating a large chunk of memory that might not be used.

I think we should be OK with this solution even on VxWorks. If it does not then I think we might have a fundamental problem in using STL containers at VxWorks in general... (and we've recently decided to go down that road with asyn)

> The simplest fix would be to simply call "delete fileName" at the end of the original code, both in writeOctet and in loadFile().

That is certainly the smallest change required to fix the memory leak. However, it is not very safe from future changes.

MarkRivers:
I think the STL containers we have used as automatic (stack) variables have been for "small" strings, so they will be OK vxWorks.  Large containers have been on the heap (e.g. class data), which are not a problem even on vxWorks.

blaz-kranjc:
If there is a problem with string allocation the previous code would have the same problem as this one:
Heap allocated char * was provided as an argument to a method taking std::string which results in a new std::string object on the stack (buffer allocated in the same place as now). This version should not introduce any new problems.

ulrikpedersen:
I think this is OK now because using std::strings even with very long strings doesn't allocate large amounts of memory on the stack.

Issue #256, opened 2017-02-26
New plugins to support parallel processing


MarkRivers:
I propose to add 2 new plugins.

* NDPluginFanout (or NDPluginRoundRobin, or NDPluginScatter?)
This plugin will have a different implementation of doCallbacksGenericPointer() from that in the asynNDArrayDriver base class.  Rather than calling all registered callback client for each NDArray, it will only call one client for each NDArray.  On each subsequent callback it cycles through the list of registered callback clients.  This allows the plugin to spread the processing load over multiple plugins.  For example, the plugin could spread the load over 4 NDPluginProcess plugins, so each one only processes every 4'th array.  This will increase by a factor of 4 the number of NDArrays that can be processed in a given amount of time, assuming that spare cores are available.  The initial implementation might be a simple round-robin, but a more sophisticated might queue the next NDArray to the client with the fewest NDArrays in its input queue.

* NDPluginFanin (or NDPluginGather?)
All existing plugins register for callbacks from a single NDArray source (driver or other plugin).  This plugin will be different, it can register for callbacks from up to N different NDArray sources, where N is defined when the plugin is created.  This would allow the outputs from the 4 NDPluginProcess plugins in the above example to be recombined into a single stream that would then be sent to a file plugin.  The initial implementation will probably not attempt to re-order the NDArray output according to the UniqueId, but a more sophisticated implementation could use a buffer to do this.  It would need to handle the case when a particular UniqueId never arrives because it was dropped by a plugin.

ulrikpedersen:
This is an issue which can be divided up in so many different ways... :-)

Also consider a plugin with a generic pool of worker threads (scaled by how many cores are available). Each incoming NDArray is just passed asynchronously to the pool which picks out an idle worker threads and process() called on an instance of the plugin. I think this would require a slightly different plugin API - one which allow multiple "New Style" plugins to exist without overwriting the same asynParameters (i.e. plugins that don't inherit from asyn port driver - but can be called from asyn port drivers).

I find that often plugins can read relevant parameters at the start of the process() call - do the work using a latched copy of the parameters - and then post the results as parameter updates at the very end. Thus the 'real' processing part could be pulled out to an interface which just takes an NDArray const reference and a list or dictionary of key/value attributes and returns a new NDArray and an updated key/value attribute list.

This would require enough interface changes (although it could be done with backwards compatibility) to require a major release like 3-0.

MarkRivers:
[...]

PR #259, opened 2017-03-05
New scatter/gather capability for plugins

MarkRivers:
Added new NDPluginScatter and NDPluginGather to do scatter/gather operations.  These allow multiple plugins to operate in parallel on different NDArrays, increasing the total number of NDArrays that can be processed per second.  This should be useful for example with NDPluginStats or NDPluginProcess where a single plugin instance is not able to keep up with the detector frame rate.  This addresses issue #256.

The NDPluginGather plugin has the option to sort the incoming plugins by UniqueId which can be convenient for output files.

The NDPluginDriver base class was modified slightly.  This change allows the NDPluginScatter plugin to set a flag when it does a callback with an NDArray indicating that if the NDArray cannot be queued that the NDPluginDriver should return a flag indicating this but not treat this as a dropped array in that plugin.  This allows the NDPluginScatter plugin to try other plugins to see if they have room in their queues before it will finally give up.

There is a detailed example of the use of the scatter/gather functionality in the [NDPluginGather documentation](http://cars.uchicago.edu/software/epics/NDPluginGather.html).


MarkRivers:
I had neglected to update the [NDPluginGather documentation](http://cars.uchicago.edu/software/epics/NDPluginGather.html) on our Web site after the latest fixes and additions.  @ulrikpedersen and @brunoseivam if you looked earlier please reload.

brunoseivam:
Hi @MarkRivers, could you please also commit the autoconverted screens for Scatter/Gather to opis?

MarkRivers:
I am not set up to do that autoconversion any more.  Tim used to do it, but it is a lot of work.  I wish we could come up with an automated script for all of them, but the CSS-Boy converter can only be run in the CSS GUI as far as I know.

brunoseivam:
OK, I will try it here on my end

MarkRivers:
I am closing this since it has been superseded by #261.

PR #261, opened 2017-03-09
Thread pool

MarkRivers:
This branch adds the ability to have a plugin run with multiple threads.  The maximum number of threads is defined in the constructor, but the actual number of threads to use can be changed at run time via an EPICS PV.  

There are now 2 message queues.  pToThreadMsgQ_ is used to send NDArrays to the threads, and replaces the previous pMsgQ.  The difference is that we now pass a structure rather than just the NDArray pointer.  The structure contains both a message type and the NDArray pointer.  This allows sending different messages to the threads.  Currently the messages types are for data, and to tell the thread to exit.  The other message queue is used for threads to send messages back to the plugin itself.  This is used to send messages indicating that the thread has started, and that the thread is about to exit.  This message queue replaces the epicsEvent that was previously used to signal thread startup.

When the queue size or the number of threads is changed the following is done.
* deleteCallbackThreads is called and it does the following:
  * Suspends callbacks from the input driver
  * Waits for the threads to finish running all NDArrays that are in the pToThreadMsgQ_.
  * Send the threads a message to exit
  * Waits for the messages that says the threads have exited
  * Deletes the threads and the message queues
* createCallbackThreads is called and it does the following:
  * Creates the message queues.  pToThreadMsgQ_ now may have a different size
  * Creates the threads
  * Calls startCallbackThreads if appropriate
* startCallbackThreads does the following
  *  Calls epicsThread->run() for each thread
  * Waits for the message from each thread that says it has started.

The following shows the simDetector generating over 1900 frames/s, 1024x1024, 8-bit.

simdetector

The following shows the Transform plugin running with 1 thread.  Note that it is processing 376 frames/s and is using 99.9% of a core.
transform_1thread
top_1thread

The following shows the Transform plugin running with 3 threads, by changing the NumThreads record to 3.  Note that it is processing 1150 frames/s and there are now 3 threads each using over 97% of a core.
transform_3thread
top_3thread

The following shows the Transform plugin running with 5 threads, by changing the NumThreads record to 3.  Note that it is processing 1696 frames/s and there are now 5 threads each using about 95% of a core.  With 5 threads the plugin is not dropping any frames, it is able to keep up with the simDetector.
transform_5thread
top_5thread

In order to take advantage of the new multi-threaded capability of NDPluginDriver the derived class must be thread-safe.  NDPluginTransform, NDPluginStdArrays are thread safe.  NDPluginStats is thread-safe for the basic statistics operations, but not yet for more complex statistics like centroid and histogram.  This is planned.  The time-series will also require some thought and work, since multiple threads will produce the time points slightly out of order.  These could be sorted by uniqueId using an std::multiset.

Some plugins can probably never be made thread-safe.  For example NDPluginProcess stores an NDArray that is used for averaging, etc. and is effectively used with read/modify/write, with a reset if the number to average has been reached.  I don't think this lends itself to multithreading.  If a plugin is not thread-safe then it just does not implement the maxThreads parameter in its constructor, and NDPluginDriver will use the default maxThreads value of 1.  Thus the user cannot set the number of threads greater than 1, and it is safe.

A multithreaded plugin has the same issue as NDPluginGather with producing its output NDArrays in the wrong order.  NDPluginGather handles this by optionally using an std::multiset to sort the output NDArrays by uniqueId.  I believe this capability should be moved to the NDPluginDriver base class, because all multi-threaded plugins will want the ability to do this.

This branch is probably not quite ready to merge, but I would appreciate reviews and comments at this point.  Note that this branch is based on the scatter_gather branch which has also not yet been merged into master.


brunoseivam:
This looks interesting. Have you considered using the [thread pool](https://github.com/epics-base/epics-base/tree/3.16/src/libCom/pool) implementation from base? It looks like it was introduced in 3.15.0.2, so I don't know if it is worth it to depend on a newer (>3.14) version of base...

MarkRivers:
I did take a quick look at epicsThreadPool from base, but I am not sure it would gain us much.  As you say, it would require base 3.15 or higher, which could be a problem, for example for Diamond.  

The total code in NDPluginDriver related to threads is 250 lines (createCallbackThreads, deleteCallbackThreads, startCallbackThreads).  Probably 75% of this would be needed even is we used epicsThreadPool because it is related to creating and deleting the epicsMessageQueues and the logic of stopping a thread which is waiting at epicsMessageQueue::receive().

MarkRivers:
The unit tests on this branch are failing, but I am struggling to figure out why.

If I run the CircularBuffTests with no modification I get the following output:
```
corvette:ADCore/ADApp/pluginTests>../../bin/linux-x86_64/plugin-test --run_test=CircularBuffTests
Running 3 test cases...
epicsMutex pthread_mutex_unlock epicsMutexOsdUnlock failed: error Operation not permitted
epicsMutex pthread_mutex_destroy failed: error Device or resource busy
epicsMutex pthread_mutex_unlock epicsMutexOsdUnlock failed: error Operation not permitted
epicsMutex pthread_mutex_destroy failed: error Device or resource busy
epicsMutex pthread_mutex_unlock epicsMutexOsdUnlock failed: error Operation not permitted
epicsMutex pthread_mutex_destroy failed: error Device or resource busy

*** No errors detected
```
Note that the tests appear to succeed but it is getting an error unlocking an epicsMutex.

If I add some printf calls to trace the flow of execution I get:

```
corvette:ADCore/ADApp/pluginTests>../../bin/linux-x86_64/plugin-test --run_test=CircularBuffTests
Running 3 test cases...
Creating dummy_driver
Creating NDPluginCircularBuff
Starting NDPluginCircularBuff
Creating ds
Entering test_BufferWrappingAndStatusMessages
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
calling cbProcess
../test_NDPluginCircularBuff.cpp(134): error in "test_BufferWrappingAndStatusMessages": check storedImages == 10 failed [1 != 10]
../test_NDPluginCircularBuff.cpp(135): error in "test_BufferWrappingAndStatusMessages": check status == "Buffer Wrapping" failed [Buffer filling != Buffer Wrapping]
epicsMutex pthread_mutex_unlock epicsMutexOsdUnlock failed: error Operation not permitted
epicsMutex pthread_mutex_destroy failed: error Device or resource busy
Creating dummy_driver
Creating NDPluginCircularBuff
Starting NDPluginCircularBuff
Creating ds
epicsMutex pthread_mutex_unlock epicsMutexOsdUnlock failed: error Operation not permitted
epicsMutex pthread_mutex_destroy failed: error Device or resource busy
Creating dummy_driver
Creating NDPluginCircularBuff
Starting NDPluginCircularBuff
Creating ds
epicsMutex pthread_mutex_unlock epicsMutexOsdUnlock failed: error Operation not permitted
epicsMutex pthread_mutex_destroy failed: error Device or resource busy

*** 2 failures detected in test suite "NDPlugin Tests"
corvette:ADCore/ADApp/pluginTests>make -sj
corvette:ADCore/ADApp/pluginTests>../../bin/linux-x86_64/plugin-test --run_test=CircularBuffTests
Running 3 test cases...
epicsMutex pthread_mutex_unlock epicsMutexOsdUnlock failed: error Operation not permitted
epicsMutex pthread_mutex_destroy failed: error Device or resource busy
epicsMutex pthread_mutex_unlock epicsMutexOsdUnlock failed: error Operation not permitted
epicsMutex pthread_mutex_destroy failed: error Device or resource busy
epicsMutex pthread_mutex_unlock epicsMutexOsdUnlock failed: error Operation not permitted
epicsMutex pthread_mutex_destroy failed: error Device or resource busy

*** No errors detected
```

Note that with the printf added it now detects errors in the tests.  This suggests a race condition?  It still gives the epicsMutex errors, but these appear to be happening in the cleanup, since they come after the unit test error messages.  

@ulrikpedersen any idea how to track this down?

I ran the unit test in gdb, but it was not helpful, since the epicsMutex errors do not cause the program execution to stop.

MarkRivers:
I figured out the problem with the epicsMutex errors.  The NDPluginDriver destructor calls deleteCallbackThreads().  That function expects to be called with the asynPortDriver lock held.  I can fix the problem by calling "cb->lock()" just before calling "delete cb" in test_CircularBuffTests.  However, a much better fix is to change the deleteCallbackThreads so that it detects if it is called with the lock held and does not call unlock() and lock() if it is not.  That will require a new asynPortDriver::isLocked() method, which seems like a good idea anyway.

MarkRivers:
I have fixed the problem by documenting that the NDPluginDriver destructor expects to be called with the asynPortDriver mutex unlocked.  The destructor now locks before calling deleteCallbackThreads() and unlocks after calling it.  This fixes the problem in a clean way with no changes to asyn.

MarkRivers:
I have pushed a new version of NDPluginStats.  This version is thread-safe for all operations.  

I measured the following increase in frames/s when using the simDetector with 1024x1024 float32 images.  The plugin was configured to do all of the statistics operations (centroid, histogram, etc.).

Threads | frames/s
---------|-----------
1 | 123
2 | 250
3 | 385
4 | 515
5 | 625

The performance is linear in the number of threads.  In this case even 5 threads was not enough to keep up with the simDetector which was generating about 900 frames/s.

I'm quite happy with this new feature.  The user can change the number of threads on the fly to match the requirements of the detector and the experiment.  It's clear that areaDetector plugins can now be  scaled to be able to handle very high data rate detectors.

MarkRivers:
This branch builds fine in ADCore and for detector IOC applications and all of the unit tests.  However, it fails to build ADSimDetector/iocs/simDetectorNoIOC.  I struggled to understand the problem, but have now isolated it to a simple test file which is now built for testing in https://github.com/areaDetector/ADCore/blob/thread_pool/ADApp/pluginSrc/test_forward_reference.cpp. 

Here it is:
```
#include <NDPluginStats.h>

void test_forward_reference()
{
// Creating the object as an automatic variable causes a compilation error on Linux but compiles fine on Windows
  NDPluginStats stats = NDPluginStats("STATS1", 20, 0, "SIM1", 0, 0, 0, 0, 0);

// Creating the object as a heap variable works fine on Linux and on Windows
  NDPluginStats *pStats = new NDPluginStats("STATS1", 20, 0, "SIM1", 0, 0, 0, 0, 0, 0);
}
```

If I compile this file as shown on Windows it builds fine.

If this file is compiled as shown on Linux it produces the errors in the [Travis log file](https://travis-ci.org/areaDetector/ADCore/builds/210754099)

If this file is compiled with the declaration of stats (i.e. the automatic variable) commented out it builds fine on Linux.

I don't understand why there is a compile error on the include file when creating the NDPluginStats plugin as an automatic (stack) variable but not when creating it as a heap variable on Linux, whereas either is fine on Windows?

I have tested on Linux with gcc 4.2.1, 4.4.7, and 4.8.2 and they all fail the same way.

The reason all of the IOCs and all of the unit tests build OK is that they always create plugin objects on the heap with “new”.  Only simDetectorNoIOC creates plugin objects on the stack without using “new”.

The problem arises because NDPluginDriver.h now forward references a sortedListElement class which it uses in a private member data std::multilist.  The class is actually defined in NDPluginDriver.cpp, where I would like to hide it so it is not exposed in the header file.  This works on Visual Studio, and works fine for heap variables, but not for automatic variables on gcc.

I can work around the problem by moving the sortedListElement class definition to NDPluginDriver.h, but I would like to avoid this because it should be possible to hide the definition of this class to the only source code file that needs it.

Can someone explain what is going on?

hinxx:
After quick glance at the log I noticed lines 1482, 1483 that say something of interest. Could that be the root of the problem?

MarkRivers:
Yes, those lines indicate the problem.  My questions are:

- Why does this not produce an error with Visual Studio?
- Why is it an error to instantiate an automatic variable but OK to instantiate a heap variable?

Sinsesa pointed me to [this Stackflow post](http://stackoverflow.com/questions/23119722/cant-allocate-class-with-forward-declared-value-in-stdmap-member-variable).

It says
>Because the C++ standard library containers are not defined to work with incomplete member types. This is by design1 – but it’s arguably an error (and might be changed in future versions of C++). The Boost.Containers library fixes this.

I guess I will need to change the code to not hide the class definition and instead put it in NDPluginDriver.h,

ulrikpedersen:
I notice that in your example code snippet the constructor is called with different number of arguments 9/10. Is this a copy'n paste error or something more significant?

> ```NDPluginStats stats = NDPluginStats("STATS1", 20, 0, "SIM1", 0, 0, 0, 0, 0);```

>```NDPluginStats *pStats = new NDPluginStats("STATS1", 20, 0, "SIM1", 0, 0, 0, 0, 0, 0);```

MarkRivers:
>I notice that in your example code snippet the constructor is called with different number of arguments 9/10. Is this a copy'n paste error or something more significant?

It is a typo in the source code, but does not matter.  The 10'th argument is defined in NDPluginStats.h as
"int maxThreads=1".  The plugin ensures that it is as least 1.

I just pushed a new version of test_forward_reference.cpp which uses 1 for the 10'th argument in both calls, but it fails the same way.

MarkRivers:
In testing plugins running with multiple threads I was getting occasional crashes.  These were due to corruption of the NDArrayPool.  NDArrays were being reserved and released when their reference count was already 0, i.e. the NDArray had already been returned to the pool.

I found the source of the problem.  Plugins and drivers have always been releasing the asynPortDriver lock when calling doCallbacksGenericPointer.  For example, the master branch of NDPluginStats contains the following code:
```
            this->getAttributes(pArrayOut->pAttributeList);
            this->unlock();
            doCallbacksGenericPointer(pArrayOut, NDArrayData, 0);
            this->lock();
```
This code fragment is common to many plugins and drivers.  However, it introduces a time when the lock is not held and thus class variables are not protected.  This worked OK when only 1 thread would  be executing processCallbacks() for a single plugin object, but it can cause corruption when multiple threads are executing processCallbacks() in the sample plugin object.

The origin of unlocking before calling doCallbacksGenericPointer() is due to some deadlock issues I observed many years ago when first developing areaDetector.  I have always wondered why it was necessary, and now I know that it is causing problems.

I have introduced a new base class method NDPluginDriver::doNDArrayCallbacks. This takes care of the above code snippet that was frequently replicated in plugins.  All plugins should call this method at the end of processCallbacks().  It checks to see if ArrayCallbacks is Enabled and acts accordingly.  It does not release the lock when calling doCallbacksGenericPointer.  I have modified a number of the plugins to call this method, and it fixes the NDArrayPool corruption problem and does not appear to introduce any deadlocks.

MarkRivers:
I have made the following changes:
- Removed the numParams argument from NDPluginDriver constructor.  This is no longer needed by asynPortDriver. Changed all plugins to remove this argument when calling NDPluginDriver constructor,
and to remove the LAST_XXX_PARAM and NUM_XXX_PARAMS code from NDPluginXXX.h.
-  Changed maxThreads argument to required, no default of 1.  This improves error checking that plugins are calling the NDPluginDriver constructor correctly.
- Renamed NDPluginDriver::processCallbacks() to NDPluginDriver::beginProcessCallbacks().
   This makes it clear that this method should be called at the start of processCallbacks() in the derived class.
- Changed NDPluginDriver::processCallbacks to a pure virtual method that all derived classes must implement.
- Added NDPluginDriver::endProcessCallbacks.  This method does the NDArray callbacks.  Replaces doNDArrayCallbacks.  Derived classes should call this method, typically at the end of processCallbacks(). For plugins that support multiple threads this method handles the output array sorting.  It also caches the last output array in pArrays[0] so plugins typically don't need to worry about that any more.  It takes two boolean flags:
  - copyArray  This flag must be true if the array being passed is this original array passed to the plugin and hence must be copied before attaching the attributes.   This flag should be false if the array is
already a copy and hence can be modified directly.
  - readAttributes  This flag should be true if getAttributes() needs to be called for the array.  It should be false if the plugin has already called getAttributes for this array.
- Renamed NDPluginDriver::pInputArray_ to pPrevInputArray_ for clarity, and changed from private to protected because derived classes may need to use it.
- Changed all plugins to call NDPluginDriver::beginProcessCallbacks(), and most to call NDPluginDriver::endProcessCallbacks().
- For plugins that still need to directly call doCallbacksGenericPointer() rather than endProcessCallbacks() removed the code that released the asynPortDriver lock during that call.
- The following plugins are now thread-safe and take a final maxThreads argument in their constructor to allow the use of multiple threads:
  - NDPluginColorConvert
  - NDPluginROI
  - NDPluginStats
  - NDPluginStdArrays
  - NDPluginTransform
 - Additional plugins will be made thread-safe as time permits.

MarkRivers:
@ulrikpedersen and @brunoseivam I am ready to have you review this pull request when you have a chance.  The only remaining thing to do is convert more plugins to be thread-safe, but this is not required.  Plugins that are not yet thread-safe simply force MaxThreads to be 1 so the user cannot set NumThreads &gt; 1.

brunoseivam:
Hi Mark,

I didn't review all of it yet, but one thing that I noticed is this:

I was doing some tests with ADSimDetector. When I set a plugin's ArrayCallbacks to 0 (say, for ROI1), I see memory consumption increase steadily, so there's a memory leak somewhere. I couldn't quite find exactly where. My suspect is `endProcessCallbacks` in `NDPluginDriver`, but I will dig deeper.

Interestingly enough, in `NDROI.cpp`'s call to `endProcessCallbacks` I changed the parameter `copyArray` from `false` to `true` and the opposite happens: memory leaks when `ArrayCallbacks` is set to 1.

MarkRivers:
Hi Bruno,
Thanks for finding that issue.  There was a logic error in NDPluginDriver::endProcessCallbacks when NDArrayCallbacks was false.  It was reserving the NDArray when copyArray is false, but it should be reserving it when copyArray is true.

> Interestingly enough, in NDROI.cpp's call to endProcessCallbacks I changed the parameter copyArray from false to true and the opposite happens: memory leaks when ArrayCallbacks is set to 1.

That is expected.  If you copyArray to true then it assumes the NDArray you are passing to NDPluginDriver::endProcessCallbacks is the original NDArray passed to the plugin.  That NDArray will be released in NDPluginDriver::processTask.  If the plugin has created a new NDArray array (which NDPluginROI does) then that NDArray has been reserved, but will never be released.

MarkRivers:
I have updated NDPluginOverlay so it is now also thread-safe.

MarkRivers:
I've changed all the plugins that will benefit from multiple threads to be thread-safe.  The list of all plugins with comments about the support of multiple threads is in [RELEASE.md](https://github.com/areaDetector/ADCore/blob/thread_pool/RELEASE.md) on the thread_pool branch.  The following support multiple threads:

- NDPluginColorConvert
- NDPluginFFT
- NDPluginOverlay
- NDPluginROI
- NDPluginROIStat
- NDPluginStats
- NDPluginStdArrays
- NDPluginTransform

I will be on vacation this week.  I am hoping @ulrikpedersen, @brunoseivam and others can review and test this pull request this week so I can merge it when I return on April 10.

ulrikpedersen:
> I will be on vacation this week. I am hoping @ulrikpedersen, @brunoseivam and others can review and test this pull request this week so I can merge it when I return on April 10.

Thanks Mark. I appreciate you setting out a schedule - I wasn't sure whether you had completed development on this rather large and complex feature. 

I will take a look to review it this week (along with @GDYendell ). FYI: I'm away on vacation next week.

ulrikpedersen:
Having reviewed with the help of @GDYendell we have a few comments:

Build
------

Builds on our Linux system (merged onto our local master branch) and runs. We have not yet tested on Windows, nor have we exercised it to its full extent. So far it looks like it works.

API changes
-------------

The Plugin API have changed with changes to processing callbacks and locking mechanism and NDPluginDriver constructor arguments. In my opinion the change of a public API (if not 100% backwards compatible) warrant a major version increment. See: http://semver.org/

I propose this changeset should be released as ADCore 3.0.

User Interface PVs
--------------------

There are a set of new PVs for each plugin to control the multi-threaded functionality. We've noticed that the NumThreads PV accepts numbers higher than the configured maximum without returning an asynError. Fortunately the NumThreads_RBV reflects the real value but it would be appropriate to return an asynError when trying to configure an invalid number of threads. Catching issues at configure time is better than unexpected behavior at runtime...

Conclusion
------------

Thanks @MarkRivers  - this is a really useful addition and will enable us to make much better use of servers with many cores/cpus. It also improves the API, in particular with respect to locking/unlocking in plugins. 

With only a brief testing session it appears to work - and the history of this PR shows several issues have already been identified and fixed. I think this feature is ready to merge into master but please consider the above comments before making a release.

FYI; I'm away for the next week but don't let me hold you up! :+1:

MarkRivers:
@ulrikpedersen and @GDYendell thanks for the comments.  I have fixed so that setting numThreads to an invalid number returns asynErrror in NDPluginDriver::writeInt32().  I have changed the next release to R3-0 rather than R2-7.

PR #262, opened 2017-03-22
Add NTNDArray IJ plugin viewer

argonnexraydetector:
This is the new IJ plugin for epics 4. It seems to work OK for viewing NTNDArrays. Colormode seems undefined in the NTNDArray, so it only works w/ black and white images.

Tim


MarkRivers:
ADCore/Viewers/ImageJ/EPICS_areaDetector contains the .jar files for jca and caj.  This has 2 advantages:
- It saves the user the trouble of finding and downloading these files.
- It guarantees that the versions of caj and jca are consistent with the current version of the plugin.

I suggest that the same be done for the required EPICS V4 jar files.  In fact there are no jar files in the EPICS V4 distribution, so it is not even possible for a user to simply go to a Web site and download them.  How should this be handled?  Should areaDetector include a script to build the required jar files, or should this be the responsibility of the V4 package?

argonnexraydetector:
I will fix up the code for tabs etc. 
There are jars distributed w/ EPICS v4 java. They are in epics-core dir. I had put all the jars in my IJ plugins dir, but forgot about them on pull req. 

Also, I will look at colorMode again. I did not find any colorMode field before, but perhaps I missed it.

T

MarkRivers:
>Also, I will look at colorMode again. I did not find any colorMode field before, but perhaps I missed it.

ColorMode is not a field, it is an attribute.  This is how you retrieve it using pvget:
```
corvette:~/scratch>pvget -r 'field(attribute)' 13SIM1:Pva1:Image
13SIM1:Pva1:Image
structure
    epics:nt/NTAttribute:1.0[] attribute
        epics:nt/NTAttribute:1.0
            string name ColorMode
            any value
                int  0
            string[] tags []
            string descriptor Color mode
            int sourceType 0
            string source Driver
        epics:nt/NTAttribute:1.0
            string name AcquireTime
            any value
                (none)
            string[] tags []
            string descriptor Camera acquire time
            int sourceType 2
            string source 13SIM1:cam1:AcquireTime
        epics:nt/NTAttribute:1.0
            string name RingCurrent
            any value
                double  93.5545
            string[] tags []
            string descriptor Storage ring current
            int sourceType 2
            string source S:SRcurrentAI
...
```
The ColorMode enums are defined in NDArray.h as follows:
```
/** Enumeration of color modes for NDArray attribute "colorMode" */
typedef enum
{
    NDColorModeMono,    /**< Monochromatic image */
    NDColorModeBayer,   /**< Bayer pattern image, 1 value per pixel but with color filter on detector */
    NDColorModeRGB1,    /**< RGB image with pixel color interleave, data array is [3, NX, NY] */
    NDColorModeRGB2,    /**< RGB image with row color interleave, data array is [NX, 3, NY]  */
    NDColorModeRGB3,    /**< RGB image with plane color interleave, data array is [NX, NY, 3]  */
    NDColorModeYUV444,  /**< YUV image, 3 bytes encodes 1 RGB pixel */
    NDColorModeYUV422,  /**< YUV image, 4 bytes encodes 2 RGB pixel */
    NDColorModeYUV411   /**< YUV image, 6 bytes encodes 4 RGB pixels */
} NDColorMode_t;
```
In the above case ColorMode is 0 which is Mono.

```
corvette:ADCore/ADApp/ADSrc>pvget -r 'field(attribute)' 13SIM1:Pva1:Image
13SIM1:Pva1:Image
structure
    epics:nt/NTAttribute:1.0[] attribute
        epics:nt/NTAttribute:1.0
            string name ColorMode
            any value
                int  2
            string[] tags []
            string descriptor Color mode
            int sourceType 0
            string source Driver
        epics:nt/NTAttribute:1.0
            string name AcquireTime
            any value
                (none)
            string[] tags []
            string descriptor Camera acquire time
            int sourceType 2
            string source 13SIM1:cam1:AcquireTime
```
  
In this case ColorMode is 2, i.e. RGB1.

NDArrays and NTNDArrays should always contain the ColorMode attribute, but they might not so you should assume monochrome if that attribute is not present.

MarkRivers:
I copied the EPICS V4 .jar files and your plugin .java code to my ImageJ/plugins directory and it is working fine!  Thanks for doing this.  

Note that ImageJ/Plugins/Compile&Run now gives an error on Windows if you are running Java 1.6, it requires Java 1.8.  This is true even on plugins that used to build fine with Java 1.6.  No big deal, just download the latest 1.8 version from the ImageJ Web site.

argonnexraydetector:
I edited the plugin to deal with ColorMode. Also removed \t characters. 
I added epics v4 jars to the repo as well. Give a try.

MarkRivers:
I just tested these changes and color mode seems to work fine.  

The performance I measured the following performance increase in displaying images in ImageJ.

|       |  NDPluginStdArrays, EPICS_ADViewer | NDPluginPva, EPICS_NTNDA_Viewer |
| --- | ---------------------------------------- | ---------------------------------------- |
|1024x1024 8-bit mono   |    63 frames/s, NDPluginStdArrays=33% CPU in IOC  |   86 frames/s, NDPluginPva=9% CPU in IOC |
|1024x1024 RGB1            |     24 frames/s, NDPluginStdArrays=43% CPU in IOC | 29 frame/s, NDPluginPva=5.5% CPU in IOC |

So using pvAccess rather than Channel Access increases number of frames/s displayed in ImageJ by 20%-36%, and reduces the CPU load on the IOC by a factor of 4-8.

MarkRivers:
I made some minor changes to the EPICS_NTNDA_Viewer.java:

Tim started from the version of EPICS_AD_Viewer.java in ADCore R2-5.  

In R2-6 I added the following enhancements:

- Automatically set the contrast when a new window is created. This eliminates the need to manually set the contrast when changing image size, data type, and color mode in many cases.
- When the image window is automatically closed and reopened because the size, data type, or color mode changes the new window is now positioned in the same location as the window that was closed.

I have applied these enhancements to EPICS_NTNDA_Viewer.java and tested them.

I reformatted the source code to have uniform indentation to make it easier to read.  If you look at the differences with 
```
git diff -w -b 8045557a678432e2adf060098e453800a7369c69 c790f152e6f20b785f24c59d0f0bf5db1fc2e8d4 EPICS_NTNDA_Viewer.java
```
you will see that the changes are just removing extra blank lines plus the changes mentioned above.

argonnexraydetector:
Good features.
Tim

PR #263, opened 2017-03-24
Unittest fixes

MarkRivers:
Fixed some problems with the unit tests.

- The tests were producing hundreds of warnings like this
```
NDArray::release: WARNING, no owner
NDArray::reserve: WARNING, no owner
```
These were happening because the NDArrays were being allocated in fillNDArrays() by directly calling "new NDArray", rather than calling NDArrayPool->alloc().  I created a new function fillNDArraysFromPool and changed all unit tests to use this new function, eliminating all of the warning messages.

- test_NDPosPlugin was failing because it was not setting NDArrayCallbacks to 1.  The NDPosPlugin had a bug and was doing NDArrayCallbacks even when NDArrayCallbacks was 0.  Fixing the plugin exposed the error in the unit test.

- test_NDPluginROI was failing because it was always comparing downstream_plugin->arrays[0], when it should have been comparing downstream_plugin->arrays.back(), i.e. the most recent NDArray from the callback.  I don't see how it ever worked, but now it is working fine.


ADApp/pluginTests/test_NDFileHDF5.cpp
```diff
@@ -27,9 +27,10 @@ using namespace std;
 #include "HDF5PluginWrapper.h"
 #include "HDF5FileReader.h"
 
+static  NDArrayPool *arrayPool;
```
ulrikpedersen: Out of curiosity: why make this a static?

ADApp/pluginTests/testingutilities.h
```diff
@@ -16,6 +16,7 @@
 #include <asynPortClient.h>
 
 void fillNDArrays(const std::vector<size_t>& dimensions, NDDataType_t dataType, std::vector<NDArray*>& arrays);
```
ulrikpedersen: Can we delete fillNDArrays() if all of the tests now use fillNDArraysFromPool?

ajgdls:
Hi I had a look and the test updates for the test_NDPosPlugin file look fine.  However, I couldn't see that the bug in the actual PosPlugin driver had been fixed (calling the callbacks without checking for NDArrayCallbacks)?

ulrikpedersen:
> However, I couldn't see that the bug in the actual PosPlugin driver had been fixed (calling the callbacks without checking for NDArrayCallbacks)?

Ah yes, it looks like that bug is still here on master: https://github.com/areaDetector/ADCore/blob/master/ADApp/pluginSrc/NDPosPlugin.cpp#L213

> The NDPosPlugin had a bug and was doing NDArrayCallbacks even when NDArrayCallbacks was 0. Fixing the plugin exposed the error in the unit test.

@MarkRivers: did you fix the plugin on another branch?

MarkRivers:
Yes I fixed the bug on the thread_pool branch. It is fixed in a way that cannot be done on master.

ADApp/pluginTests/test_NDFileHDF5.cpp
MarkRivers: Are you suggesting to make it global so we don't need to pass the arrayPool argument to fillNDArraysFromPool?

ADApp/pluginTests/testingutilities.h
MarkRivers: Yes, it could be deleted if everyone is happy with requiring the use of NDArrayPool to create new NDArrays.

PR #268, opened 2017-04-20
Attribute improvements

MarkRivers:


MarkRivers:
This pull request does the following:
* asynNDArrayDriver, NDFileNexus
  * Changed XML file parsing from using TinyXml to using libxml2. TinyXml was originally used because libxml2 was not available for vxWorks and Windows. libxml2 was already used for NDFileHDF5 and NDPosPlugin, originally using pre-built libraries for Windows in ADBinaires. ADSupport now provides libxml2, so it is available for all platforms, and there is no need to continue building and using TinyXml. This change means that libxml2 is now required, and so the build option WITH_XML2 is no longer used. XML2_EXTERNAL is still used, depending on whether the version in ADSupport or an external version of the library should be used.
  * Added support for macro substitution in the XML files used to define NDAttributes. There is a new NDAttributesMacros waveform record that contains the macro substitution strings, for example "CAMERA=13SIM1:cam1:,ID=ID34:".
  * Added a new NDAttributesStatus mbbi record that contains the status of reading the attributes XML file. It is used to indicate whether the file cannot be found, if there is an XML syntax error, or if there is a macro substitutions error.

MarkRivers:
I think this is ready to merge.  Can @ulrikpedersen and @brunoseivam review when you get a chance?

ADApp/ADSrc/asynNDArrayDriver.cpp
```diff
@@ -430,8 +510,9 @@ asynStatus asynNDArrayDriver::writeOctet(asynUser *pasynUser, const char *value,
     /* Set the parameter in the parameter library. */
     status = (asynStatus)setStringParam(addr, function, (char *)value);
 
-    if (function == NDAttributesFile) {
-        this->readNDAttributesFile(value);
+    if ((function == NDAttributesFile) ||
+        (function == NDAttributesMacros)) {
+        this->readNDAttributesFile();
```
brunoseivam: Minor observation: on IOC loading, autosave will write sequentially to `NDAttributesFile` and then to `NDAttributesMacros`, causing the file to be loaded and parsed twice (which can take a few hundred of ms each time, from what I observed). A minor workaround would be to reverse the order of the records in `NDArrayBase.template` so that `NDAttributesMacros` is restored first but won't trigger a file reading/parsing since `NDAttributesFile` will be empty at that time. 

Ideally there would be some mechanism to set both parameters at the same time, but having something like `NDAttributesFileProcess` would be a bit cumbersome and break previous expectations.

MarkRivers:
> Minor observation: on IOC loading, autosave will write sequentially to NDAttributesFile and then to NDAttributesMacros, causing the file to be loaded and parsed twice (which can take a few hundred of ms each time, from what I observed). A minor workaround would be to reverse the order of the records in NDArrayBase.template so that NDAttributesMacros is restored first but won't trigger a file reading/parsing since NDAttributesFile will be empty at that time.

I have changed NDArrayBase.template as Bruno suggested.  I verified that previously it read the attributes file twice at iocInit, and now it only reads it once.  Note that it takes 500 ms each time it is loaded because there is an epicsThreadSleep(0.5) after the file is read to allow PVAttributes time to connect to the EPICS PVs.  This time could be reduced or even eliminated if we are willing to accept that the first few arrays may be collected with undefined attributes if acquisition is in progress when the attributes file is read.

PR #275, opened 2017-05-15
Add NDPva vs NDStdArrays performance comparison to documentation

brunoseivam:
I ran some performance tests here and I thought it would be nice to have them in `NDPva`'s documentation.


ulrikpedersen:
Good to see CPU performance :+1:

Even more to see network bandwidth utilization when pushing the system on a 10gbps NIC - and even **more** interesting would be to see BW performance when streaming from Windows to Linux :wink:

MarkRivers:
Was that the CPU performance on the client or server computer?  Presumably the client since running CSS made a big difference?  The CPU % is for a single core, i.e. 65% of 1 core, not 65% of 12 cores, right?  It would also be interesting to see the CPU load on the server when running all 8 cameras simultaneously and the cameras were all connected to CSS-Studio clients, first using NDPluginPva and then using NDPluginStdArrays.

MarkRivers:
New documentation is now on the Web:
http://cars.uchicago.edu/software/epics/NDPluginPva.html

ulrikpedersen:
> Was that the CPU performance on the client or server computer? Presumably the client since running CSS made a big difference? The CPU % is for a single core, i.e. 65% of 1 core, not 65% of 12 cores, right? It would also be interesting to see the CPU load on the server when running all 8 cameras simultaneously and the cameras were all connected to CSS-Studio clients, first using NDPluginPva and then using NDPluginStdArrays.

I think the analysis that you can do on the measurements presented in the graph there is just a comparison of CA (stdArr) vs PVA (NDPva) CPU performance - and the effect of adding a client. I'm assuming the percentages are for a single CPU (normally CPU-load across multiple CPUs are illustrated in >100% numbers).

I agree that there are a lot of other measurements that would be interesting to compare (and benchmark). I'm mostly interested in utilization of network bandwidth these days :smiley:

brunoseivam:
I think I could clarify it in the docs. These were CPU usage percentages as reported by `top` on the server while all cameras were simultaneously running (so single core usage). As Ulrik said I tried to see the effect of having/not having clients listening to the PVs. For the measures with CS-Studio, all 8 cameras were opened for display.

PR #276, opened 2017-05-18
Built and tested the ADURL module from GraphicsMagick sources.

pheest:
Hi Mark,

As we discussed.

### DIFF
--- a/ADApp/ADSrc/Makefile
+++ b/ADApp/ADSrc/Makefile
@@ -52,18 +52,18 @@ ifeq ($(EPICS_LIBCOM_ONLY),YES)
 else
   LIB_LIBS += $(EPICS_BASE_IOC_LIBS)
 endif
-ifeq ($(XML2_EXTERNAL),NO)
+ifeq ($(XML2_EXTERNAL), NO)
   LIB_LIBS += xml2
 else
+  ifdef XML2_INCLUDE
+    USR_INCLUDES += -I$(XML2_INCLUDE)
+  endif
   ifdef XML2_LIB
     xml2_DIR     = $(XML2_LIB)
     LIB_LIBS     += xml2
   else
     LIB_SYS_LIBS += xml2
   endif
-endif
-ifdef XML2_INCLUDE
-  USR_INCLUDES += $(XML2_INCLUDE)
 endif
 
 # Install commonDriverMakefile and commonLibraryMakefile in top-level cfg directory.
--- a/ADApp/commonDriverMakefile
+++ b/ADApp/commonDriverMakefile
@@ -64,14 +64,16 @@ ifeq ($(WITH_JPEG),YES)
   endif
 endif
 
-ifeq ($(XML2_EXTERNAL),NO)
-  PROD_LIBS += xml2
-else
-  ifdef XML2_LIB
-    xml2_DIR     = $(XML2_LIB)
-    PROD_LIBS     += xml2
+ifeq ($(WITH_XML2),YES)
+  ifeq ($(XML2_EXTERNAL),NO)
+    PROD_LIBS += xml2
   else
-    PROD_SYS_LIBS += xml2
+    ifdef XML2_LIB
+      xml2_DIR     = $(XML2_LIB)
+      PROD_LIBS     += xml2
+    else
+      PROD_SYS_LIBS += xml2
+    endif
   endif
 endif
 
@@ -138,18 +140,15 @@ endif
 
 # The following libraries are only needed for GraphicsMagick
 ifeq ($(WITH_GRAPHICSMAGICK), YES)
+  ifeq ($(GRAPHICSMAGICK_EXTERNAL),NO)
   $(PROD_NAME)_DBD += NDFileMagick.dbd
-  ifdef GRAPHICS_MAGICK_LIB
-    GraphicsMagick_DIR     = $(GRAPHICS_MAGICK_LIB)
-    GraphicsMagick++_DIR   = $(GRAPHICS_MAGICK_LIB)
-    GraphicsMagickWand_DIR = $(GRAPHICS_MAGICK_LIB)
-    PROD_LIBS               += GraphicsMagick++ GraphicsMagickWand GraphicsMagick
+    PROD_LIBS               += Magick++ Magick coders jbig jp2 bzlib png webp lcms ttf wmf filters
   else
-    PROD_SYS_LIBS           += GraphicsMagick++ GraphicsMagickWand GraphicsMagick
+    PROD_SYS_LIBS           += Magick++ Magick coders jbig jp2 bzlib png webp lcms ttf wmf filters
+    PROD_SYS_LIBS_Linux    += gomp X11 png12 bz2 Xext freetype 
+    PROD_SYS_LIBS_Darwin   += png bz2
   endif
   # The following system libraries must be installed to use GraphicsMagick
-  PROD_SYS_LIBS_Linux    += gomp X11 png12 bz2 Xext freetype 
-  PROD_SYS_LIBS_Darwin   += png bz2
 endif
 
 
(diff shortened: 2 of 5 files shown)


MarkRivers:
Peter, the Travis CI is also failing with your pull request.  It is not obvious to me why this is, but can you please see if you can figure it out?

pheest:
>> As of ADCore R3-0 WITH_XML2 is no longer supported
The version I forked and merged with - just a couple of days ago (and well after you announced the release of 3.0) - still had this usage in it.

But I the version I downloaded just a moment ago, does not.
I don't understand this. Has something gone wrong with the technolgy?
Does this relate to your comment about the CI?

If WITH_XML2=NO is no longer an option, then it should of course be removed.

>> I don't understand why you removed the lines with GRAPHICS_MAGICK_LIB
I'm not sure I understand how and where these variables are used.
I don't see any usage of them in any of the code that I have here.
Please could you explain?

>> I don't understand why you needed to add H5_BUILT_AS_DYNAMIC_LIB to pluginSrc/Makefile
I have just re-verified that, without this change, 4 of the build configurations (those with either debug or 64-bit enabled), fail. I will send you the build logs.

MarkRivers:
@pheest can you please update your pull request to remove WITH_XML2?  I would like to test this.

pheest:
Hi Mark,

Probably not till next week, as I’m a bit busy just now.
(NB, Monday is a public holiday in the UK, and Tuesday the CCFE site is closed also.)

You could make this change yourself, surely?
I was assuming that you had.

Cheers,

Peter.

MarkRivers:
I can't modify the pull request myself, only you can do that.  I can merge the pull request and then revert the files you changed to again remove the WITH_XML2 stuff.  It makes the Git commit history ugly, but that is how I will proceed.

MarkRivers:
I said:
> I don't understand why you removed the lines with GRAPHICS_MAGICK_LIB

@pheest replied:
> I'm not sure I understand how and where these variables are used.
I don't see any usage of them in any of the code that I have here.
Please could you explain?

The R3-0 version of commonDriverMakefile and commonLibraryMakefile contained these lines:
```
   GraphicsMagick_DIR     = $(GRAPHICS_MAGICK_LIB)
   GraphicsMagick++_DIR   = $(GRAPHICS_MAGICK_LIB)
   GraphicsMagickWand_DIR = $(GRAPHICS_MAGICK_LIB)
```
The definitions of GraphicsMagick_DIR, etc. will be used by the EPICS build system rules.  If your link command contains the library GraphicsMagick and if GraphicsMagick_DIR is defined then make will look in GraphicsMagick_DIR for that library.  This is required in order to use GraphicsMagick libraries if they are not located either in ADSupport or in a standard system library location.

I am restoring these lines to the Makefiles.

MarkRivers:
>Peter, the Travis CI is also failing with your pull request. It is not obvious to me why this is, but can you please see if you can figure it out?

@pheest I figured out why the Travis build was failing and fixed it.  There were 2 errors in the R3-0 pluginTests/Makefile
- The lines to add XXX_INCLUDE to USR_INCLUDES were missing the -I in front of them.  You fixed this.
- There was no conditional to only add XXX_INCLUDE to USR_INCLUDES if XXX_INCLUDE was defined.  That led to compiler commands that had "-I -I -I" in them, which then confused the compiler which interpreted the next token as a directory.  I fixed this by adding the conditional, and Travis now succeeds.

MarkRivers:
I removed XML2_EXTERNAL test in adding XML2_INCLUDE to USR_INLCUDES, no other libraries do this.  If the user defines XML2_EXTERNAL=NO then they should not define XML2_INCLUDE.

I removed -DH5_BUILT_AS_DYNAMIC_LIB from pluginSrc/Makefile.  This should not be needed since this is defined in commonLibraryMakefile.

I have successfully built all of areaDetector (ADSupport, ADCore, most detectors) with these changes.  I am building as follows:

- Everything built with WITH_XXX_EXTERNAL=NO so it uses ADSupport for everything.
- WITH_GRAPHICS_MAGICK=NO so I am just testing that previous builds work OK.
- Base 3.15.5.  
- Linux is Centos 7, gcc 4.8.5.
- master branch of ADCore and ADSupport as of the time of this message.
- Windows 7 with both Visual Studio 2010 (Professional) and Visual Studio 2015 (Community Edition).

I built for these architectures:

linux-x86
linux-x86_64
win32-x86   (Visual Studio 2010, 32-bit, dynamic)
win32-x86-static (Visual Studio 2010, 32-bit, static)
win32-x86-static-vs2015 (Visual Studio 2015, 32-bit, static)
win32-x86-vs2015 (Visual Studio 2015, 32-bit, dynamic)
windows-x64 (Visual Studio 2010, 64-bit, dynamic)
windows-x64-static (Visual Studio 2010, 64-bit, static)
windows-x64-static-vs2015 (Visual Studio 2015, 64-bit, static)
windows-x64-vs2015 (Visual Studio 2015, 64-bit, dynamic)

The following still need to be done.
- vxWorks build is failing in ADSupport/supportApp/xml2Src/nanoftp.c.  Hopefully I just need to work on the os/vxWorks/config.h
- Test WITH_GRAPHICS_MAGICK=YES and GRAPHICS_MAGICK_EXTERNAL=NO.

pheest:
It would be an option to close the pull request and then re-create it with accumulated corrections.
In order to preserve a cleaner history of the repo.
I'm not able to follow all of your comments.

pheest:
Are these lines not associated with use header and LIB files that are stored in ADBinaries?
I had assumed this usage would no longer be a supported option.

There are 3 conceptually feasible approaches:
a. Header and LIB files at some arbitrary path (but typically ADBinaries).
b. Header and LIB files present as installed-product (system) folders.
c. Header and source files present in build.

It seems to me that for all three approaches to be available isn’t desirable.

Please correct me if my understanding of this is wrong.

Cheers,

Peter.

pheest:
DH5_BUILT_AS_DYNAMIC_LIB is not required in pluginSrc/Makefile, but it would be very helpful to improve the logic condition in commonLibraryMakefile so that a manually-added definition of HDF5_STATIC_BUILD isn’t required.
Making the condition:
  ifeq ($(SHARED_LIBRARIES), YES)
works perfectly.

ulrikpedersen:
> DH5_BUILT_AS_DYNAMIC_LIB is not required in pluginSrc/Makefile, but it would be very helpful to improve the logic condition in commonLibraryMakefile so that a manually-added definition of HDF5_STATIC_BUILD isn’t required.
Making the condition:
  ifeq ($(SHARED_LIBRARIES), YES)
works perfectly.

We can't do that close coupling of the EPICS and HDF5 builds as we sometimes need a combination of shared/static build.

MarkRivers:
@pheest wrote:
>Are these lines not associated with use header and LIB files that are stored in ADBinaries?
I had assumed this usage would no longer be a supported option.
There are 3 conceptually feasible approaches:
a. Header and LIB files at some arbitrary path (but typically ADBinaries).
b. Header and LIB files present as installed-product (system) folders.
c. Header and source files present in build.
It seems to me that for all three approaches to be available isn’t desirable.
Please correct me if my understanding of this is wrong.

No, those lines had nothing to do with ADBinaries.  In fact because ADBinaries was organized as an EPICS support directory, and was included in ADCore/configure/RELEASE,  _DIR and _INCLUDE were not needed for ADBinaries, it was automatically searched by the EPICS build system.

The cases we need to support are:
- Header and LIB files at some arbitrary path.  This is intended for the case when the the system version of the library is the wrong version, or the library cannot be installed as a system library for some reason.  For example the EPICS developer wants to install and use a new version of HDF5 without affecting all other users by replacing the system version.  The installation can be done using the normal autoconfigure mechanism, but not in a system directory.  This is the only case that requires defining XXX_LIB and XXX_INCLUDE.
- Header and LIB files present as installed-product (system) folders.
- Header and source files present in build, e.g. ADSupport.

PR #278, opened 2017-06-07
Fix NDCircularBuff EDM screens for PV name changes

GDYendell:
* POST_COUNT -> PostCount
    * PRE_COUNT -> PreCount


GDYendell:
An update for the EDM screens for NDPluginCircularBuffer for the change to some PV names in commit d0443a9.

PR #282, opened 2017-07-03
add LZF and bitshuffle+lz4 compression

xiaoqiangwang:
Both these two filters claim fast compression. This could be useful for fast detectors.

A preliminary setup with simDetector creating 1024x1024 int32 image at 100 Hz (400MB/s), and HDF5 plugin writing to a ramdisk. With bitshuffle+lz4, the writing speed is as fast as no compression.

### DIFF
--- a/ADApp/Db/NDFileHDF5.template
+++ b/ADApp/Db/NDFileHDF5.template
@@ -477,6 +477,10 @@ record(mbbo, "$(P)$(R)Compression")
     field(TWVL, "2")
     field(THST, "zlib")
     field(THVL, "3")
+    field(FRST, "lzf")
+    field(FRVL, "4")
+    field(FVST, "bshuf")
+    field(FVVL, "5")
     info(autosaveFields, "VAL")
 }
 
@@ -494,6 +498,10 @@ record(mbbi, "$(P)$(R)Compression_RBV")
     field(TWVL, "2")
     field(THST, "zlib")
     field(THVL, "3")
+    field(FRST, "lzf")
+    field(FRVL, "4")
+    field(FVST, "bshuf")
+    field(FVVL, "5")
 }
 
 record(longout, "$(P)$(R)NumDataBits")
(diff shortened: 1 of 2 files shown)


ulrikpedersen:
Thanks for the contribution - I think that could be a useful addition but there's a bit more work to be done before its ready for users...

So this is using the HDF5 dynamically loaded filter feature, presumably? We would need some instructions for users on how to configure their environment to use these.

Does this work on Windows?

> With bitshuffle+lz4

The way you have added these two new filter options you can't have both in combination. I'm guessing you are referring to using bitshuffle with its internal use of lz4 then (like the Eiger does)?

> the writing speed is as fast as no compression.

The simDetector generate compression-friendly data. Imaging cameras with CCD/CMOS sensors does not. In my experience this might work well for things like diffraction but not so well for tomography. Basically the milage may vary (greatly) depending on experiment and detector.

xiaoqiangwang:
My quick test is to patch ADSupport to include source codes of lz4 and bitshuffle (as copied from https://github.com/kiyo-masui/bitshuffle), instead of using dynamic library.

I agree that compression speed/ratio is dataset dependant. None is universal. Another option I was looking is [blosc](https://github.com/Blosc/c-blosc) which brings multiple compression algorithms under one scheme.

ulrikpedersen:
> My quick test is to patch ADSupport to include source codes of lz4 and bitshuffle (as copied from https://github.com/kiyo-masui/bitshuffle), instead of using dynamic library

That would be a good starting point, combined with instructions in docs.

> I agree that compression speed/ratio is dataset dependant. None is universal. Another option I was looking is blosc which brings multiple compression algorithms under one scheme.

This is a more interesting filter as it allow running compression and decompression in multiple threads. It comes with LZ4 and I think they have recently added a copy of bitshuffle. This would be my preferred option.

xiaoqiangwang:
Ok. I take back this PR and test the blosc filter.

PR #283, opened 2017-07-07
add support of blosc filter

xiaoqiangwang:
This PR when combined with https://github.com/areaDetector/ADSupport/pull/14, adds support of blosc filter.


MarkRivers:
I have made the following changes:
- Modified commonDriverMakefile and commonLibraryMakefile to link with the blosc library if WITH_BLOSC=YES
- Tweaked the layout of NDFileHDF5.adl
- Change NDFileHDF5.template so that all output records have PINI=YES.  This is the normal setting, so the value in the database or autosave is actually sent to the driver.  Removed PINI=NO from all input records since this is the default, and it makes the file more readable to remove them.

PR #286, opened 2017-08-08
Removed sseq

stuwilkins:
This file sseq_settings.req appears to be no longer in the repo. Not sure if it should be removed from the request file or if it needs to be back in. 

This PR stoped the errors on the IOC shell.


stuwilkins:
Sorry, I see that this is actually in the CALC module and would be enabled by the lines
```
# Optional: load sseq record for acquisition sequence
#dbLoadRecords("$(CALC)/calcApp/Db/sseqRecord.db", "P=$(PREFIX), S=AcquireSequence")
#set_requestfile_path("$(CALC)/calcApp/Db")
```

If this is not set then the requestfile path is not set.... Sorry, I will close

PR #300, opened 2017-11-22
Changed Capture mode. 

MarkRivers:
Previously Capture mode created NumCapture NDArrays using the "new" operator.

As NDArrays arrived it copied the data and attributes into these arrays.
This had several problems:
- The NDArrays were not allocated from the NDArrayPool, so memory limits were not enforced.
- Because they were not allocated from the NDArrayPool if they were passed to other functions
  or plugins there would be problems with attempts to called NDArray::reserve() or release().
- The copy operation is inefficient and not a good idea.

The change was to simply allocate an array of NumCapture pointers.  As NDArrays arrive
their reference count is incremented with reserve() and the pointer is copied to the array.

After the files are written the cleanup routine now simply decrements the reference counter
with release(), rather than having to delete the NDArray.

The new scheme is much cleaner.  It will require setting the memory and array limits
for the NDArrayPool to be large enough to buffer NumCapture frames, whereas previously
this would not have been required.  It may thus require some changes to startup scripts.

### DIFF
--- a/ADApp/pluginSrc/NDPluginFile.cpp
+++ b/ADApp/pluginSrc/NDPluginFile.cpp
@@ -379,7 +379,7 @@ void NDPluginFile::freeCaptureBuffer(int numCapture)
     for (i=0; i<numCapture; i++) {
         pArray = this->pCapture[i];
         if (!pArray) break;
-        delete pArray;
+        pArray->release();
     }
     free(this->pCapture);
     this->pCapture = NULL;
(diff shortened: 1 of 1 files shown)


MarkRivers:
Mike Dunning at SLAC tested this yesterday with his Andor camera.  Previously saving TIFF files with Capture mode would issue the messages:
```
NDArray::release: WARNING, no owner
NDArray::reserve: WARNING, no owner
```
for each file.  It would also crash after saving a "handful" of files.  With this fix he was able to save 10000 files with no messages or errors.

PR #308, opened 2018-01-15
New NDArray allocation strategy

MichaelHuth:
The current implementation of the NDArray allocation in combination with the freelist selection poses some severe disadvantages when using small and very large NDArrays at the same time.

This PR uses another strategy, which is better suited for such cases while keeping reallocations of similar sized NDArrays fast.

Best regards,
Michael

### DIFF
--- a/ADApp/ADSrc/NDArrayPool.cpp
+++ b/ADApp/ADSrc/NDArrayPool.cpp
@@ -7,6 +7,7 @@
  */
 
 #include <stdlib.h>
+#include <stdint.h>
 
 #include <cantProceed.h>
 #include <epicsExport.h>
@@ -69,9 +70,9 @@ NDArray* NDArrayPool::alloc(int ndims, size_t *dims, NDDataType_t dataType, size
   epicsMutexLock(listLock_);
 
   /* Find a free image */
-  pArray = (NDArray *)ellFirst(&freeList_);
+  NDArray* freeArray = (NDArray *)ellFirst(&freeList_);
 
-  if (!pArray) {
+  if (!freeArray) {
     /* We did not find a free image.
      * Allocate a new one if we have not exceeded the limit */
     if ((maxBuffers_ > 0) && (numBuffers_ >= maxBuffers_)) {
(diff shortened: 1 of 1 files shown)


MichaelHuth:
Hello Mark,

I reviewed the formatting and set pre existing comments to their original format.
I also added a size equality check for the first array and another for the subsequent list scanning.

Best regards,
Michael

MarkRivers:
I am confused about how this code is supposed to work.  Imagine that there are currently 2 NDArrays on the free list, one with a size of 950, and one with a size of 1100.  NDArrayPool::alloc is called for an array whose size is 1000.  My expectation is that it should return the NDArray whose size is 1100, because it is big enough (>=1000) but less than the threshold (1500, i.e. 1.5 times 1000).  But as I trace the code I don't think this is what happens.  Because the 950 array is "closer" in size to 1000 than the 1100 array, it will select this one.  But then it will get further down and discover that 950 is not large enough and hence free its memory and reallocate 1000.  This seems inefficient?

I also don't understand the code in lines 171-172:
```
        size_t thresholdSize = dataSize + dataSize / 2;
        thresholdSize = (thresholdSize < dataSize) ? dataSize : thresholdSize;
```
The first line effective assigns threshold size to dataSize*1.5.  So I don't understand the need for the second line.  How can thresholdSize ever be less than dataSize?  Both are size_t so they are unsigned and cannot be negative.

It seems like the logic can be simplified.  On entry calculate datasize, and then search for an NDArray in the freelist whose size is >=datasize and <= datasize*threshold_ratio. If none is found then reallocate the array that was closest in size.  In the current PR there is lots of redundant code, basically initialization a new NDArray twice.

MichaelHuth:
Hello Mark,

your description is correct and I did not include the minor optimization to prefer arrays larger than the allocated size. However for our application the changes solved the main blocker, which was that 4 GB arrays stayed in the freeArray pool. I was aware of the limitation the closest array as choice had, but we never have this situation in our IOC.
Lines 171, 172:
The first line calculated the thresholdSize, but if dataSize is greater than MAX(size_t)/2 an unsigned overflow occurs and thresholdSize will be smaller than dataSize. The second line is a check for this overflow and sets thresholdSize to be at least dataSize. In the current commit I have changed it such that thresholdSize is SIZE_MAX in this case.

The current commit uses the logic you described where a found freeArray within dataSize and thresholdSize is preferred for reuse. I have also cleaned up the cases for freeArray / passed pArray / new pArray, where the array is initialized only at one location now.

Best regards,
Michael

MarkRivers:
I plan to merge this after ADCore R3-2 is released.  I want to wait because I plan to make additional changes with the NDArrayPool in R3-3.

PR #311, opened 2018-01-31
Return error on attempt to Capture in NDFileModeSingle so Capture is cleared.

bhill-slac:
Also added error message for user.

### DIFF
--- a/ADApp/pluginSrc/NDPluginFile.cpp
+++ b/ADApp/pluginSrc/NDPluginFile.cpp
@@ -423,8 +423,12 @@ asynStatus NDPluginFile::doCapture(int capture)
     switch(fileWriteMode) {
         case NDFileModeSingle:
             /* It is an error to set capture=1 in this mode, set to 0 */
+            asynPrint(this->pasynUserSelf, ASYN_TRACE_ERROR,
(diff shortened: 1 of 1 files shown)


bhill-slac:
To reproduce the bug, set a File plugin to Single mode and write 1 to it's Capture PV.
Capture isn't supported in Single mode and is quietly ignored, but Capture is left enabled as it's re-enabled by the code from pull request #100, b6d18f5e.
Change to Capture mode and iocsh spews errors like this:
2018/01/31 01:07:06.194 NDPluginFile::isFrameValid: no init array info was registered
2018/01/31 01:07:06.202 NDPluginFile::isFrameValid: no init array info was registered
2018/01/31 01:07:06.210 NDPluginFile::isFrameValid: no init array info was registered
2018/01/31 01:07:06.218 NDPluginFile::isFrameValid: no init array info was registered
2018/01/31 01:07:06.226 NDPluginFile::isFrameValid: no init array info was registered
2018/01/31 01:07:06.234 NDPluginFile::isFrameValid: no init array info was registered
2018/01/31 01:07:06.242 NDPluginFile::isFrameValid: no init array info was registered
...

PR #315, opened 2018-01-31
wider field for EPICS name PV

prjemian:
field is too narrow when converted to caQtDM .ui file


MarkRivers:
This got accidentally shortened when I change size->text_to_contents on all text graphics widgets.  Most of the widget in this part of the display are text update widgets (i.e. have PVs) and don't change when that is done.  But this is a plain text widget.

PR #318, opened 2018-02-02
Set severity fields for ArrayCounter, ArrayRate, EnableCallbacks, ...

bhill-slac:
Helps new users to see potential issues w/ their plugins by making all the enabled plugins that are in use and enabled be NO_ALARM, while disabled plugins are in MINOR state.

Not sure if you'll want to merge this as it is the opposite of current usage where enabled plugins are MINOR severity and disabled plugins are NO_ALARM.


MarkRivers:
I think this will confuse users since it reverses the longstanding conventions where active plugins are MINOR.  MINOR is used to indicate that resources are being used, NO_ALARM is no resources are being used.

PR #319, opened 2018-02-02
Add macro defaults ADDR=0 and TIMEOUT=1

bhill-slac:
Not sure if you want to merge this one.
We're careful to make sure our substitution files preserve these defaults
so they can still be overridden by a dbLoadRecords() call in st.cmd,
but don't require both to be defined in every dbLoadRecords() call for AD files.


MarkRivers:
I found that there were still a few template files in ADCore that did not always define the defaults ADDR=0 and TIMEOUT=1.  I have fixed these and pushed to Github.

Issue #307, opened 2018-01-12
PV to indicate when all arrays are done processing


tacaswell:
Discussed this with @MarkRivers at BNL today.

One problem with non-blocking callbacks is that when the acquire bit flips to low the arrays have not in general been processed by all of the plugins in the processing chain.  Knowing when the arrays have been fully-processed is important for higher-level sequencing.

It would be very useful to have a PV on the cam indicating when all of the arrays it has emitted into the processing pipeline have either been fully processed or dropped.

MarkRivers:
I've thought about this a bit.  It is not as easy as we first thought, where an additional counter would be added to the NDArray, indicating how many plugins were processing it, and when it got to 0 a PluginsDone record could process.  The problem is that plugins may not be processing the original array, but rather a copy of that array, which would have its own counter.

One solution I am thinking of is to combine this issue with the desire to move to a single NDArrayPool per driver, and plugins allocate from that same pool.  This solves 2 problems:
  - A single pool enables effective limits on memory usage.  Currently with multiple pools it is difficult to really limit the memory use because it is the total use that matters but there is no way to predict how much memory each pool will need.
- The single pool would be the object that could keep track of how many plugins are currently active for that driver, because it is the object they all have in common.  It is also associated with the driver, so the PluginsBusy PV will belong to that driver.  If there are multiple asynNDArray drivers in an IOC then each will have its own pool and its own PluginsBusy PV.

brunoseivam:
If I understand correctly, the problem is that in the non-blocking mode the _driver_ might be done producing arrays while the _plugin_ of interest is still processing the data. 

Wouldn't the following workaround work? At the beginning of an acquisition you can zero the plugin's `ArrayCounter` and `DroppedArrays` and then wait for `(driver's) Acquire == 0 && ArrayCounter_RBV + DroppedArrays_RBV == N`. This could even be done in a calc record.

mp49:
@brunoseivam That's how I've worked around this issue in the past. For example, if I have an ROI plugin that is read as soon as the acquire is completed by the camera driver it will almost always be an NDArray behind. But Marks proposal seems like a useful feature and a more generic solution.

tacaswell:
Watching the number of images on the plugins requires knowing which plugins to watch.  I would like to generically know when _all_ plugins are done without having to know anything about the topology of the processing chain at the next control layer up.

MarkRivers:
[...]

PR #323, opened 2018-05-23
Count number of NDArrays queued or processing in plugins

MarkRivers:
- Count number of NDArrays queued or processing in plugins. 
- Plugins no longer have their own NDArrayPool, they use the driver pNDArrayPool pointer in the NDArray passed to them.
- NDArrayPool->alloc() no longer uses maxBuffers, only maxMemory.

New records in ADBase.template:
- NumActivePlugins  The number of NDArrays processing or queued in plugins for this driver.
- WaitForPlugins This record determines whether AcquireBusy waits for NumActivePlugins to go to 0 before changing to 0 when acquisition completes.
- AcquireBusy This is a busy record that is set to 1 when Acquire changes to 1. It changes back to 0 when acquisition completes, i.e. when Acquire_RBV=0. If WaitForPlugins is Yes then it also waits for NumActivePlugins to go to 0 before changing to 0.

The Acquire record has changed from a busy record to a bo record with info(asyn:READBACK=1).  This is required to implement the logic described above.  ca_put_callback still works because Acquire has a FLNK to AcquireBusy.

### DIFF
--- a/ADApp/ADSrc/ADDriver.cpp
+++ b/ADApp/ADSrc/ADDriver.cpp
@@ -100,7 +100,7 @@ ADDriver::ADDriver(const char *portName, int maxAddr, int numParams, int maxBuff
                    int interfaceMask, int interruptMask,
                    int asynFlags, int autoConnect, int priority, int stackSize)
 
-    : asynNDArrayDriver(portName, maxAddr, maxBuffers, maxMemory,
+    : asynNDArrayDriver(portName, maxAddr, true, maxMemory,
           interfaceMask | asynInt32Mask | asynFloat64Mask | asynOctetMask | asynGenericPointerMask | asynDrvUserMask,
           interruptMask | asynInt32Mask | asynFloat64Mask | asynOctetMask | asynGenericPointerMask,
           asynFlags, autoConnect, priority, stackSize)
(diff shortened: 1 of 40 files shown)


MarkRivers:
This pull request fixes #307.

MarkRivers:
> A minor nitpick is that I think initializing pointers to NULL instead of 0 is preferable, since it makes clear that the variables being initialized are pointers. In this case, however, the variable name is already prefixed with p so it might not make a difference in the end.

I've looked at some style guides for help on this.  Google suggests using nullptr if the code is C++11 (which areaDetector is not) and NULL rather than 0 when coding for older standards.
https://google.github.io/styleguide/cppguide.html#0_and_nullptr/NULL

On the other hand Stroustrup recommends 0 rather than using the NULL macro:
http://www.stroustrup.com/bs_faq2.html#null

I'll try to use NULL since I do think it is clearer.

MarkRivers:
I changed the logic to improve backwards compatibility and simplicity.

- The asynNDArrayDriver constructor has reverted to the same as R3-2, with maxBuffers argument, rather than isDriver.  
- The asynNDArrayDriverconstructor always creates an NDArrayPool, as it always has.  It now assigns this to the pNDArrayPoolPvt_ pointer which always points to the pool created in the constructor.  The public pNDArrayPool pointer is initialized to pNDArrayPoolPvt_.
- Plugins will thus use their private NDArrayPool for allocating any NDArrays before the first callback.  This removes any possibility of problems with backwards compatibility, and allowed reverting the change made to NDPluginTimeSeries, which does allocate an NDArray before the first callback.
- Once the first callback occurs the NDPluginDriver base class sets the pNDArrayPool to the one passed in the NDArray, which will be the driver's pool.  Thus, after the first callback the plugin will allocate NDArrays from the driver's pool, which is what we want.

PR #328, opened 2018-06-18
Improvements to NDArrayPool

MarkRivers:
I changed NDArrayPool so the freeList is no longer a simple ellList, but is now an std::multiset.  The multiset is sorted by NDArray.dataSize.  This now makes it easy to find optimum size array on the freeList without doing a linear search.  It also makes it easy to see if there is no array that is large enough.

The asynReport command with details >5 will now print a report that includes the index, dataSize and pArray of each element of the freeList, as shown here:
```
epics> asynReport 6 SIM1
SIM1 multiDevice:No canBlock:No autoConnect:Yes
    enabled:Yes connected:Yes numberConnects 1
    nDevices 0 nQueued 0 blocked:No
    asynManagerLock:No synchronousLock:No
...

SIM1: private NDArrayPool report

NDArrayPool:
  numBuffers=112, numFree=105
  memorySize=105891216, maxMemory=0
  freeList: (index, dataSize, pArray)
    0 400 0x7f8ba400c420
    1 1024 0x7f8bcc0117c0
    2 65536 0x7f8bcc0119d0
    3 409600 0x7f8c08065860
    4 409600 0x7f8c0812f700
    5 409600 0x7f8b80402300
    6 409600 0x7f8b74136e90
    7 409600 0x7f8c08000990
    8 409600 0x7f8af8100c20
    9 409600 0x7f8c0812ea40
    10 409600 0x7f8b80200a10
    11 409600 0x7f8b7419bd90
    12 409600 0x7f8b74265a70
    13 409600 0x7f8b74008240
    14 409600 0x7f8b74200c00
    15 409600 0x7f8b740d1fe0
    16 409600 0x7f8c080ca7f0
    17 409600 0x7f8b7406d170
    18 1048576 0x7f8af1916490
    19 1048576 0x7f8b759e6be0
    20 1048576 0x7f8c082424e0
    21 1048576 0x7f8b76efbd10
    22 1048576 0x7f8b75ae7af0
    23 1048576 0x7f8b75fec550
    24 1048576 0x7f8b811054d0
    25 1048576 0x7f8b75be89d0
    26 1048576 0x7f8b770fe540
    27 1048576 0x7f8b748d75a0
    28 1048576 0x7f8b771ff440
    29 1048576 0x7f8b75ce98b0
    30 1048576 0x7f8b74ad9280
    31 1048576 0x7f8b773002e0
    32 1048576 0x7f8b75dea790
    33 1048576 0x7f8b774011e0
    34 1048576 0x7f8b775020a0
    35 1048576 0x7f8b749d8410
    36 1048576 0x7f8b74bda0f0
    37 1048576 0x7f8b77703df0
    38 1048576 0x7f8b75eeb670
    39 1048576 0x7f8b747d6730
    40 1048576 0x7f8b77602f90
    41 1048576 0x7f8b77804d00
    42 1048576 0x7f8b77a069b0
    43 1048576 0x7f8b758e5d70
    44 1048576 0x7f8b77b078c0
    45 1048576 0x7f8b74ddbdd0
    46 1048576 0x7f8b77c087a0
    47 1048576 0x7f8b802010b0
    48 1048576 0x7f8b77d09680
    49 1048576 0x7f8b74cdaf60
    50 1048576 0x7f8b77905b40
    51 1048576 0x7f8c08343370
    52 1048576 0x7f8b77e0a5c0
    53 1048576 0x7f8b77f0b400
    54 1048576 0x7f8b760eed50
    55 1048576 0x7f8b80202e80
    56 1048576 0x7f8b811057a0
    57 1048576 0x7f8b74edcc40
    58 1048576 0x7f8af02010e0
    59 1048576 0x7f8b761efc20
    60 1048576 0x7f8b74fddab0
    61 1048576 0x7f8af0301f80
    62 1048576 0x7f8af0402e80
    63 1048576 0x7f8b762f0af0
    64 1048576 0x7f8af0503d50
    65 1048576 0x7f8b750de920
    66 1048576 0x7f8b763f19c0
    67 1048576 0x7f8af0604c20
    68 1048576 0x7f8b751df790
    69 1048576 0x7f8af0705af0
    70 1048576 0x7f8b764f2890
    71 1048576 0x7f8af08069c0
    72 1048576 0x7f8af0907890
    73 1048576 0x7f8b752e0600
    74 1048576 0x7f8af0a08760
    75 1048576 0x7f8b765f3760
    76 1048576 0x7f8af0b09630
    77 1048576 0x7f8b753e1470
    78 1048576 0x7f8af0c0a500
    79 1048576 0x7f8b766f4630
    80 1048576 0x7f8af0d0b3d0
    81 1048576 0x7f8b755e3150
    82 1048576 0x7f8af0e0c2a0
    83 1048576 0x7f8b768f6410
    84 1048576 0x7f8b743d2d30
    85 1048576 0x7f8b744d3be0
    86 1048576 0x7f8af0f0d170
    87 1048576 0x7f8b81105970
    88 1048576 0x7f8b754e22e0
    89 1048576 0x7f8af110ee10
    90 1048576 0x7f8b769f72b0
    91 1048576 0x7f8af120fce0
    92 1048576 0x7f8b757e4ea0
    93 1048576 0x7f8af1310bb0
    94 1048576 0x7f8b76af8180
    95 1048576 0x7f8b756e3fe0
    96 1048576 0x7f8af1411a80
    97 1048576 0x7f8af1512950
    98 1048576 0x7f8b76bf9040
    99 1048576 0x7f8b746d58c0
    100 1048576 0x7f8af1613820
    101 1048576 0x7f8af17146f0
    102 1048576 0x7f8b76cf9f50
    103 1048576 0x7f8b745d4a50
    104 2097152 0x7f8b81105e50

SIM1: pAttributeList report

NDAttributeList: address=0x2ef3610:
  number of attributes=17
```
This should implement the same logic as the changes @MichaelHuth made, but in 80 fewer lines of code. 

It should also simplify the problem of derived classes for @sveseli because there is now no ellNodeOffset to worry about.

I hope that both @MichaelHuth and @sveseli can test this branch soon.

Please don't merge this branch until the tests are complete.


ulrikpedersen:
@ajgdls: can you please help me review this week? Mark is trying to get R3-3 out and we don't want to hold that up...

MarkRivers:
Added a new unit test, test_NDArrayPool to test the behavior of NDArrayPool::alloc().

Fixed some memory leaks, and fixed the behavior when freeList elements must be deleted to free up enough memory to allocate a new array.

ulrikpedersen:
Great, unit-tests was the first thing I was going to ask for 👍 

Travis is terrible for debugging... The crash on travis:

> unknown location(0): fatal error in "test_Pool": memory access violation at address: 0x032c6600: invalid permissions
> ../test_NDArrayPool.cpp(127): last checkpoint

... could just be because you're running out of memory? Could the unittest in total be allocating more than 2GB? https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System

MarkRivers:
The problem is that I am passing maxMemory=60000 to the asynNDArray driver constructor, which passes it to the NDArrayPool constructor. This works fine on my Centos 7 system. But on Travis maxMemory is not getting passed correctly. Very strange. I am throwing in more printf calls to track it down.

MarkRivers:
I have fixed the problem with the Travis tests.  The problem was that test_NDPluginCircularBuff.cpp and test_NDArrayPool.cpp both defined the test structure with the same name:
```
struct PluginFixture
{
    NDArrayPool *arrayPool;
...
    PluginFixture()
    {
        std::string dummy_port("simPort"), testport("testPort");
...
```
This actually works fine on my Centos 7 system.  However, on Travis these names clearly conflict because the test_NDArrayPool was not executing the initialization code in its PluginFixture() but was rather executing the code in the NDPluginCircularBuff PluginFixture().  That is why maxMemory was not being sent, and why it was calling the asynNDArrayConstructor twice, rather than just once.

ADApp/ADSrc/NDArray.cpp
```diff
@@ -119,8 +114,23 @@ int NDArray::getInfo(NDArrayInfo_t *pInfo)
...
+  return ND_SUCCESS;
+}
+/** Convenience method returns information about an NDArray, including the total number of elements, 
+  * the number of bytes per element, and the total number of bytes in the array.
+  \param[out] pInfo Pointer to an NDArrayInfo_t structure, must have been allocated by caller. */
+int NDArray::getInfo(NDArrayInfo_t *pInfo)
+{
+  int i;
+  NDAttribute *pAttribute;
+  size_t *dims_t = new size_t(this->ndims);
```
ajgdls: Instead of allocating memory here with new and then having to manage the deletion could we instead use std::vector on the stack?
Something like

```
std::vector<size_t> dims_t;
for (i=0; i<this->ndims; i++){
  dims_t.push_back(this->dims[i].size);
}
```

ajgdls: Review with @ulrikpedersen

ADApp/ADSrc/NDArray.h
```diff
@@ -124,6 +126,26 @@ class epicsShareClass NDArray {
...
 
+// This class defines the object that is contained in the std::multilist for sorting NDArrays in the freeList_.
+// It defines the < operator to use the NDArray::dataSize field as the sort key
+
+// We would like to hide this class definition in NDArrayPool.cpp and just forward reference it here.
+// That works on Visual Studio, and on gcc if instantiating plugins as heap variables with "new", but fails on gcc
+// if instantiating plugins as automatic variables.
+//class sortedListElement;
+
+class freeListElement {
```
ajgdls: Could we have a private default constructor to ensure we aren't accidentally creating invalid objects when using them with std::multiset (or any other container)?

ADApp/ADSrc/NDArray.h
```diff
@@ -124,6 +126,26 @@ class epicsShareClass NDArray {
...
+    public:
+        freeListElement(NDArray *pArray, size_t dataSize) {
+          pArray_ = pArray;
+          dataSize_ = dataSize;}
+        friend bool operator<(const freeListElement& lhs, const freeListElement& rhs) {
+            return (lhs.dataSize_ < rhs.dataSize_);
+        }
+        NDArray *pArray_;
+        size_t dataSize_;
+};
```
ajgdls: Discussed assignment operator and copy constructor but not necessary as the private members include a simple pointer and standard size_t type; default compiler generated methods will work.

ADApp/ADSrc/NDArrayPool.cpp
```diff
@@ -121,156 +98,97 @@ NDArray* NDArrayPool::getNextFreeArray(NDArray *pArray)
...
 
-  /* Find a free image */
-  freeArray = getFirstFreeArray();
+  // Compute the required NDArray size
+  NDArray::computeArrayInfo(ndims, dims, dataType, &arrayInfo);
+  if (dataSize == 0) {
+    dataSize = arrayInfo.totalBytes;
+  } 
+  else if (dataSize < arrayInfo.totalBytes) {
+      printf("%s: ERROR: required size=%d passed size=%d is too small, using required size\n",
```
ajgdls: Should this be an asynPrintf call or removed?

ADApp/ADSrc/NDArrayPool.cpp
```diff
@@ -121,156 +98,97 @@ NDArray* NDArrayPool::getNextFreeArray(NDArray *pArray)
...
       }
     }
-  }
-
-  if (pArray) {
-    /* If the caller passed a valid buffer use that, trust that its allocated size is correct */
-    if (pData) {
-      pArray->pData = pData;
+    if ((maxMemory_ > 0) && ((memorySize_ + dataSize) > maxMemory_)) {
+      printf("%s: error: reached limit of %ld memory (%d buffers)\n",
```
ajgdls: Should this be an asynPrintf call or removed?  Probably there is no asynUser available to pass to the asyn print calls...

ADApp/ADSrc/NDArrayPool.cpp
MarkRivers: There is no asynUser available in NDArrayPool.  There is potentially NDArrayPool::pDriver_, which points to an asynNDArrayDriver.  However, that is a forward referenced class in NDArrayPool.  Furthermore the asynNDArrayDriver::pasynUserSelf member is protected, so NDArrayPool does not have access to it.

ADApp/ADSrc/NDArrayPool.cpp
MarkRivers: Same comment as above.

ADApp/ADSrc/NDArray.h
MarkRivers: I have made the freeListElement constructor private, and made this a friend of the NDArrayPool class.

ADApp/ADSrc/NDArray.cpp
MarkRivers: I've changed to std::vector. It needs to be passed to computeArrayInfo as &dims_t[0], which is a bit ugly.

MarkRivers:
>General question, is this thread safe? The ellist is thread safe but the multiset stl container is not, should we be worried about this?

There is a mutex protecting all accesses to the multiset (freeList_).  There were already there in the alloc(), reserve(), and release() methods.  I just added them to getNumFree() and report().  Isn't this sufficient to make it thread safe?  The freeList_ is certainly being accessed from many threads, since there is now a single NDArrayPool that is used by all of the plugins.  I have seen no indication of bad behavior.

Thanks for the careful review.

ADApp/ADSrc/NDArray.cpp
ulrikpedersen: > I've changed to std::vector. It needs to be passed to computeArrayInfo as &dims_t[0], which is a bit ugly.

Ah yes, and if the vector is empty then the behaviour of dims_t[0] is undefined (could lead to some difficult bugs). You could go all-out STL and make computeArrayInfo work with a vector instead of a ptr...

ADApp/ADSrc/NDArrayPool.cpp
ulrikpedersen: In this case would it then be better to use the EPICS logging framework instead: errMessage, errPrintf, or epicsPrintf?

MarkRivers: I have made the changes in NDArray.h, asynNDArrayDriver.h, and NDArrayPool.cpp to use asynPrint rather than printf.   The unit test was already sufficient to test that this works.

ADApp/ADSrc/NDArray.cpp
MarkRivers: I tried using std::array rather than std::vector, which is much cleaner.  But it is only C++11 which we don't want to require (yet!).

brunoseivam: I think something like this would also work:

``` c++
size_t dims_t[this->ndims];
for (i=0; i<this->ndims; i++)
  dims_t[i] = this->dims[i].size;
```

ulrikpedersen: @brunoseivam that (i.e. initializing using a dynamic size: size_t dims_t[variable]) would only work with C++11...

ulrikpedersen:
> There is a mutex protecting all accesses to the multiset (freeList_). There were already there in the alloc(), reserve(), and release() methods. I just added them to getNumFree() and report(). Isn't this sufficient to make it thread safe?

Yes, that would do the job as long as we always remember to lock and unlock it when accessing it. A nice-to-have (and more C++ way of things, but not strictly required) would be to wrap the multiset in a class with built-in mutex ensuring locking/unlocking on *all* access.

ADApp/ADSrc/NDArray.h
ulrikpedersen: We actually meant the "default constructor" which is a constructor that takes no arguments and the compiler will implicitly define it if we do not do it. This implicitly defined constructor will not initialise the variables properly and the object will be invalid. If we declare it private then we are prevented from accidentally create objects like that.

https://en.cppreference.com/w/cpp/language/default_constructor

ADApp/ADSrc/NDArray.cpp
brunoseivam: It worked with `-std=c++98`, but it might be a gcc-ism...

MarkRivers: It compiles with gcc 4.8.5 without warning that C++11 is needed.  However, it fails to compile on VS2015 and VS2017.

MarkRivers: I am reverting to __new__ and __delete__ because I think it is more efficient and legible than the std::vector version.

ADApp/ADSrc/NDArray.h
MarkRivers: OK, I understand.  I have made the non-default constructor public and the default constructor private.

MarkRivers:
> Yes, that would do the job as long as we always remember to lock and unlock it when accessing it. A nice-to-have (and more C++ way of things, but not strictly required) would be to wrap the multiset in a class with built-in mutex ensuring locking/unlocking on all access.

The access to the std::multiset is in few enough places that I think its OK not to wrap it.

@ajgdls said that the ellList is thread-safe without using a mutex.  This may be true, but it is not documented as such in the EPICS Developer's Guide, which simply refers one to the vxWorks documentation.  Because it is not documented to be thread-safe I was already locking a mutex before all ellList calls.  I didn't have to change this when switching to the std::multiset.

MarkRivers:
@ajgdls said:
>General question, is this thread safe? The ellist is thread safe but the multiset stl container is not, should we be worried about this?

I looked up ellList in the 3.16 Application Developer's Guide.  It says this:
> ellLib.h describes a double linked list library. It provides functionality similar to the vxWorks lstLib library. See the vxWorks documentation for details. There is an ellXXX() routine to replace most vxWorks lstXXX() routines.

Note that it does not mention anything about thread safety.  However, it implies that it is compatible with the vxWorks lstLib functions.

The vxWorks lstLib documentation says the following:
> The linked-list library simply manipulates the linked-list data structures; no kernel functions are invoked. In particular, linked lists by themselves provide no task synchronization or mutual exclusion. If multiple tasks will access a single linked list, that list must be guarded with some mutual-exclusion mechanism (e.g., a mutual-exclusion semaphore).

So it explicitly says that lstLib is __not__ thread safe.

ajgdls:
Am I allowed to deflect the thread safety issue to @ulrikpedersen ?  Apologies for us stating the ellLib is thread safe, I've checked the class also and I see no protection so we were not correct in stating that.  We should have just asked to check that the calls are thread safe and they are.

ajgdls:
And to be clear I mean thread safe because you are locking the access, which you were already doing.

ulrikpedersen:
Whoops! Yes, you're right and I was wrong about ellList being thread safe. I'm sorry about that (and now I wonder what assumptions I've made in past implementations using ellList...)

PR #332, opened 2018-06-21
NDCodec

brunoseivam:
Due to our discussion on the last areaDetector workgroup meeting, I am proposing this set of changes to address compression in areaDetector.

This PR does four things:

1. Modifies `NDArray` and `NDArrayPool` to support two new fields in `NDArray`: `codec` and `compressedSize`.

2. Modifies `NDPluginDriver` to add a new parameter: `compressionAware`. This new parameter is default-constructed to `false`, and must be set to `true` by plugins that can deal with compressed NDArrays. NDArrays received by plugins that set `compressionAware` to `false` are dropped.

We could, instead of dropping arrays, do automatic decompression. See next point.

3. Introduces a new plugin, NDCodec, which can compress/decompress NDArrays and makes use of the above mentioned fields.

NDCodec's compression and decompression functions can be used by any other part of areaDetector, so it would be easy to do automatic decompression using them.

4. Modifies `ntndArrayConverter` to be able to properly convert compressed NDArrays. NDPva was also modified to set `compressionAware` to `true`.

Regarding the last point, there is a small issue that I'd want to discuss. While compressed `NTNDArrays` will have `ubyte` type, there is no natural field to hold the *original* (uncompressed) data type. This information is important on the decompression side. I solved this issue by using the `codec.parameters` field, which is of type "any" (variant union). I simply shoved a PVInt32 there and set its value to the `NDDataType_t` value of the original array.

However, I worry that this solution might not be future-proof. I'd like your opinion on this.

### Another approach

Another approach to implementing this functionality is to move the NDCodec parameters into the NDPluginDriver class. Then the compression functionality would be available automatically to all plugins. Arrays could be automatically decompressed by `beginProcessCallbacks` depending on a new flag passed to it (or maybe based on a NDPluginDriverAutoDecompress parameter) and compressed by `endProcessCallbacks` based on other parameters.


ulrikpedersen:
Wow I should go to meetings and mention crazy ideas more often! 😄 I’ve had this on the back burner for so long, I first started investigating blosc for another project in 2015!

So it looks like a good start but I’d like to look it over a bit more carefully before I comment - my last PR review had a few too many mistakes 😉

Thanks!

brunoseivam:
Hey guys, did you have a chance to take a look at this? This feature would be very useful for us at FRIB.

MarkRivers:
I think this approach is probably preferred to putting it in the NDPluginDriver base class.  I think the compress/decompress is likely to be used in specialized cases and not normally when passing NDArrays between plugins. 

The code looks good to me, but I will wait for Ulrik (and others) to comment before I merge.

MarkRivers:
I merged this pull request.  However, I ran into a number of problems.

1) NDPluginCodec.cpp did not compile on Windows with Visual Studio.  I have fixed this and pushed to master.

2) NDArrayPool::alloc() needs to set codec="" each time it allocates an array from the pool.  It not then there are errors that the NDArray is already compressed.  I have fixed this and pushed to master.

3) It often crashes when doing jpeg compression.  Here is a stack trace:
```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Core was generated by `../../bin/linux-x86_64/simDetectorApp st.cmd.linux'.
Program terminated with signal 6, Aborted.
#0  0x00007f0b6acfa277 in raise () from /lib64/libc.so.6
Missing separate debuginfos, use: debuginfo-install glibc-2.17-222.el7.x86_64 libX11-1.6.5-1.el7.x86_64 libXau-1.0.8-2.1.el7.x86_64 libXext-1.3.3-3.el7.x86_64 libgcc-4.8.5-28.el7_5.1.x86_64 libstdc++-4.8.5-28.el7_5.1.x86_64 libxcb-1.12-1.el7.x86_64 ncurses-libs-5.9-14.20130511.el7_4.x86_64 readline-6.2-10.el7.x86_64 sssd-client-1.16.0-19.el7_5.8.x86_64
(gdb) bt
#0  0x00007f0b6acfa277 in raise () from /lib64/libc.so.6
#1  0x00007f0b6acfbaa8 in abort () from /lib64/libc.so.6
#2  0x0000000000a1d578 in MagickPanicSignalHandler (signo=6) at ../magick.c:840
#3  <signal handler called>
#4  0x00007f0b6acfa277 in raise () from /lib64/libc.so.6
#5  0x00007f0b6acfb968 in abort () from /lib64/libc.so.6
#6  0x00007f0b6ad3cd37 in __libc_message () from /lib64/libc.so.6
#7  0x00007f0b6ad46c86 in _int_malloc () from /lib64/libc.so.6
#8  0x00007f0b6ad4984c in malloc () from /lib64/libc.so.6
#9  0x0000000000bdd968 in jpeg_get_small (cinfo=<optimized out>, sizeofobject=<optimized out>) at ../jmemnobs.c:37
#10 0x0000000000bdd7e8 in jinit_memory_mgr (cinfo=cinfo@entry=0x7f0af4f02a40) at ../jmemmgr.c:1060
#11 0x0000000000bc8bbf in jpeg_CreateCompress (cinfo=0x7f0af4f02a40, version=<optimized out>, structsize=584) at ../jcapimin.c:59
#12 0x0000000000627894 in compressJPEG (pool=pool@entry=0x2c51740, input=input@entry=0x7f0aac000950, quality=85) at ../NDPluginCodec.cpp:115
#13 0x000000000062883a in NDPluginCodec::processCallbacks (this=0x36e6770, pArray=0x7f0aac000950) at ../NDPluginCodec.cpp:481
#14 0x000000000064fd06 in NDPluginDriver::processTask (this=0x36e6770) at ../NDPluginDriver.cpp:485
#15 0x000000000119ff49 in epicsThreadCallEntryPoint (pPvt=0x371d910) at ../../src/osi/epicsThread.cpp:83
#16 0x00000000011a598c in start_routine (arg=0x371c1d0) at ../../src/osi/os/posix/osdThread.c:403
#17 0x00007f0b6bf09e25 in start_thread () from /lib64/libpthread.so.0
#18 0x00007f0b6adc2bad in clone () from /lib64/libc.so.6
(gdb)
```
I have not figured out this problem yet, so any help from @brunoseivam would be welcome.

I have also done the following:
- Added 2 NDPluginCodec plugins to EXAMPLE_commonPlugins.cmd, EXAMPLE_commonPlugin_settings.req, and commonPlugins.adl, .opi, .ui, .edl.
- Improved the formatting in NDCodec.adl.

brunoseivam:
Hi Mark,

Thanks for merging this! I will try to reproduce the issue here.

MarkRivers:
I believe I have found a simple way to reproduce the crash with simDetector:

image

image

image

image

- Configure the system as shown above
- Set the Codec1: plugin for Compressor=Blosc
- Press Acquire in simDetector.  It will acquire 10 images.
- Change the Codec1: Compressor to JPEG
- Press Acquire again.  It crashes every time for me.

MarkRivers:
I just pushed another fix in NDCodec_settings.req.  Changed PV name Algo to Compressor.

brunoseivam:
Hmmm, I am getting a different error as soon as I start JPEG compression. I wonder what changed. I had run a lot of tests back in August and these kind of issues didn't show up.

For this one in particular, I think the Codec code might be somehow tainting the NDArrays it uses and then simDetector crashes when it tries to reuse them. Looking into it.

```
epics> 
Thread 8 "SimDetTask" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff5136700 (LWP 31622)]
__memcpy_avx_unaligned () at ../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S:238
238	../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S: No such file or directory.
(gdb) bt
#0  __memcpy_avx_unaligned () at ../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S:238
#1  0x00000000005e1a50 in memcpy (__len=<optimized out>, __src=<optimized out>, __dest=<optimized out>)
    at /usr/include/x86_64-linux-gnu/bits/string3.h:53
#2  NDArrayPool::convert (this=0x12ddb00, pIn=0x7fff700008c0, ppOut=0x12ddc70, dataTypeOut=NDUInt8, 
    dimsOut=0x7ffff5135d00) at ../NDArrayPool.cpp:601
#3  0x000000000054ffa0 in simDetector::computeImage (this=this@entry=0x12d9d90) at ../simDetector.cpp:669
#4  0x00000000005509ba in simDetector::simTask (this=0x12d9d90) at ../simDetector.cpp:767
#5  0x0000000000d1149c in start_routine (arg=0x12e24b0) at ../../src/osi/os/posix/osdThread.c:403
#6  0x00007ffff79686ba in start_thread (arg=0x7ffff5136700) at pthread_create.c:333
#7  0x00007ffff67ab41d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
(gdb) 
```

MarkRivers:
I think I know why the behavior has changed.  Prior to ADCore R3-3 each plugin had its own NDArrayPool.  In R3-3 I changed it so that plugins get their NDArrays from the NDArrayPool created by the detector driver.  This allows effective enforcement of memory limits. This means that previously the simDetector was not getting NDArrays that had been modified by NDPluginCodec, but now it is.  

Could it have to do with the dataSize or compressedSize fields?

brunoseivam:
Yes, that's exactly what I am thinking.

Maybe the pool has to be smarter when dealing with NDArrays, maybe setting dataSize = compressedSize when reclaiming an NDArray? I'm investigating.

brunoseivam:
OK, I think I figured out my SEGFAULT (I didn't get to yours yet).

I caught libjpeg red-handed modifying pData's allocated buffer (see gdb output below). When that happens, `dataSize` is not valid anymore. When this buffer is returned to the pool, the pool is unaware that the underlying malloc'ed buffer is no longer `dataSize` long, and an attempt by `::convert` to write to this "tainted" buffer will fail with the SEGFAULT.

This doesn't happen every single time, only sometimes.

One way to fix it: detect when `pData` changes and then set `dataSize=compressedSize`.

One problem with this fix: if there's no `maxMemory` set for the pool, memory usage will grow indefinitely, apparently.


```
Thread 43 "CODEC1_Plugin_1" hit Hardware watchpoint 19: *0x7fffe0000b50

Old value = -596549616
New value = -136327152
term_mem_destination (cinfo=0x7ffff4782a00) at ../jdatadst.c:185
185	  *dest->outsize = dest->bufsize - dest->pub.free_in_buffer;
(gdb) bt
#0  term_mem_destination (cinfo=0x7ffff4782a00) at ../jdatadst.c:185
#1  0x0000000000a663fe in jpeg_finish_compress (cinfo=0x7ffff4782a00) at ../jcapimin.c:186
#2  0x00000000007fab5a in compressJPEG (pool=0x1849c30, input=0x7fff78000bd0, quality=85)
    at ../NDPluginCodec.cpp:243
#3  0x00000000007fb3ee in NDPluginCodec::processCallbacks (this=0x18f6910, pArray=0x7fff78000bd0)
    at ../NDPluginCodec.cpp:481
#4  0x000000000082b4b6 in NDPluginDriver::processTask (this=0x18f6910) at ../NDPluginDriver.cpp:485
#5  0x000000000082cb9e in NDPluginDriver::run (this=0x18f6910) at ../NDPluginDriver.cpp:866
#6  0x00000000011fb1ae in epicsThreadCallEntryPoint (pPvt=0x1949e20) at ../../src/osi/epicsThread.cpp:83
#7  0x0000000001203ed9 in start_routine (arg=0x1bfc3e0) at ../../src/osi/os/posix/osdThread.c:403
#8  0x00007ffff7bc16ba in start_thread (arg=0x7ffff4783700) at pthread_create.c:333
#9  0x00007ffff6a0441d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
```

MarkRivers:
I think the issue is jpeg_mem_dest.  It gets passed a char**, not a char*, so it is allowed to call malloc().
```
    jpeg_mem_dest(&jpegInfo,
                  (unsigned char **)&output->pData,
                  (unsigned long*) &output->compressedSize);
```
It looks like the correct solution is to write our own "destination manager" like I did in NDFileJPEG.cpp.  Then we have control over the buffer.

brunoseivam:
Alternatively, we can let libjpeg allocate the buffer itself and *then* give that buffer to `alloc`. This is what I am testing now.

I'll look into the destination manager stuff too.

brunoseivam:
Proposed fix on #348

MarkRivers:
I made 2 changes to NDPluginCode.cpp

The first uses pArray->pNDArrayPool for the pool, rather than the one private to the NDPluginCodec object.  This is what plugins should do now so memory limits can be enforced.

The second changes the definition of CompressionFactor.  Rather than going from 0 to 100, it is a true factor, i.e. originalSize/ compressedSize, so it goes from 1 to infinite.  I think it is easier to understand the  difference between 100 and 200 times smaller than it is between 99.0 and 99.5.

```
corvette:ADCore/ADApp/pluginSrc>git diff NDPluginCodec.cpp
diff --git a/ADApp/pluginSrc/NDPluginCodec.cpp b/ADApp/pluginSrc/NDPluginCodec.cpp
index 4139bb4..b2e7126 100644
--- a/ADApp/pluginSrc/NDPluginCodec.cpp
+++ b/ADApp/pluginSrc/NDPluginCodec.cpp
@@ -459,7 +459,7 @@ void NDPluginCodec::processCallbacks(NDArray *pArray)
     NDPluginDriver::beginProcessCallbacks(pArray);

     NDArray *result;
-    NDArrayPool *pool = pNDArrayPool;
+    NDArrayPool *pool = pArray->pNDArrayPool;

     int mode, algo;
     getIntegerParam(NDCodecMode, &mode);
@@ -509,8 +509,7 @@ void NDPluginCodec::processCallbacks(NDArray *pArray)
         if (result && result != pArray) {
             NDArrayInfo_t info;
             pArray->getInfo(&info);
-            factor = 1.0 - (1.0*result->compressedSize) / (1.0*info.totalBytes);
-            factor *= 100.0;
+            factor = (double)info.totalBytes / (double)result->compressedSize;
         }
     } else {
         if (pArray->codec.empty()) {
```

ADApp/Db/NDCodec.template
```diff
@@ -0,0 +1,178 @@
...
+}
+
+record(longin, "$(P)$(R)BloscCLevel_RBV")
+{
+    field(DTYP, "asynInt32")
+    field(INP,  "@asyn($(PORT),$(ADDR=0),$(TIMEOUT=1))BLOSC_CLEVEL")
+    field(SCAN, "I/O Intr")
+}
+
+record(bo, "$(P)$(R)BloscShuffle")
```
ulrikpedersen: This should probably be an mbbo as blosc now have 3 shuffle options: no shuffle, bit shuffle and byte shuffle.

ulrikpedersen: https://github.com/Blosc/c-blosc/blob/master/blosc/blosc.h#L51

brunoseivam: Agreed. I pushed that change: c66b5e6

PR #333, opened 2018-06-24
NDPluginStats time series now uses NDPluginTimeSeries

MarkRivers:
Reduced code by 240 lines and added significant new functionality


brunoseivam:
Hi Mark,

I know of a couple of beamlines @ NSLS-II that are (were?) using the Stats plugin time series occasionally.

@danielballan @tacaswell this might break AMX/FMX stuff when they upgrade. See [here](https://github.com/NSLS-II-FMX/profile_collection/blob/b1206217476442570c9608fa0c4310ee81ac9e1c/startup/95-custom_plans.py#L271).

PR #336, opened 2018-07-10
Fix SIZE_MAX error in NDArrayPool for GCC 4.4.7

jlmuir:
NDArrayPool uses SIZE_MAX, but in GCC 4.4.7 (the default compiler
on RHEL 6), SIZE_MAX is not defined by stdint.h in C++ mode unless
explicitly requested by defining __STDC_LIMIT_MACROS.  This could be
worked around by defining the __STDC_LIMIT_MACROS macro, but instead,
avoid this altogether by using std::numeric_limits.

### DIFF
--- a/ADApp/ADSrc/NDArrayPool.cpp
+++ b/ADApp/ADSrc/NDArrayPool.cpp
@@ -8,7 +8,7 @@
 
 #include <stdlib.h>
 #include <dbDefs.h>
-#include <stdint.h>
+#include <limits>
 
 #include <cantProceed.h>
 #include <epicsExport.h>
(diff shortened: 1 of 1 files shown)


jlmuir:
Hmm, it seems `NDArrayPool.cpp` has undergone significant change in commit c288362, so now this fix is no longer needed because `NDArrayPool.cpp` does not use `SIZE_MAX` anymore.  I'll close this PR.

BTW, `NDArrayPool.cpp` still includes `stdint.h`, but since it no longer uses `SIZE_MAX`, I don't think it needs that include.  (I haven't tried removing it, though.)

PR #347, opened 2018-10-18
DOC: add missing maxThreads argument to NDROIConfigure

mrakitin:
Noticed that issue while tried to configure the `maxThreads` parameter for detectors at the NSLS-II SMI beamline. The corresponding source code is: https://github.com/areaDetector/ADCore/blob/7fc9b3f1d5f7f81bfd11c552380d4af069b189ac/ADApp/pluginSrc/NDPluginROI.cpp#L367-L370.

I suspect similar update should be done for NDPluginStats.html, and some other files (leaving it up to the developers).


mrakitin:
Thank you for reviewing and merging, @MarkRivers!

PR #348, opened 2018-11-15
Fix issues with NDCodec

brunoseivam:
Follows discussion on #332


brunoseivam:
Travis is complaining, and it looks like a race condition in `asynNDArrayDriver::updateQueuedArrayCount`'s `epicsEventWait` call.

I am looking into it.

brunoseivam:
The introduction of compressed arrays changes the semantics a little bit. `dataSize` can be less than the calculated `totalBytes`. So I think the semantics should be:

* `dataSize` must refer to the size of pData buffer
* `arrayInfo.totalBytes` refers to the size of the *uncompressed* array
* `compressedSize` refers to the size of the *compressed* array

`dataSize` >= `compressedSize` should hold. The test I deleted prevented me from having `dataSize` < `totalBytes`.

PR #352, opened 2018-11-27
Codec improvements

MarkRivers:
Found and fixed a problem with NDPluginCodec.  It was not copying the NDAttributes from the input array to the output array.  Sometimes it worked: if the NDArray in the pool already had the attributes it worked, since they don't get removed.  But sometimes there were no attributes on the output array, including ColorMode. This led to problems in the ImageJ decompressor.  Changed alloc() so it adds all the metadata when the output array is created.

Changed errror handling from using fprintf to setting a string in a PV called CodecError.  This makes it easy for the user to see errors.  Added a new CodecStatus PV to show if the codec succeeded or failed. This required making some non-class functions into class methods so the errorMessage variable was accessible.

### DIFF
--- a/ADApp/ADSrc/NDArray.h
+++ b/ADApp/ADSrc/NDArray.h
@@ -162,6 +162,7 @@ public:
     NDArrayPool  (class asynNDArrayDriver *pDriver, size_t maxMemory);
     virtual ~NDArrayPool() {}
     NDArray*     alloc(int ndims, size_t *dims, NDDataType_t dataType, size_t dataSize, void *pData);
+    NDArray*     copy(NDArray *pIn, NDArray *pOut, bool copyData, bool copyDimensions, bool copyDataType);
     NDArray*     copy(NDArray *pIn, NDArray *pOut, int copyData);
 
     int          reserve(NDArray *pArray);
--- a/ADApp/ADSrc/NDArrayPool.cpp
+++ b/ADApp/ADSrc/NDArrayPool.cpp
@@ -220,7 +220,7 @@ NDArray* NDArrayPool::alloc(int ndims, size_t *dims, NDDataType_t dataType, size
   * object already exists (pOut!=NULL) then it must have sufficient memory allocated to
   * it to hold the data.
   */
-NDArray* NDArrayPool::copy(NDArray *pIn, NDArray *pOut, int copyData)
+NDArray* NDArrayPool::copy(NDArray *pIn, NDArray *pOut, bool copyData, bool copyDimensions, bool copyDataType)
 {
   //const char *functionName = "copy";
   size_t dimSizeOut[ND_ARRAY_MAX_DIMS];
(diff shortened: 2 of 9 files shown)


MarkRivers:
> Regarding errorMessage, I think I prefer your first solution: make the functions members of NDCodec. I originally didn't make them members because I wanted to leave open the option of these functions being used directly by other parts of areaDetector (automatic decompression, for instance). At this point, however, I think they should be member functions. When setting NDCodecCodecError, we can momentarily acquire and then release the lock, since arguably we're not in the hot path at that point anymore.

I don't think that will work.  Recall that NDPluginDriver now allows running multiple threads in the same plugin.  If errorMessage is a member variable then one thread could set it and before it is displayed another thread could clear or change it.  I think errorMessage really needs to be an automatic variable so it is private to that thread.

> If we go that route, we can also maybe stop passing the pool argument around and make it a member.

The pool actually now comes from the input array, so you are correct that we don't need to pass it, because we are already passing the input array.  But it does not need to be a member variable.

I don't like the code I had to add to the alloc() function, because it may need to change in the future if additional fields are added to NDArray.  What should be used is NDArrayPool->copy(pIn, pOut, copyData).  It currently almost does what we want if pOut is not NULL and copyData is 0.  However, it will currently copy the dataType, which we don't want.  I think there is an easy solution.  Make a new version of NDArrayPoolCopy(pIn, pOut, copyData, copyDimensions, copyDataType).  We will call it with copyData=0, copyDimensions=1, copyDataType=0.  The existing version will just call the new version with CopyDimensions=1, copyDataType=1, so it will be backwards compatible.

> Another thing that I thought about is if we should drop arrays that couldn't be compressed/decompressed instead of passing them along. Maybe that could be a toggle.

> Also, what should be the semantics of a second Compression plugin receiving already compressed data? Pass it along unmodified if the compression algo is the same? Drop it?

I think in both cases we should pass them with a warning.  I can change CodecStatus from bi to mbbi and add a third warning state?

brunoseivam:
> Recall that NDPluginDriver now allows running multiple threads in the same plugin

Oh, right! Nevermind that change, then

> I don't like the code I had to add to the alloc() function, because it may need to change in the future if additional fields are added to NDArray.

Agreed. We could get away with not modifying the current `NDArrayPool::copy` if you want, by calling `output->dataType = dt` right after `copy`.

> I think in both cases we should pass them with a warning. I can change CodecStatus from bi to mbbi and add a third warning state?

I guess it depends on where we want to drop unwanted arrays. If there's a downstream plugin expecting data with a specific format (say, if we decided to make NDFileJpeg use NDCodec), then NDFileJpeg will be the one dropping arrays.

MarkRivers:
I have pushed the changes I proposed in the previous comment.

Here are the results of a test showing the importance of allowing multiple threads in the NDPluginCodec.

image

image

Linear ramp simulation, Int32 data type, Gain=1.0, 1024x1024 array, AcquireTime=.001.  It is producing~550 arrays/s = 2.2GB/s.  Blosc ZLIB compression, comp. level=5, Shuffle=Byte.

Running on a 20-core Linux machine.

 Plugin threads | Blosc threads | Compressed frames/s 
 ------------       | ---------         |  ---------- 
 1   | 1 | 29 
 1 | 2 | 48 
 1 | 4 | 104 
1 | 8 | 189
1 | 12 | 213
1 | 16 | 215
2 | 1 | 58
2 | 2 | 110
2 | 4 | 196
2 | 8 | 270
2 | 12 | 262
2 | 16 | 248
4 | 1 | 118
4 | 2 | 209
4 | 4 | 296
4 | 6 | 306
4 | 8 | 305
5 | 1 | 149
5 | 2 |  240
5 | 3 | 297
5 | 4 | 316

So with only 1 plugin thread it can only do 215 frames/s, no matter how many blosc threads.  But by using multiple plugin threads this can be increased to 316 frames/s.

Note that this test was just to show scaling with threads.  By changing to LZ4 with Bit shuffle it can compress 1100 frames/s with the same compression factor (~40) as ZLIB with Byte shuffle.  This is compressing 4.4 GB/s.

MarkRivers:
I have added the third warning state.  All compress/decompress functions are now also passed a status argument that they can set to NDCODEC_ERROR or NDCODEC_WARNING.

I'm pretty happy with the current code.  @brunoseivam OK to merge?

brunoseivam:
Looks good to me!

PR #354, opened 2018-11-29
Rate limit plugin

brunoseivam:
This is a small plugin to limit the data rate / array rate flowing through it using the token bucket algorithm. This is useful to use in a plugin pipeline right before NDStdArray or NDPva to limit the amount of data going out to the network.


MarkRivers:
I think having the ability to limit rates is a good idea.  However, I wonder if making a new plugin is the best way to do this?  An alternative would be to add this logic to the NDPluginDriver base class.
* NDPluginDriver already has a MinCallbackTime parameter.  This is used to throttle the rate to a fixed time.
* We could rename that parameter to ThrottleValue and add a new parameter ThrottleMode.  The choices could be "None", "Min. Time", "Array Rate", and "Byte Rate".
* This way the control is available to all plugins, without the need to wire in a new plugin. If it is a new plugin how many of them do we load in commonPlugins.cmd?  1 is probably not enough because one might want to throttle both Channel Access and PVAccess.
* We can use the existing dropped arrays record, I don't think we need a new one.
* Do we need the diagnostic readback PVs? 

Mark

brunoseivam:
Hi Mark,

You're right. I thought the plugin would be a less invasive way of doing this, but I can see how all other plugins would benefit from it. I opened #355 to replace this PR.

PR #355, opened 2018-11-30
Implement rate limiting on NDPluginDriver

brunoseivam:
Replaces #354 .

PR #357, opened 2018-12-02
Improved rate limit logic

MarkRivers:
This is a new version of the rate limit logic.  It is based on #355 from @brunoseivam but with the changes described below.

* Keeps the existing MinCallbackTime on the master branch unchanged.
* Removes the ThrottleMode parameter.
* Replaces ThrottleValue with MaxByteRate. Added this to NDPluginDriverFull.* OPI screens.
* Changes throttled() to be protected rather than private so it can be used in derived classes.
* Changes throttled() to return false if MaxByteRate is 0.
* The test for throttled() is done in NDPluginDriver::endProcessCallbacks(),  because we are trying to throttle based on the size of the output data from the plugin, not the size of the input data.
* DroppedOutputArrays is incremented if throttling preventing output callbacks.

NDPluginPva and NDPluginStdArrays needed to be treated differently because the output we generally want to throttle is not the NDArray passed to downstream plugins, but rather the size of the output for the pvaServer (NDPluginPva) or the size of the arrays passed back to device support waveform records (NDPluginStdArrays).  For these plugins the throttling logic was thus also implemented inside the plugin. 

Note that for NDPluginStdArrays the original idea to use a new plugin to do the throttling would not have been effective.  The reason is that the output datatype can be different from the input.  For example the input might by type UInt8, but if the user has configured the database waveform record to be FTVL=DOUBLE then the ByteRate is actually 8 times larger then input data rate.  This implementation handles this correctly.

If the NDPluginPva and NDPluginStdArrays  plugins are throttled then they really do no useful work, and so ArrayCounter is not incremented (actually it is incremented in the base class and decremented in the derived class).  This makes the ArrayRate reflect the actual rate at which the plugin is doing useful work.  For NDPluginStdArrays this is also important because clients (e.g. ImageJ) monitor the ArrayCounter_RBV field to decide when to read the array and update the display.


MarkRivers:
Note that I have tested that setting MaxByteRate on NDPluginCodec, NDPluginPva, and  NDPluginStdArrays appears to work correctly.  DroppedOutputArrays increments when expected, and the ArrayRate for NDPluginPva and NDPluginStdArrays also behaves as expected.

GDYendell:
HI @MarkRivers and @brunoseivam 

I am having a problem with this change. On IOC boot I get the following errors:

```
dbLoadRecords 'db/simDetector_expanded.db'
Can't set "TESTSIMDETECTOR:ARR:MaxArrayRate_RBV.DTYP" to "Soft Channel"
Error at or before ")" in file "db/simDetector_expanded.db" line 3073
Can't set "TESTSIMDETECTOR:TIFF:MaxArrayRate_RBV.DTYP" to "Soft Channel"
ErrorCan't set "TESTSIMDETECTOR:JPEG:MaxArrayRate_RBV.DTYP" to "Soft Channel"
ErrorCan't set "TESTSIMDETECTOR:HDF:MaxArrayRate_RBV.DTYP" to "Soft Channel"
ErrorCan't set "TESTSIMDETECTOR:OVR:MaxArrayRate_RBV.DTYP" to "Soft Channel"
ErrorCan't set "TESTSIMDETECTOR:ROI:MaxArrayRate_RBV.DTYP" to "Soft Channel"
ErrorCan't set "TESTSIMDETECTOR:TRS:MaxArrayRate_RBV.DTYP" to "Soft Channel"
ErrorCan't set "TESTSIMDETECTOR:STAT:MaxArrayRate_RBV.DTYP" to "Soft Channel"
```

It looks to me that calc does not support setting DTYP:

```
$ caget -d31 TESTSIMDETECTOR:HDF:MaxArrayRate_RBV
TESTSIMDETECTOR:HDF:MaxArrayRate_RBV
    Native data type: DBF_DOUBLE
    Request type:     DBR_CTRL_ENUM
    Element count:    1
    Value:            Enum Index Overflow (0)
    Status:           NO_ALARM
    Severity:         NO_ALARM
    Enums:            ( 0)
```

I this is fixed by removing the DTYP field. I can then set and readback MaxArrayRate and the throttling works nicely. Might this be a problem with calc record support at DLS, or should the DTYP just not be set? I am testing ADSimDetector against ADCore R3-4 with EPICS base R3.12.14.7.

MarkRivers:
I just tested simDetector in 7.0.1 and I see the same errors.  I had not noticed them in previous testing because the were before iocInit and I did not scroll back.

The DTYP field is present in dbCommon.dbd in 3.14.12, 3.15.5, and 7.0.1:
```
corvette:/usr/local/epics>grep -H DTYP */dbd/dbCommon.dbd
base-3.14.12.6/dbd/dbCommon.dbd:        field(DTYP,DBF_DEVICE) {
base-3.15.5/dbd/dbCommon.dbd:   field(DTYP,DBF_DEVICE) {
base-7.0.1/dbd/dbCommon.dbd:    field(DTYP,DBF_DEVICE) {
```
However, no versions support the calc record in devSoft.dbd, only the calcout record.
```
corvette:/usr/local/epics>grep -H calc */dbd/devSoft.dbd
base-3.14.12.6/dbd/devSoft.dbd:device(calcout,CONSTANT,devCalcoutSoft,"Soft Channel")
base-3.14.12.6/dbd/devSoft.dbd:device(calcout,CONSTANT,devCalcoutSoftCallback,"Async Soft Channel")
base-3.15.5/dbd/devSoft.dbd:device(calcout,CONSTANT,devCalcoutSoft,"Soft Channel")
base-3.15.5/dbd/devSoft.dbd:device(calcout,CONSTANT,devCalcoutSoftCallback,"Async Soft Channel")
base-7.0.1/dbd/devSoft.dbd:device(calcout,CONSTANT,devCalcoutSoft,"Soft Channel")
base-7.0.1/dbd/devSoft.dbd:device(calcout,CONSTANT,devCalcoutSoftCallback,"Async Soft Channel")
base/dbd/devSoft.dbd:device(calcout,CONSTANT,devCalcoutSoft,"Soft Channel")
base/dbd/devSoft.dbd:device(calcout,CONSTANT,devCalcoutSoftCallback,"Async Soft Channel")
```

I have removed DTYP from MaxArrayRate_RBV in NDPluginBase.template and pushed to the master and add_bitshuffle branches.

I am hoping R3-5 will be tagged soon because we want to support bitshuffle/lz4 in ADEiger.  If not, then I will make a R3-4-1 release with this fix.

PR #361, opened 2018-12-10
remove class prefix

xiaoqiangwang:
It is probably a typo.


MarkRivers:
Thanks for spotting that.

Issue #281, opened 2017-06-28
Add Direct Chunk Write feature to HDF5 file writer plugin


ulrikpedersen:
The performance of the internal HDF5 library pipeline when doing basic I/O operations can be a bottleneck for high datarates (somewhere above 1GB/s). 

The HDF Group have implemented a way to circumvent this internal pipeline (and chunk cache) with the [Direct Chunk Write](https://support.hdfgroup.org/HDF5/doc/Advanced/DirectChunkWrite/) feature.

See also tech-talk discussion: http://www.aps.anl.gov/epics/tech-talk/2017/msg01034.php

We should add the Direct Chunk Write feature to the HDF5 file writer.

ulrikpedersen:
There are some subtleties to using Direct Chunk Write. It cannot use the internal HDF5 chunk cache so our chunk configuration would have to be restricted to fit within an individual frame. Otherwise we would have to implement our own pipeline with copies and chunk cache...

xiaoqiangwang:
Since that requires an pre-compressed NDArray, would it be a new NDPluginCompressor? That would bring it in line with NTNDArray specification.

ulrikpedersen:
> Since that requires a pre-compressed NDArray

No, it does not __require__ a pre-compressed NDArray. Direct Chunk Write works just as well (and with a performance increase) with uncompressed data.

> would it be a new NDPluginCompressor?

It doesn't have to be (see my comment just above) - but it is an interesting suggestion and I would like to see something like that.

> That would bring it in line with NTNDArray specification.

Again, that would be a good idea to have but there is no need to tightly couple the work of adding Direct Chunk Write functionality to changing the NDArray definition.

It is possible to "simply" add Direct Chunk Write functionality without coupling it with major changes to ADCore and NDArrays. I like the idea of small incremental changes to continuously improve performance/usability.

xiaoqiangwang:
Ok. I was assuming the file writer will be similar to the Dectris Eiger detector.

The compressor plugin would equally reduce the traffic when transferred though NDStdArray plugin.

ulrikpedersen:
That's effectively #349 

I agree this has sat around for long enough now without any real effort. I'll discuss priorities with @ajgdls, @GDYendell and @skyfrench.

GDYendell:
I am starting work on this now.

Issue #349, opened 2018-11-16
HDF5 Plugin to support compressed NDArrays


ulrikpedersen:
With the recent work from @brunoseivam on #332 it is now possible to compress NDArrays with a range of compression algorithms. The HDF5 file writer should be adapted to support this. This can be done using the HDF5 ["Direct Chunk Write"](https://confluence.hdfgroup.org/x/WNnoAg) feature where a single chunk can be written from a pre-compressed buffer and circumvent the HDF5 pipeline. Even with relatively small compression factors this could significantly increase file writer performance.

There are a number of things to consider:

When to create the dataset?
-----------------------------------

* When writing a pre-compressed buffer using Direct Chunk Write, the compression filter ID and associated parameters need to be set already when creating the dataset. Currently this information is only available once the first frame arrive in the HDF5 plugin - which could be too late as files are often created earlier.

* The chunking of the HDF5 dataset must be configured so that a full NDArray fits in a single chunk. Again, that is a requirement of Direct Chunk Write. The chunking configuration is set on dataset creation time.

These two issues could be solved in a couple of different ways:

* When using "Lazy Open" the HDF5 file is not created until the first frame arrives - so it can be interrogated for compression settings before creating the HDF5 dataset. The downside is that creating the file can be time consuming and is often best done prior to starting acquisition to avoid dropping the first few frames.
* When **not** using "Lazy Open" we could have parameters on the HDF5 plugin that allow pre-defining what compression settings we expect NDArrays to arrive with. We can use these to create the HDF5 dataset - but will then need to validate compression parameters on every incoming frame (we probably need to do that anyway because we can't change compression mid-way through writing a dataset).

HDF5 support for JPEG filter?
-------------------------------------

The Blosc library comes with a registered HDF5 filter plugin (ID: 32001). However, it is not clear (to me at least) if there is an equivalent to the JPEG compression - a number of JPEG filters are listed, but would they work with the particular JPEG implementation in libjpeg that NDCodec uses?  https://support.hdfgroup.org/services/filters.html

GDYendell:
Is this complete as of #365 @ulrikpedersen ?

ulrikpedersen:
Implemented in #365 

Thanks @GDYendell

PR #365, opened 2018-12-18
Direct Chunk Write

GDYendell:
This is an initial working version of direct chunk write support. I have tested with data direct from ADSimDetector and from NDPlugin Codec. 

The logic is as follows:
- HDF5 plugin settings match the incoming NDArrays (compressed or otherwise) - Use direct chunk write
- HDF5 chunking does not match, or compression configured and array not already compressed - Use standard write
- Data is compressed but does not match the HDF5 plugin (compression settings or chunking) - Error

Outstanding Points:
- I have tested the blosc compressors, but I haven't played with the level or shuffle yet
- I think the shuffle mode enums on HDF5 and NDPluginCodec are not in the same order, so that might need changing to allow validation of the NDArrays.
- I have not done any performance testing as I don't currently have an appropriate test server with a GPFS mount.

### DIFF
--- a/ADApp/pluginSrc/NDFileHDF5Dataset.cpp
+++ b/ADApp/pluginSrc/NDFileHDF5Dataset.cpp
@@ -156,31 +156,61 @@ asynStatus NDFileHDF5Dataset::extendDataSet(int extradims, hsize_t *offsets)
 }
 
 /**
- * Check if pArray dimensions match configured chunk dimensions
+ * Check if pArray dimensions and codec match hdf5 dataset definition
  * \param[in] pArray - The NDArray containing the data to verify.
  */
 asynStatus NDFileHDF5Dataset::verifyChunking(NDArray *pArray)
 {
+  // If compression is enabled, it must match our configuration
+  if (!pArray->codec.empty() || !this->codec.empty()) {
+    if (pArray->codec != this->codec) {
+      asynPrint(this->pAsynUser_, ASYN_TRACE_FLOW,
+                "Dataset codec '%s' [%d, %d, %d] does not match NDArray codec '%s' [%d, %d, %d]\n",
+                this->codec.name.c_str(), this->codec.level, this->codec.shuffle, this->codec.compressor,
+                pArray->codec.name.c_str(), pArray->codec.level, pArray->codec.shuffle, pArray->codec.compressor);
+      return asynError;
+    }
+  }
+  // Check dimensions of data
   if (pArray->ndims > 2) {
+    asynPrint(this->pAsynUser_, ASYN_TRACE_FLOW,
+              "Data must be 1D or 2D to use direct chunk write\n");
     return asynError;
   }
   // If chunk spans multiple frames (or multiple rows of a 1D dataset), then we require the HDF5
   // processing pipeline to stitch together the NDArrays
+  bool mismatch = false;
   if (this->chunkdims_[2] != 1 ||
       (pArray->ndims == 1 && this->chunkdims_[1] != 1)) {
-    return asynError;
+    mismatch = true;
   }
   if (this->chunkdims_[1] != 1 &&
       pArray->dims[0].size != this->chunkdims_[0]) {
-    return asynError;
+    mismatch = true;
   }
   // Remaining dimensions must match chunk definition
   for (int index = 0; index < pArray->ndims; index++) {
     if (pArray->dims[index].size != this->chunkdims_[index]) {
-      return asynError;
+      mismatch = true;
     }
   }
+  if (mismatch) {
+    asynPrint(this->pAsynUser_, ASYN_TRACE_FLOW,
+              "Dataset chunk dimensions [%d, %d, %d] do not match Array dimensions [%d, %d]\n",
+              (int)this->chunkdims_[0], (int)this->chunkdims_[1], (int)this->chunkdims_[2],
+              (int)pArray->dims[0].size, (int)pArray->dims[1].size);
+      return asynError;
+  }
   return asynSuccess;
+}
+
+/**
+ * Store codec definition
+ * \param[in] codec - Codec definition.
+ */
+void NDFileHDF5Dataset::configureCompression(Codec_t codec)
+{
+  this->codec = codec;
 }
 
 /** writeFile.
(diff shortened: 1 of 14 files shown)


GDYendell:
Fixes #281 #349

GDYendell:
This doesn't build because ADSupport uses hdf5 1.10.1. In this version the method to use is `H5DOwrite_chunk`. In 1.10.3+, this is deprecated (and the build uses `H5_NO_DEPRECATED_SYMBOLS`) and `H5Dwrite_chunk` was added. We will need to require <1.10.3, require >=1.10.3, or put in some version checks.

ulrikpedersen:
Thanks @GDYendell.

@brunoseivam @MarkRivers @ajgdls: obviously still a few points left todo before this is complete (as Gary mentions above) so please don't merge just yet, but feel free to test and review the solution.

> In this version the method to use is H5DOwrite_chunk. In 1.10.3+, this is deprecated (and the build uses H5_NO_DEPRECATED_SYMBOLS) and H5Dwrite_chunk was added. We will need to require <1.10.3, require >=1.10.3, or put in some version checks.

We discussed this today and this should be solved quite easily with a macro...

ulrikpedersen:
> It looks like the code is requiring the user to configure the HDF5 plugin codec settings to match those of the incoming NDArrays?
> The HDF5 plugin gets the correct settings for things like data type, dimensions, etc. from the last previous array (LazyOpen=false) or from the first array to be written (LazyOpen=true). Why can't it just get the codec settings the same way?

Yes, that's right. This is to support the LazyOpen=false scenario. We didn't think it would be safe to depend on the last previous array settings. Although if we do depend on the last previous array, then it would arguably make the interface simpler and we could just ignore the chunk and compression parameters on the HDF5 plugin...?

> Why does it need to know the Blosc shuffle type, for example, when using DirectChunkWrite? That information is not needed to decode the chunk, because that information is in the Blosc header.

We need to set the parameters used to compress the NDArray in the HDF5 filter params (i.e. the `cd_values[]` array. This includes shuffle, compressor algorithm and level. The HDF5 library doesn't know about the blosc header (or any other compression library header) so must be told explicitly what parameters were set.

Also, in a more abstract view, I think the NDArray **should** contain all the information about how it was encoded in a generic fashion. The proposed struct covers all cases of compression that we're currently discussing.

ulrikpedersen:
The HDF5 library is told about the compression parameters so that it can mark the dataset as compressed and store some meta data about how it was compressed. In the case of Blosc, you're correct @MarkRivers  that it is not strictly required to decompress the data - but it is used to make the HDF5 file accurately describe the dataset. See for example the output of h5dump:

```
[up45@pc0009 testdata]$ h5dump -pH excalibur_26291_000001.h5 
HDF5 "excalibur_26291_000001.h5" {
GROUP "/" {
   DATASET "data" {
      DATATYPE  H5T_STD_U16LE
      DATASPACE  SIMPLE { ( 3025, 1536, 2048 ) / ( H5S_UNLIMITED, 1536, 2048 ) }
      STORAGE_LAYOUT {
         CHUNKED ( 1, 1536, 2048 )
         SIZE 251936364 (75.542:1 COMPRESSION)  <--  Nice level of compression achieved here :-)
      }
      FILTERS {
         USER_DEFINED_FILTER {
            FILTER_ID 32001                   <--  Blosc filter ID
            PARAMS { 2 2 2 6291456 4 1 1 }    <--- Compression parameters (cd_values[7])
         }
      }
      FILLVALUE {
         FILL_TIME H5D_FILL_TIME_IFSET
         VALUE  0
      }
      ALLOCATION_TIME {
         H5D_ALLOC_TIME_INCR
      }
   }
}
}
```

GDYendell:
As Ulrik said, I believe the compression parameters must be passed to set_filter so that reader applications can seamlessly open the datasets without the user even realising that it is compressed, as long as the have the filter library on the path.

I do like the idea of making it support LazyOpen, though. In the same way the chunk parameters are set automatically from the frame dimensions, it could also just copy the codec from the NDArray and use that to create the dataset and call H5Pset_filter.

ulrikpedersen:
> I do like the idea of making it support LazyOpen, though. In the same way the chunk parameters are set automatically from the frame dimensions, it could also just copy the codec from the NDArray and use that to create the dataset and call H5Pset_filter.

@GDYendell: I thought it did! Let's discuss - I'm coming over :smiley:

ulrikpedersen:
We had another look at this and @GDYendell is looking into using the previous array pArray[0] to check for compression settings on the incoming arrays and if the incoming NDArrays are pre-compressed, use those settings to override the HDF5 plugin compression parameters (to give feedback on _RBV so users can see what compression is actually used). Then lazyOpen should automatically do the right thing both when set to true and false as lazyOpen is fully implemented in the parent class of the HDF5 plugin...

MarkRivers:
> In the case of Blosc, you're correct @MarkRivers that it is not strictly required to decompress the data -

That is actually true of all of the existing supported NDArray codecs (Blosc, JPEG, lz4, bitshuffle/lz4).  ImageJ can decompress all of them from NTNDArrays.  The only information needed is the name of the codec, and the data type of the uncompressed data, not any of the compression parameters.

I agree that it is nice for the NDArray to have all of the compression information.  It appears that all of the compression parameters in HDF5 are required to be of type "int", correct?  

How should we handle compression parameters in NTNDArrays?  Currently codec.parameters contains only a single integer, which is the data type of the uncompressed data.  Should we also store the compression parameters there?

ulrikpedersen:
> That is actually true of all of the existing supported NDArray codecs (Blosc, JPEG, lz4, bitshuffle/lz4). ImageJ can decompress all of them from NTNDArrays. The only information needed is the name of the codec, and the data type of the uncompressed data, not any of the compression parameters.

Yes, that's probably correct. I'm concerned about the HDF5 Dynamically Loadable Filters - it is down to each of these filters to decide how to decode the compressed data from a HDF5 dataset and I wouldn't want to make the assumption that none of these hdf5 filters use the information stored in the HDF5 filter parameters.

> I agree that it is nice for the NDArray to have all of the compression information. It appears that all of the compression parameters in HDF5 are required to be of type "int", correct?

Yes, the [H5Pset_filter](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetFilter) function takes an unsigned int array and number of parameters as input.

> How should we handle compression parameters in NTNDArrays? Currently codec.parameters contains only a single integer, which is the data type of the uncompressed data. Should we also store the compression parameters there?

I haven't looked into NTNDArrays much, but yes, I would suggest we should code the same information in there - just don't know enough about it to say how. We should have enough information in the NTNDArray to completely regenerate an exact copy of the NDArray at the other end - and be able to store that in a HDF5 file too.

:christmas_tree: :santa: By the way, we are about to break for the holidays now. I think @GDYendell has already left and I'm off in about half an hour. We must continue with this in the new year  :fireworks: Merry Christmas! :christmas_tree:

GDYendell:
I have added macros to protect against using write chunk in <1.8.9 and to call the correct write_chunk method. I have also added hdf5_hl to the Makefile to use H5DOwrite_chunk for <1.10.3.

MarkRivers:
Are you ready for me to merge this?

ulrikpedersen:
> Are you ready for me to merge this?

@GDYendell ?

ADApp/pluginSrc/NDFileHDF5Dataset.cpp
```diff
@@ -188,8 +254,43 @@ asynStatus NDFileHDF5Dataset::writeFile(NDArray *pArray, hid_t datatype, hid_t d
...
   }
+
   // Write the data to the hyperslab.
-  hdfstatus = H5Dwrite(this->dataset_, datatype, dataspace, fspace, H5P_DEFAULT, pArray->pData);
+  if (H5_VERSION_GE(1, 8, 11) && verifyChunking(pArray) == asynSuccess) {
+    // The chunking and compression settings match - use direct chunk write
+    asynPrint(this->pAsynUser_, ASYN_TRACE_FLOW,
+              "%s::%s NDArray correctly chunked. Using direct chunk write\n",
+              fileName, functionName);
+    #if H5_VERSION_GE(1, 8, 11)
```
ajgdls: You test for H5 version greater than 1.8.11 here and then inside the if statement you test again with a #define.  That code can never be executed.

GDYendell:
@MarkRivers I think this is ready now.

MarkRivers:
I found 2 problems with this after I merged:

- NDFileHDF5.cpp I needed to add {} to avoid a compiler error
- commonDriverMakefile needs hdf5_hl library for direct chunk write (on some HDF5 versions).  It had been added to commonLibraryMakefile but not to commonDriverMakefile.  pluginTests failed to build.

I have made these changes and pushed.

PR #369, opened 2019-01-10
Broke apart docs and fixed syntax errors

stuwilkins:
Follow on to #368 .. version wasn't pushed properly


stuwilkins:
Self merging to follow on from #368

stuwilkins:
Self merging to test build of docs in areaDetector/areaDetector

PR #374, opened 2019-01-29
Include Boost and plugin libraries

kgofron:


MarkRivers:
I had to add this line to ADCore/ci/install-packages.sh because the ADCore/ADApp/pluginTests requires WITH_BOOST=YES for the boost test framework, and that means that commonDriverMakefile now needs libboost_system.so
```
  sudo apt-get install libboost-system-dev
```

PR #376, opened 2019-02-04
Fix acquire busy2

### LINKED ISSUE #375: Problems with WaitForPlugins
ADCore R3-3 introduced the ability for drivers to wait for plugins to complete processing.  ca_put_callback to Acquire will not complete until both the driver and the plugins are done processing if WaitForPlugins is true.

However, there have been several problems with the implementation of this.
- In R3-3, R3-3-1, and R3-3-2 asynNDArrayDriver::incrementQueuedArrayCount and asynNDArrayDriver::incrementQueuedArrayCount were modifying the driver's parameter library (to update NumQueuedArrays) without taking the asynPortDriver lock.  This led to errors in the parameter library, typically manifested with drivers getting stuck with Acquire=1.  An IOC restart was required.

- Fixing this problem was not trivial because those functions cannot actually take the driver's asynPortDriver lock.  This is because those functions are called from the plugin threads, and deadlocks will arise because the plugins are called from the driver with the locks taken in the opposite order.

- R3-4 fixed the problem by using a separate mutex to just protect asynNDArrayDriver::queuedArrayCount_.  It creates a separate thread in ADDriver to update NumQueuedArrays in the parameter library and do the callbacks.  The increment and decrement functions just use this mutex when modifying the counter, and then send a signal to wake up this thread.

- We recently discovered a race condition with this R3-4 solution.  The database logic was waiting for both ADAcquire and NumQueuedArrays to be 0 if WaitForPlugins was true.  The problem was that ADAcquire could go to 0 before NumQueuedArrays had changed to 1.  This was due to the delay in waking the thread that updated NumQueuedArrays.  The net result was that the driver sometimes would not wait for the plugins to complete before ca_put_callback returned.  This showed up in scans using a statistics plugin, for example, as 2 points with duplicate values.  This was because the detector returned for the second point before the statistics plugin has posted a new value.

- I have tried various fixes for this.  It proved a bit of a challenge to fix the race condition while not introducing deadlocks or other issues.  I believe I now have a solution in the fixAcquireBusy2 branch.
I have tested the following with simDetector:
  - Very fast acquisition (1000 frames/s) in Continuous and Multiple mode. I can turn it off and back on without waiting for NumQueuedArrays to go to 0 first.  This was an issue with one of my test fixes.
  - Using Single mode with the sscan record and a statistics plugin as a counter.  The plugin has BlockingCallbacks=No, and the driver has WaitForPlugins=Yes.  The simDetector ADAcquireTime is 0.0001 seconds.
  - I did 1000 point scans many times and there are no duplicate values.  I also used the time series plugin in the stats plugin to collect the same array information, and compared to the array in the sscan record.  They were always identical.

MarkRivers:
This is intended to fix #375.

Moved the logic for determining when acquisition is complete, (including plugins if WaitForPlugins is true), from the database into ADDriver.

- This is simpler to understand, and more robust.
- Added 2 new parameters, ADWaitForPlugins and ADAcquireBusy.
- Overloaded asynPortDriver::setIntegerParam() method to trap changes to ADAcquire and NumQueuedArrays.  This function detects when acquisition is "really done" and sets ADAcquireBusy=0 when it is.
- Added new asynNDArrayDriver::getQueuedArrayCount() method that returns asynNDArrayDriver::queuedArrayCount_.  This is updated immediately on plugin callbacks, unlike NumQueuedArrays in the parameter library which has a slight delay.
- When ADAcquire goes to 0 asynPortDriver::setIntegerParam() must use getQueuedArrayCount() rather then NumQueuedArrays from the parameter library, because NumQueuedArrays is updated in a separate thread and might not have been set yet.  getQueuedArrayCount updates immediately.

I have tested the following with simDetector:
  - Very fast acquisition (1000 frames/s) in Continuous and Multiple mode. I can turn it off and back on without waiting for NumQueuedArrays to go to 0 first.  This was an issue with one of my test fixes.
  - Using Single mode with the sscan record and a statistics plugin as a counter.  The plugin has BlockingCallbacks=No, and the driver has WaitForPlugins=Yes.  The simDetector ADAcquireTime is 0.0001 seconds.
  - I did 1000 point scans many times and there are no duplicate values.  I also used the time series plugin in the stats plugin to collect the same array information, and compared to the array in the sscan record.  They were always identical.

### DIFF
--- a/ADApp/ADSrc/ADDriver.cpp
+++ b/ADApp/ADSrc/ADDriver.cpp
@@ -54,6 +54,43 @@ void ADDriver::setShutter(int open)
     }
 }
 
+/** Sets the value for an integer in the parameter library.
+  * \param[in] index The parameter number
+  * \param[in] value Value to set. 
+  * This function was added to trap the driver setting ADAcquire to 0 and
+  * asynNDArrayDriver setting NumQueuedArrays.  It implements the logic of
+  * setting ADAcquireBusy to reflect whether acquisition is done.
+  * If WaitForPlugins is true then this includes waiting for NumQueuedArrays to be 0.
+  * When ADAcquire goes to 0 it must use getQueuedArrayCount rather then NumQueuedArrays
+  * from the parameter library, because NumQueuedArrays is updated in a separate thread
+  * and might not have been set yet.  getQueuedArrayCount updates immediately. */
+asynStatus ADDriver::setIntegerParam(int index, int value)
+{
+    int waitForPlugins;
+
+    getIntegerParam(ADWaitForPlugins, &waitForPlugins);
+
+    if (index == ADAcquire) {
+        if (value == 0) {
+            if (waitForPlugins) {
+                int count = getQueuedArrayCount();
+                if (count == 0) {
+                    asynNDArrayDriver::setIntegerParam(ADAcquireBusy, 0);
+                }
+            } else {
+                asynNDArrayDriver::setIntegerParam(ADAcquireBusy, 0);
+            }
+        } else {
+            asynNDArrayDriver::setIntegerParam(ADAcquireBusy, 1);
+        }
+    }
+    else if ((index == NDNumQueuedArrays) && (value == 0)) {
+        asynNDArrayDriver::setIntegerParam(ADAcquireBusy, 0);
+    }
+    return asynNDArrayDriver::setIntegerParam(index, value);
+}
+
+
 /** Sets an int32 parameter.
   * \param[in] pasynUser asynUser structure that contains the function code in pasynUser->reason. 
   * \param[in] value The value for this parameter 
@@ -138,6 +175,8 @@ ADDriver::ADDriver(const char *portName, int maxAddr, int numParams, int maxBuff
     createParam(ADStatusString,              asynParamInt32, &ADStatus);
     createParam(ADTriggerModeString,         asynParamInt32, &ADTriggerMode);
     createParam(ADAcquireString,             asynParamInt32, &ADAcquire);
+    createParam(ADAcquireBusyString,         asynParamInt32, &ADAcquireBusy);
+    createParam(ADWaitForPluginsString,      asynParamInt32, &ADWaitForPlugins);
     createParam(ADShutterControlString,      asynParamInt32, &ADShutterControl);
     createParam(ADShutterControlEPICSString, asynParamInt32, &ADShutterControlEPICS);
     createParam(ADShutterStatusString,       asynParamInt32, &ADShutterStatus);
--- a/ADApp/ADSrc/ADDriver.cpp
+++ b/ADApp/ADSrc/ADDriver.cpp
@@ -55,6 +55,16 @@ void ADDriver::setShutter(int open)
 }
 
 /** Sets the value for an integer in the parameter library.
+  * Calls setIntegerParam(0, index, value) i.e. for parameter list 0.
+  * \param[in] index The parameter number
+  * \param[in] value Value to set. */
+asynStatus ADDriver::setIntegerParam(int index, int value)
+{
+    return this->setIntegerParam(0, index, value);
+}
+
+/** Sets the value for an integer in the parameter library.
+  * \param[in] list The parameter list number.  Must be < maxAddr passed to asynPortDriver::asynPortDriver.
   * \param[in] index The parameter number
   * \param[in] value Value to set. 
   * This function was added to trap the driver setting ADAcquire to 0 and
@@ -64,30 +74,30 @@ void ADDriver::setShutter(int open)
   * When ADAcquire goes to 0 it must use getQueuedArrayCount rather then NumQueuedArrays
   * from the parameter library, because NumQueuedArrays is updated in a separate thread
   * and might not have been set yet.  getQueuedArrayCount updates immediately. */
-asynStatus ADDriver::setIntegerParam(int index, int value)
+asynStatus ADDriver::setIntegerParam(int list, int index, int value)
 {
     int waitForPlugins;
 
-    getIntegerParam(ADWaitForPlugins, &waitForPlugins);
+    getIntegerParam(list, ADWaitForPlugins, &waitForPlugins);
 
     if (index == ADAcquire) {
         if (value == 0) {
             if (waitForPlugins) {
                 int count = getQueuedArrayCount();
                 if (count == 0) {
-                    asynNDArrayDriver::setIntegerParam(ADAcquireBusy, 0);
+                    asynNDArrayDriver::setIntegerParam(list, ADAcquireBusy, 0);
                 }
             } else {
-                asynNDArrayDriver::setIntegerParam(ADAcquireBusy, 0);
+                asynNDArrayDriver::setIntegerParam(list, ADAcquireBusy, 0);
             }
         } else {
-            asynNDArrayDriver::setIntegerParam(ADAcquireBusy, 1);
+            asynNDArrayDriver::setIntegerParam(list, ADAcquireBusy, 1);
         }
     }
     else if ((index == NDNumQueuedArrays) && (value == 0)) {
-        asynNDArrayDriver::setIntegerParam(ADAcquireBusy, 0);
+        asynNDArrayDriver::setIntegerParam(list, ADAcquireBusy, 0);
     }
-    return asynNDArrayDriver::setIntegerParam(index, value);
+    return asynNDArrayDriver::setIntegerParam(list, index, value);
 }
 
 
--- a/ADApp/ADSrc/ADDriver.h
+++ b/ADApp/ADSrc/ADDriver.h
@@ -148,6 +148,7 @@ public:
     /* These are the methods that we override from asynPortDriver */
     virtual asynStatus writeInt32(asynUser *pasynUser, epicsInt32 value);
     virtual asynStatus setIntegerParam(int index, int value);
+    virtual asynStatus setIntegerParam(int list, int index, int value);
 
     /* These are the methods that are new to this class */
     virtual void setShutter(int open);
--- a/ADApp/ADSrc/ADDriver.h
+++ b/ADApp/ADSrc/ADDriver.h
@@ -109,6 +109,8 @@ typedef enum
 #define ADAcquirePeriodString       "ACQ_PERIOD"            /**< (asynFloat64,  r/w) Acquisition period between images */
 #define ADStatusString              "STATUS"                /**< (asynInt32,    r/o) Acquisition status (ADStatus_t) */
 #define ADAcquireString             "ACQUIRE"               /**< (asynInt32,    r/w) Start(1) or Stop(0) acquisition */
+#define ADAcquireBusyString         "ACQUIRE_BUSY"          /**< (asynInt32,    r/w) 0 when acquire done including plugins */
+#define ADWaitForPluginsString      "WAIT_FOR_PLUGINS"      /**< (asynInt32,    r/w) Wait for plugins before AcquireBusy goes to 0 */
 
     /* Shutter parameters */
 #define ADShutterControlString      "SHUTTER_CONTROL"       /**< (asynInt32,    r/w) (ADShutterStatus_t) Open (1) or Close(0) shutter */
@@ -145,6 +147,7 @@ public:
 
     /* These are the methods that we override from asynPortDriver */
     virtual asynStatus writeInt32(asynUser *pasynUser, epicsInt32 value);
+    virtual asynStatus setIntegerParam(int index, int value);
 
     /* These are the methods that are new to this class */
     virtual void setShutter(int open);
@@ -179,6 +182,8 @@ protected:
     int ADStatus;
     int ADTriggerMode;
     int ADAcquire;
+    int ADAcquireBusy;
+    int ADWaitForPlugins;
     int ADShutterControl;
     int ADShutterControlEPICS;
     int ADShutterStatus;
--- a/ADApp/ADSrc/asynNDArrayDriver.cpp
+++ b/ADApp/ADSrc/asynNDArrayDriver.cpp
@@ -708,7 +708,6 @@ static void updateQueuedArrayCountC(void *drvPvt)
 
 void asynNDArrayDriver::updateQueuedArrayCount()
 {
-    int arrayCount;
     while (queuedArrayUpdateRun_) {
         epicsEventWait(queuedArrayEvent_);
         // Exit early
@@ -716,14 +715,19 @@ void asynNDArrayDriver::updateQueuedArrayCount()
             break;
 
         lock();
-        queuedArrayCountMutex_->lock();
-        arrayCount = queuedArrayCount_;
-        queuedArrayCountMutex_->unlock();
-        setIntegerParam(NDNumQueuedArrays, arrayCount);
+        setIntegerParam(NDNumQueuedArrays, getQueuedArrayCount());
         callParamCallbacks();
         unlock();
     }
     epicsEventSignal(queuedArrayUpdateDone_);
+}
+
+int asynNDArrayDriver::getQueuedArrayCount()
+{
+    queuedArrayCountMutex_->lock();
+    int count = queuedArrayCount_;
+    queuedArrayCountMutex_->unlock();
+    return count;
 }
 
 asynStatus asynNDArrayDriver::incrementQueuedArrayCount() 
--- a/ADApp/ADSrc/asynNDArrayDriver.h
+++ b/ADApp/ADSrc/asynNDArrayDriver.h
@@ -143,6 +143,7 @@ public:
 
     asynStatus incrementQueuedArrayCount();
     asynStatus decrementQueuedArrayCount();
+    int getQueuedArrayCount();
     void updateQueuedArrayCount();
     
     class NDArrayPool *pNDArrayPool;     /**< An NDArrayPool pointer that is initialized to pNDArrayPoolPvt_ in the constructor.
--- a/ADApp/Db/ADBase.template
+++ b/ADApp/Db/ADBase.template
@@ -459,17 +459,27 @@ record(busy, "$(P)$(R)AcquireBusy") {
    field(VAL,  "0")
 }
 
-record(calcout, "$(P)$(R)ClearAcquireBusy")
-{
-    field(INPA, "$(P)$(R)WaitForPlugins CP")
-    field(INPB, "$(P)$(R)Acquire_RBV CP")
-    field(INPC, "$(P)$(R)NumQueuedArrays CP")
-    field(CALC, "A ? B || (C > 0) : B")
+record(bi, "$(P)$(R)AcquireBusyCB") {
+   field(DTYP, "asynInt32")
+   field(INP, "@asyn($(PORT),$(ADDR=0),$(TIMEOUT=1))ACQUIRE_BUSY")
+   field(ZNAM, "Done")
+   field(ZSV,  "NO_ALARM")
+   field(ONAM, "Acquiring")
+   field(OSV,  "MINOR")
+   field(SCAN, "I/O Intr")
+}
+
+record(calcout, "$(P)$(R)ClearAcquireBusy") {
+    field(INPA, "$(P)$(R)AcquireBusyCB CP")
+    field(CALC, "A")
     field(OOPT, "Transition To Zero")
     field(OUT,  "$(P)$(R)AcquireBusy PP")
 }
 
 record(bo, "$(P)$(R)WaitForPlugins") {
+   field(PINI, "YES")
+   field(DTYP, "asynInt32")
+   field(OUT, "@asyn($(PORT),$(ADDR=0),$(TIMEOUT=1))WAIT_FOR_PLUGINS")
    field(ZNAM, "No")
    field(ONAM, "Yes")
    field(VAL,  "0")
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -31,6 +31,11 @@ R3-5 (February XXX, 2018)
     html in the files.
   * The new documentation is hosted at [areaDetector.github.io](https://areaDetector.github.io).
   * Many thanks to Stuart Wilkins for this major effort.
+### ADDriver, asynNDArrayDriver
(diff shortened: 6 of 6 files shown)


ulrikpedersen:
@thomascobb could you please take a look at the changes here and consider the logic for determining when the acquisition is complete?

@skyfrench, @GDYendell could you please help with review, building & testing it?

MarkRivers:
My suggestions for @thomascobb, @skyfrench, @GDYendell testing:

- First test with R3-4 to reproduce the problem
- Then test with this branch to see if it is fixed

I tested with the following simDetector settings:
image

image

The NDPluginStats settings
image

Note that I set 1000 points for the TimeSeries and started TimeSeries acquisition

And the following sscan settings.  An sscan record is loaded in commonPlugins.cmd.
image

When you press the SCAN button it collects 1000 frames, and has a detector which is the net counts in the statistics plugin.

To see if there is a problem I read both the sscan record array into variable "v" and the stats time series array into variable "x" in IDL;

IDL> t = caget('13SIM1:scan1.D01CA', v, max=1000)
IDL> t = caget('13SIM1:Stats5:TSNet', x, max=1000)

I then computed an array of indices "b" where v[N] and v[N+1] were identical, i.e. duplicate values which which not occur.  b is -1 which means there are no duplicate points.
IDL> b = where(v-shift(v,1) eq 0)
IDL> print, b
          -1

I also computed the indices in "b2" where x and v are different.  There are no differences, which is also what is expected if the bug is fixed.
IDL> b2 = where(v-x ne 0)
IDL> print, b2
          -1

thomascobb:
This looks like the correct logic, and it's definitely more readable in code than in database logic. I'll leave it to Sky and Gary to do some testing...

A general question, why is updateQueuedArrayCount() in a different thread the other side of the event loop? Is it so plugins do not need to take the lock() on the driver?

MarkRivers:
> A general question, why is updateQueuedArrayCount() in a different thread the other side of the event loop? Is it so plugins do not need to take the lock() on the driver?

I'm not sure what you mean by "the other side of the event loop"?

updateQueuedArrayCount was put in a different thread in R3-4 to avoid deadlocks.  If it is not in a separate thread then incrementQueuedArrayCount and decrementQueuedArrayCount need to take the driver lock to access the parameter library.  That leads to deadlocks because when the driver calls the plugin to queue a new array it takes locks in order:
1. driver lock
2. plugin lock

When the plugins call the driver to increment/decrement the QueuedArrayCount they take the locks in the order:

1. plugin lock
2. driver lock

Because the locks are taken in the opposite order this leads to deadlocks.

thomascobb:
> I'm not sure what you mean by "the other side of the event loop"?

I should have said "listening to an epicsEvent" rather than being in the code that sets the epicsEvent.

Your description of the ordering of locks makes sense as to why this is the case.

GDYendell:
I have done some testing with the following:
- SimDetector with Mark's above settings (except using 0.005s acquire period, the stats plugin dropped frames going any faster), Single mode
- sscan (via generic module 1D scan) running scans of 1000 points. Trigger: TESTSIMDETECTOR:CAM:Acquire, Det 1: TESTSIMDETECTOR:STAT:MinX_RBV (I found that Net_RBV was the same for every frame). sscan presses Acquire, waits for Acquire go to 0, pulls a value, repeated 1000 times
- Python script poking sscan, reading D01CA and checking for duplicates next to each other, repeating forever

Results:
- With R3-4 I get ~100 duplicates per 1000 frame scan (often with a missing value after the duplication)
- With the latest master I get 0 duplicates, over 100s of runs.

MarkRivers:
Thanks for doing the testing.  NSLS-II was where this problem was observed first, on ADProsilica detectors.  They were planning to test with those yesterday.  I'll find out if they also found the issue to be fixed, and if so I'll make a new release soon.

GDYendell:
I think I have had a deadlock in my IOC after running acquisitions for a few hours. My epics console is still active and I can caget and caput, but my screens have all gone invalid. I have attached a debugger and I can see one thread is here:

```
void asynNDArrayDriver::updateQueuedArrayCount()
{
    while (queuedArrayUpdateRun_) {
        epicsEventWait(queuedArrayEvent_);
        // Exit early
        if (!queuedArrayUpdateRun_)
            break;

        lock();   <----------
        setIntegerParam(NDNumQueuedArrays, getQueuedArrayCount());
        callParamCallbacks();
        unlock();
    }
    epicsEventSignal(queuedArrayUpdateDone_);
}
```

I'm not sure which of the other threads in the clash. It might be this:

```
extern "C" {static asynStatus writeInt32(void *drvPvt, asynUser *pasynUser, 
                            epicsInt32 value)
{
    asynPortDriver *pPvt = (asynPortDriver *)drvPvt;
    asynStatus status;
    
    pPvt->lock();   <----------
    status = pPvt->writeInt32(pasynUser, value);
    pPvt->unlock();
    return(status);
}}
```

The rest of the threads seem to just be on sleep, condWait, dbScanLock or recv.

MarkRivers:
Darn, sorry to hear that.

In order for a thread to deadlock it has to be trying to lock 2 mutexes.  

updateQueuedArrayCount takes the driver lock and then the ringBufferLock in devAsynInt32.c.  That should not cause a deadlock because no thread should ever take those locks in the opposite order.

You said some threads were blocking on dbScanLock.  Those are the ones I suspect.  Can you see what those threads are doing, i.e. what the thread name is and what record is being processed?  

How can I try to reproduce the problem?  What driver and plugins were running, and with what settings?

GDYendell:
There are four similar threads calling dbScanLock. They all have the same stack (except for the addresses): 

```
#0  0x00007f9b061454ed in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x00007f9b06140de6 in _L_lock_941 () from /lib64/libpthread.so.0
#2  0x00007f9b06140cdf in pthread_mutex_lock () from /lib64/libpthread.so.0
#3  0x00007f9b0738e596 in mutexLock (id=<optimized out>) at ../../../src/libCom/osi/os/posix/osdMutex.c:44
#4  epicsMutexOsdLock (pmutex=0x2c496b0) at ../../../src/libCom/osi/os/posix/osdMutex.c:116
#5  0x00007f9b07c4cb0b in dbScanLock (precord=0x255b700) at ../dbLock.c:264
#6  0x00007f9b07c61851 in db_get_field_and_count (paddr=paddr@entry=0x7f9a5c007f88, buffer_type=14, pbuffer=0x7f9a8c020f50, nRequest=nRequest@entry=0x7f9a7b40acf0, pfl=pfl@entry=0x0) at ../db_access.c:185
#7  0x00007f9b0829c3b3 in read_reply (pfl=0x0, eventsRemaining=1, paddr=0x7f9a5c007f88, pArg=0x7f9a7b40ad10) at ../camessage.c:565
#8  read_notify_action (mp=<optimized out>, pPayload=<optimized out>, client=<optimized out>) at ../camessage.c:764
#9  0x00007f9b0829d452 in camessage (client=client@entry=0x7f9a88000ae0) at ../camessage.c:2594
#10 0x00007f9b0829a060 in camsgtask (pParm=0x7f9a88000ae0) at ../camsgtask.c:116
#11 0x00007f9b0738c75f in start_routine (arg=0x7f9a8c043380) at ../../../src/libCom/osi/os/posix/osdThread.c:389
#12 0x00007f9b0613edd5 in start_thread () from /lib64/libpthread.so.0
#13 0x00007f9b06864ead in clone () from /lib64/libc.so.6
```

The precord.name in dbScanLock is $(P)$(R)Acquire. The pthreadinfo.name in start_routine is CAS-client.

One other also stuck waiting:

```
#0  0x00007f9b061454ed in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x00007f9b06140de6 in _L_lock_941 () from /lib64/libpthread.so.0
#2  0x00007f9b06140cdf in pthread_mutex_lock () from /lib64/libpthread.so.0
#3  0x00007f9b0738e596 in mutexLock (id=<optimized out>) at ../../../src/libCom/osi/os/posix/osdMutex.c:44
#4  epicsMutexOsdLock (pmutex=0x2c496b0) at ../../../src/libCom/osi/os/posix/osdMutex.c:116
#5  0x00007f9b07c4cb0b in dbScanLock (precord=0x255b700) at ../dbLock.c:264
#6  0x00007f9b08b23b08 in interruptCallbackOutput (drvPvt=0x296d3b0, pasynUser=0x296dfa8, value=0) at ../../asyn/devEpics/devAsynInt32.c:601
#7  0x00007f9b08b18cea in paramList::int32Callback (this=this@entry=0x2415010, command=command@entry=79, addr=addr@entry=0) at ../../asyn/asynPortDriver/asynPortDriver.cpp:583
#8  0x00007f9b08b191fa in paramList::callCallbacks (this=0x2415010, addr=0) at ../../asyn/asynPortDriver/asynPortDriver.cpp:747
#9  0x00007f9b0947b92d in simDetector::simTask (this=0x2341d00) at ../simDetector.cpp:823
#10 0x00007f9b0738c75f in start_routine (arg=0x24442f0) at ../../../src/libCom/osi/os/posix/osdThread.c:389
#11 0x00007f9b0613edd5 in start_thread () from /lib64/libpthread.so.0
#12 0x00007f9b06864ead in clone () from /lib64/libc.so.6
```

And this one has taken the lock:

```
#0  0x00007f9b061454ed in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x00007f9b06140de6 in _L_lock_941 () from /lib64/libpthread.so.0
#2  0x00007f9b06140cdf in pthread_mutex_lock () from /lib64/libpthread.so.0
#3  0x00007f9b0738e596 in mutexLock (id=<optimized out>) at ../../../src/libCom/osi/os/posix/osdMutex.c:44
#4  epicsMutexOsdLock (pmutex=0x243dfa0) at ../../../src/libCom/osi/os/posix/osdMutex.c:116
#5  0x00007f9b08af37a4 in queueRequest (pasynUser=0x28ce3b8, priority=<optimized out>, timeout=0) at ../../asyn/asynDriver/asynManager.c:1533
#6  0x00007f9b08b3086b in processAi (pr=0x254cc40) at ../../asyn/devEpics/devAsynFloat64.c:590
#7  0x00007f9b088b8b08 in readValue (prec=0x254cc40) at ../aiRecord.c:432
#8  process (precord=0x254cc40) at ../aiRecord.c:167
#9  0x00007f9b07c4f18e in dbProcess (precord=precord@entry=0x254cc40) at ../dbAccess.c:640
#10 0x00007f9b07c5a355 in scanList (psl=0x2c8a6e0) at ../dbScan.c:750
#11 0x00007f9b07c5a620 in periodicTask (arg=0x2c8a6e0) at ../dbScan.c:601
#12 0x00007f9b0738c75f in start_routine (arg=0x2d86240) at ../../../src/libCom/osi/os/posix/osdThread.c:389
#13 0x00007f9b0613edd5 in start_thread () from /lib64/libpthread.so.0
#14 0x00007f9b06864ead in clone () from /lib64/libc.so.6
```

but is then stuck on epicsMutexMustLock with pport.portName CAM1.CAM (simDetector). In processAi the record is $(P)$(R)PoolUsedMem.

GDYendell:
This is the script I was running, which includes some of the settings:

```
caput("TESTSIMDETECTOR:CAM:WaitForPlugins", 1)
caput("TESTSIMDETECTOR:CAM:AcquireTime", 0.005)
caput("TESTSIMDETECTOR:CAM:XSine1Amplitude", 10.0)
caput("TESTSIMDETECTOR:CAM:XSine2Amplitude", 0.0)
caput("TESTSIMDETECTOR:CAM:YSine1Amplitude", 5.0)
caput("TESTSIMDETECTOR:CAM:YSine2Amplitude", 0.0)
caput("TESTSIMDETECTOR:CAM:XSine1Frequency", 200.0)
caput("TESTSIMDETECTOR:CAM:XSine2Frequency", 50.0)
caput("TESTSIMDETECTOR:CAM:YSine1Frequency", 100.0)
caput("TESTSIMDETECTOR:CAM:YSine2Frequency", 10.0)
caput("TESTSIMDETECTOR:CAM:ImageMode", "Single")
caput("TESTSIMDETECTOR:SCAN.NPTS", 1000)
caput("TESTSIMDETECTOR:STAT:EnableCallbacks", "Enable")
caput("TESTSIMDETECTOR:STAT:TS:TSNumPoints", 1000)
caput("TESTSIMDETECTOR:SCAN:STRD01", "TESTSIMDETECTOR:STAT:MaxX_RBV")
caput("TESTSIMDETECTOR:SCAN:STRT1", "TESTSIMDETECTOR:CAM:Acquire")

TS_ACQUIRE = "TESTSIMDETECTOR:STAT:TS:TSAcquire"
CAPTURE = "TESTSIMDETECTOR:SCAN.EXSC"
CAPTURING = "TESTSIMDETECTOR:CAM:Acquire"
SSCAN_STATE = "TESTSIMDETECTOR:SCAN.SMSG"

duplicates = 0
counter = 1
while True:

    caput(CAPTURE, 1)
    while caget(SSCAN_STATE) == "SCAN Complete":
        Sleep(0.01)
    while caget(SSCAN_STATE) != "SCAN Complete":
        Sleep(0.01)

    array = np.array(caget("TESTSIMDETECTOR:SCAN.D01CA"))
    # print(array)

    for idx in range(len(array) - 1):
        if array[idx] == array[idx + 1]:
            print("DUPLICATE VALUE")
            print(array[idx-3:idx+3])
            duplicates += 1
    print("Total duplicates: {}".format(duplicates))

    print(counter)
    counter += 1
```


This is the boot script, which shows the plugins and settings:

```
# This file was automatically generated on Tue 12 Feb 2019 09:22:01 GMT from
# source: /dls_sw/work/R3.14.12.7/support/ADSimDetector/etc/makeIocs/simDetector.xml
# 
# *** Please do not edit this file: edit the source file instead. ***
# 
cd "/dls_sw/work/R3.14.12.7/support/ADSimDetector/iocs/simDetector"

epicsEnvSet "EPICS_CA_MAX_ARRAY_BYTES", '2000000'
epicsEnvSet "EPICS_TS_MIN_WEST", '0'


# Loading libraries
# -----------------

# Device initialisation
# ---------------------

cd "/dls_sw/work/R3.14.12.7/support/ADSimDetector/iocs/simDetector"

dbLoadDatabase "dbd/simDetector.dbd"
simDetector_registerRecordDeviceDriver(pdbbase)

# simDetectorConfig(portName, maxSizeX, maxSizeY, dataType, maxBuffers, maxMemory)
simDetectorConfig("CAM1.CAM", 1024, 1024, 1, 50, 0)

# NDStdArraysConfigure(portName, queueSize, blockingCallbacks, NDArrayPort, NDArrayAddr, maxMemory, priority, stackSize, maxThreads)
NDStdArraysConfigure("CAM1.ARR", 2, 0, "CAM1.CAM", 0, 0, 0, 0, 1)

# NDFileTIFFConfigure(portName, queueSize, blockingCallbacks, NDArrayPort, NDArrayAddr, maxBuffers, maxMemory)
NDFileTIFFConfigure("CAM1.TIFF", 2, 0, "CAM1.CAM", 0, 0, 0)

# NDFileJPEGConfigure(portName, queueSize, blockingCallbacks, NDArrayPort, NDArrayAddr, maxBuffers, maxMemory)
NDFileJPEGConfigure("CAM1.JPEG", 2, 0, "CAM1.CAM", 0, 0, 0)

# NDFileHDF5Configure(portName, queueSize, blockingCallbacks, NDArrayPort, NDArrayAddr, maxBuffers, maxMemory)
NDFileHDF5Configure("CAM1.HDF", 2, 0, "CAM1.CAM", 0, 0, 0)

# NDOverlayConfigure(portName, queueSize, blockingCallbacks, NDArrayPort, NDArrayAddr, NOverlays, maxBuffers, maxMemory, priority, stackSize, maxThreads)
NDOverlayConfigure("CAM1.OVR", 2, 0, "CAM1.CAM", 0, 8, 0, 0, 0, 0, 1)

# NDROIConfigure(portName, queueSize, blockingCallbacks, NDArrayPort, NDArrayAddr, maxBuffers, maxMemory, priority, stackSize, maxThreads)
NDROIConfigure("CAM1.ROI", 2, 0, "CAM1.CAM", 0, 0, 0, 0, 0, 1)

# NDTransformConfigure(portName, queueSize, blockingCallbacks, NDArrayPort, NDArrayAddr, maxBuffers, maxMemory, priority, stackSize, maxThreads)
NDTransformConfigure("CAM1.TRS", 2, 0, "CAM1.CAM", 0, 0, 0, 0, 0, 1)

# ADCore path for manual NDTimeSeries.template to find base plugin template
epicsEnvSet "EPICS_DB_INCLUDE_PATH", "/dls_sw/work/R3.14.12.7/support/ADCore/db"

# NDStatsConfigure(portName, queueSize, blockingCallbacks, NDArrayPort, NDArrayAddr, maxBuffers, maxMemory, priority, stackSize, maxThreads)
NDStatsConfigure("CAM1.STAT", 2, 0, "CAM1.CAM", 0, 0, 0, 0, 0, 1)
# NDTimeSeriesConfigure(portName, queueSize, blockingCallbacks, NDArrayPort, NDArrayAddr, maxSignals)
NDTimeSeriesConfigure("CAM1.STAT_TS", 2, 0, "CAM1.STAT", 1, 23)
# Load time series records
dbLoadRecords("/dls_sw/work/R3.14.12.7/support/ADCore/db/NDTimeSeries.template",  "P=TESTSIMDETECTOR,R=:STAT:TS:, PORT=CAM1.STAT_TS ,ADDR=0,TIMEOUT=1,NDARRAY_PORT=CAM1.STAT,NDARRAY_ADDR=1,NCHANS=2048,ENABLED=1")

# Final ioc initialisation
# ------------------------
cd "/dls_sw/work/R3.14.12.7/support/ADSimDetector/iocs/simDetector"
dbLoadRecords 'db/simDetector_expanded.db'
iocInit

dbpf TESTSIMDETECTOR:CAM:NDAttributesFile, simDetectorApp/data/CAM1.CAM.xml
dbpf TESTSIMDETECTOR:STAT:NDAttributesFile, simDetectorApp/data/CAM1.STAT.xml
```

I think it must be unlikely to happen, as I ran it all last night and it was still going this morning. I then started the script again (but not the IOC) and it was stuck when I came back ~ 4 hours later. I can run it again tonight and see what happens.

MarkRivers:
I believe this issue is a problem in devAsynInt32.c.  It was discussed in this tech-talk thread:
https://epics.anl.gov/tech-talk/2018/msg00786.php

That problem occurs with output records that have the info tag asyn:READBACK=1 and drivers that have ASYN_CANBLOCK=0.

In that thread we never came to a definite conclusion about a deadlock, but it was strongly suspected.

I believe there is indeed a deadlock in this case.  The problem is that if the driver has ASYN_CANBLOCK=0 then when the ADAcquire record processes the locks are taking in this order:

dbScanLock(Acquire record)
asynPortDriverLock(ADSimDetector)

However when the driver does callbacks to devAsynInt32.c to update the Acquire value (because asyn:READBACK tag is 1) then the locks are taken in this order:

asynPortDriverLock(ADSimDetector)
dbScanLock(Acquire record)

and we have a deadlock.

I see 3 potential solutions to this problem:
1. Change ADSimDetector to have ASYN_CANBLOCK=1.  This is probably how it should be.  If ASYN_CANBLOCK is 1 then the record processing does not have the dbScanLock held when it calls the driver because it is done in an asynManager callback from pasynManager->queueRequest().  We would change asyn to document that asyn:READBACK is only allowed for blocking drivers.

2. Change devAsynInt32 so it does not call dbScanLock in interruptCallbackOutput.  There are 2 reasons it is calling dbScanLock.
  - To protect pr->pact
  - To protect pPvt->numDeferredOutputCallbacks

3. Change ADBase.template so it uses another mechanism to set Acquire back to 0, rather than asyn:READBACK.

pPvt->numDeferredOutputCallbacks could be protected by pPvt->ringBufferLock instead.  That would require taking that lock in the process functions.

I'm not sure pr->pact actually needs to be protected.  We are just reading it, which is probably atomic since it is an int?

@GDYendell can you make the simple change of making ASYN_CANBLOCK=1 in ADSimDetector and see if the deadlocks go away?

MarkRivers:
I discussed this with @anjohnson.  He suggested using a variable in the device support private structure to hold the information encoded in PACT, i.e. whether the record is in the middle of asynchronous processing.  That way the device support callback function does not need to call dbScanLock and the deadlock is avoided.  I have done this in a new fix_output_callback branch on asyn
 
https://github.com/epics-modules/asyn/tree/fix_output_callback

I have been using a version of the script from @GDYendell to test with simDetector with the master branch in hopes of reproducing the deadlock.  I ran 1500 scans of 1000 images each (1.5M images) with no problems. I have also had no differences between the statistics data collected with the time-series code inside the plugin with the data collected with the EPICS PV in the sscan record.  That is a better test than looking for duplicate value in the sscan record, since those can happen by unlucky coincidence every once in a while.

@GDYendell if you can test the new asyn branch to see if the deadlocks disappear that would be great.

GDYendell:
Unfortunately(?) I have run the script all night (almost 7000 1000 point scans) and it has not had a deadlock. But, also no duplicates.

I have set up two extra IOCs plus two with the new asyn. They are running 500x500, 100 frames,  0.001s, so an acquisition per second. Do we have any estimates on the occurrence rate?

I have actually had some "duplicates", but they all appear in one acquisition where it gets stuck on the same value for about 50-100 points. I am fairly sure it is not relevant to this fix.

MarkRivers:
I also ran the script overnight, using the new asyn branch.  It has run 3300 scans (3.3M images) with no deadlock and no duplicates.  I think another test is needed to make the deadlock happen more often. That test would rapidly write to an output record that has asyn:READBACK without waiting for the readback (which the scan does).  Then it is likely to have writes and callbacks overlapping, which causes the deadlock.

I am now quite confident that the areaDetector problem is fixed and any remaining problems are in asyn.

GDYendell:
> I am now quite confident that the areaDetector problem is fixed and any remaining problems are in asyn.

I agree. The deadlock is a separate issue. There is no need to hold up this PR.

MarkRivers:
Closed via local merge

MarkRivers:
I believe I have fixed the deadlock as well.  See the asyn PR https://github.com/epics-modules/asyn/pull/81 and the asyn issue https://github.com/epics-modules/asyn/issues/80.

PR #378, opened 2019-02-11
Implemented the ability to force an HDF5 flush with new FlushNow record

MattTaylorDLS:
Up to date version of the original changes made by @ajgdls on https://github.com/dls-controls/ADCore/tree/manual-flush


MattTaylorDLS:
This PR adds a new feature to the HDF5 class to command an immediate flush of data to file, through a new busy record that has been added, called FlushNow.

This work was originally done by @ajgdls , and has since been tested at Diamond, and updates made as a result of this testing. This should be backwards compatible as it is only the addition of a new feature.

PR #381, opened 2019-02-25
Tests for windows

stuwilkins:


MarkRivers:
Now it is building OK on Linux but some of the tests are failing:
```
Entering test case "test_DatasetLayout1"
../test_NDFileHDF5.cpp(279): error in "test_DatasetLayout1": check odims[0] == 10 failed [0 != 10]
../test_NDFileHDF5.cpp(288): error in "test_DatasetLayout1": check odims[0] == 10 failed [0 != 10]
../test_NDFileHDF5.cpp(295): error in "test_DatasetLayout1": check odims[0] == 10 failed [0 != 10]
Leaving test case "test_DatasetLayout1"; testing time: 13363mks
Leaving test suite "NDFileHDF5Tests"
Entering test case "test_AttributeOriginalDataset"
../test_NDFileHDF5AttributeDataset.cpp(277): error in "test_AttributeOriginalDataset": check dims[0] == 32 failed [0 != 32]
```
Do you know why these are failing?

Are you building locally on Windows and getting link errors?  If so please include the error messages.

stuwilkins:
@MarkRivers I fixed the linux issues. 

1. I converted everything from a mix of boost and tr1 shared pointers to using `std::shared_ptr`
2. Added the option `USR_CXXFLAGS_Linux += -std=c++11` to the makefile to build against the C++11 standard
3. Fixed a bug introduced when I converted to dynamic arrays to using the `new` / `delete` construct.

I see travis builds, and I tested on my debian machine. I am now building on windows to see if the linker error still persists.

stuwilkins:
This is the msvc link error

```
link -nologo   -incremental:no -opt:ref -release  -MACHINE:X64    /SUBSYSTEM:CONSOLE /LIBPATH:C:\local\boost\installation\lib           -out:plugin-test.exe   plugin-test.obj test_NDPluginCircularBuff.obj test_NDFileHDF5.obj test_NDFileHDF5AttributeDataset.obj test_NDFileHDF5ExtraDimensions.obj test_NDPosPlugin.obj test_NDPluginTimeSeries.obj test_NDPluginFFT.obj test_NDPluginAttrPlot.obj test_NDPluginROI.obj test_NDPluginOverlay.obj test_NDArrayPool.obj     ../../../lib/windows-x64-static/ADTestUtility.lib  ../../../lib/windows-x64-static/NDPlugin.lib  ../../../lib/windows-x64-static/ADBase.lib  Z:/EPICS/epics-modules-win//epics-base/lib/windows-x64-static/qsrv.lib  ../../../lib/windows-x64-static/ntndArrayConverter.lib  Z:/EPICS/epics-modules-win//epics-base/lib/windows-x64-static/nt.lib  Z:/EPICS/epics-modules-win//epics-base/lib/windows-x64-static/pvDatabase.lib  Z:/EPICS/epics-modules-win//epics-base/lib/windows-x64-static/pvAccessIOC.lib  Z:/EPICS/epics-modules-win//epics-base/lib/windows-x64-static/pvAccessCA.lib  Z:/EPICS/epics-modules-win//epics-base/lib/windows-x64-static/pvAccess.lib  Z:/EPICS/epics-modules-win//epics-base/lib/windows-x64-static/pvData.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/netCDF.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/Magick++.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/coders.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/Magick.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/jbig.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/jp2.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/bzlib.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/png.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/webp.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/lcms.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/ttf.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/wmf.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/filters.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/tiff.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/jpeg.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/xml2.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/NeXus.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/hdf5.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/hdf5_hl.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/bitshuffle.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/blosc.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/szip.lib  Z:/EPICS/epics-modules-win//areaDetector/ADSupport/lib/windows-x64-static/zlib.lib  Z:/EPICS/epics-modules-win//asyn/lib/windows-x64-static/asyn.lib  Z:/EPICS/epics-modules-win//epics-base/lib/windows-x64-static/dbRecStd.lib  Z:/EPICS/epics-modules-win//epics-base/lib/windows-x64-static/dbCore.lib  Z:/EPICS/epics-modules-win//epics-base/lib/windows-x64-static/ca.lib  Z:/EPICS/epics-modules-win//epics-base/lib/windows-x64-static/Com.lib gdi32.lib oleaut32.lib psapi.lib ws2_32.lib advapi32.lib user32.lib kernel32.lib winmm.lib     
   Creating library plugin-test.lib and object plugin-test.exp
test_NDFileHDF5ExtraDimensions.obj : error LNK2019: unresolved external symbol "public: enum asynStatus __cdecl NDFileHDF5::configureDims(class NDArray *)" (?configureDims@NDFileHDF5@@QEAA?AW4asynStatus@@PEAVNDArray@@@Z) referenced in function "public: void __cdecl test_PluginExtraDimensions::test_method(void)" (?test_method@test_PluginExtraDimensions@@QEAAXXZ)
test_NDFileHDF5ExtraDimensions.obj : error LNK2019: unresolved external symbol "public: void __cdecl NDFileHDF5::calcNumFrames(void)" (?calcNumFrames@NDFileHDF5@@QEAAXXZ) referenced in function "public: void __cdecl test_PluginExtraDimensions::test_method(void)" (?test_method@test_PluginExtraDimensions@@QEAAXXZ)
plugin-test.exe : fatal error LNK1120: 2 unresolved externals
make[1]: *** [Z:/EPICS/epics-modules-win//epics-base/configure/RULES_BUILD:207: plugin-test.exe] Error 1120
make[1]: Leaving directory 'Z:/EPICS/epics-modules-win/areaDetector/ADCore/ADApp/pluginTests/O.windows-x64-static'
make: *** [Z:/EPICS/epics-modules-win//epics-base/configure/RULES_ARCHS:58: install.windows-x64-static] Error 2
```

MarkRivers:
```
test_NDFileHDF5ExtraDimensions.obj : error LNK2019: unresolved external symbol "public: enum asynStatus __cdecl NDFileHDF5::configureDims(class NDArray *)" (?configureDims@NDFileHDF5@@QEAA?AW4asynStatus@@PEAVNDArray@@@Z) referenced in function "public: void __cdecl test_PluginExtraDimensions::test_method(void)" (?test_method@test_PluginExtraDimensions@@QEAAXXZ)
test_NDFileHDF5ExtraDimensions.obj : error LNK2019: unresolved external symbol "public: void __cdecl NDFileHDF5::calcNumFrames(void)" (?calcNumFrames@NDFileHDF5@@QEAAXXZ) referenced in function "public: void __cdecl test_PluginExtraDimensions::test_method(void)" (?test_method@test_PluginExtraDimensions@@QEAAXXZ)
```

I see the problem.  To me the surprise is that it actually works on Linux, not that it fails on Windows.

This is from NDFileHDF5.h
```

class epicsShareClass NDFileHDF5 : public NDPluginFile
{
  public:
    static const char *str_NDFileHDF5_extraDimSize[MAXEXTRADIMS];
    static const char *str_NDFileHDF5_extraDimName[MAXEXTRADIMS];
    static const char *str_NDFileHDF5_extraDimChunk[MAXEXTRADIMS];
...
#ifndef _UNITTEST_HDF5_
  protected:
#endif
    /* plugin parameters */
    int NDFileHDF5_nRowChunks;
    #define FIRST_NDFILE_HDF5_PARAM NDFileHDF5_nRowChunks
    int NDFileHDF5_nColChunks;
...
#ifndef _UNITTEST_HDF5_
  private:
#endif
    /* private helper functions */
    inline bool IsPrime(int number)
    {
      if (((!(number & 1)) && number != 2) || (number < 2) || (number % 3 == 0 && number != 3)){
        return false;
      }

      for(int k = 1; 36*k*k-12*k < number;++k){
        if ((number % (6*k+1) == 0) || (number % (6*k-1) == 0)){
          return false;
        }
      }
      return true;
    };

    hid_t typeNd2Hdf(NDDataType_t datatype);
    asynStatus configureDatasetDims(NDArray *pArray);
    asynStatus configureDims(NDArray *pArray);
    asynStatus configureDatasetCompression();
...
    asynStatus writePerformanceDataset();
    void calcNumFrames();
...
    char *dimsreport;       /** < A string which contain a verbose report of all dimension sizes. The method getDimsReport fill in this */
};
```

Note that if \_UNITTEST_HDF5\_ is not defined then configureDims and calcNumFrames are declared private.  If \_UNITTEST_HDF5_\ is defined then they are public.  \_UNITTEST_HDF5\_ is not defined when building the plugin library.  However it is defined when building the NDPluginTests.
```
corvette:ADCore/ADApp/pluginTests>grep _UNITTEST_HDF5_ *.cpp
test_NDFileHDF5.cpp:#define _UNITTEST_HDF5_
test_NDFileHDF5ExtraDimensions.cpp:#define _UNITTEST_HDF5_
```

This means that the library was built declaring these methods private, but the unit tests are built declaring them pubic.  Windows is smart enough to not export those symbols in the DLL if they are declared private when compiling the class.  Apparently Linux still exports those symbols as class members even though they were declared private?

MarkRivers:
The conditional declarations of `protected` and `private` were done in commit 1af94bfb6fe049ef7362614ecd031f5f0109cd4c by @ajgdls.  Is declaring a method private when building a library but public when calling the library something that is supposed to work on Linux?

ajgdls:
Hi @MarkRivers I'm not sure I understand the question.  The unit tests originally compiled the classes into the executable with those methods as public.  This is a (not very elegant) solution to allow unit testing of private methods.
And so the library should have the methods declared as private, and no code should be trying to call them.  The unit tests should be building the classes directly into the executable with the _UNITTEST_HDF5_ set to allow the unit tests to call the methods.  This is what was happening on Linux.

ajgdls:
I should add that I don't think we should perform (b)!  and I'm not really happy with (a) either.

Is the windows build attempting to link the unit tests to the built library?

MarkRivers:
> Hi @MarkRivers I'm not sure I understand the question. The unit tests originally compiled the classes into the executable with those methods as public. This is a (not very elegant) solution to allow unit testing of private methods.

> The unit tests should be building the classes directly into the executable with the _UNITTEST_HDF5_ set to allow the unit tests to call the methods. This is what was happening on Linux.

Hi @ajgdls I think I am not understanding something.  Let's consider the NDFileHDF5::calcNumFrames method.  If `_UNITTEST_HDF_` is not defined then that method will be private.  The NDFileHDF5.cpp source code is only compiled once, when building NDPlugin library in pluginSrc.  When it is compiled `_UNITTEST_HDF_`  will not be defined, so the method will be private.

When the pluginTests directory is built it compiles test_NDFileHDF5ExtraDimensions.cpp which does define `_UNITTEST_HDF_`, so it will expect NDFileHDF5::calcNumFrames to be public.  But it does not actually recompile NDFileHDF5.cpp to make it public.  When the test application is linked it uses ADApp/commonDriverMakefile to link in libNDPlugin.so, which was built with NDFileHDF5::calcNumFrames  declared private.  

So how does it work?  I don't understand what you mean by 
>The unit tests should be building the classes directly into the executable with the _UNITTEST_HDF5_ set to allow the unit tests to call the methods.

> It buillds the classes into the executable by linking to a library where those methods were declared private.

What am I missing?

I can show that even though libNDPlugin.so was built with NDFileHDF5::calcNumFrames declared private it is still being exported in the symbol table by that library. Why does gcc do this?

```
corvette:ADCore/lib/linux-x86_64>nm libNDPlugin.so | grep calcNumFrames
00000000000978e0 T _ZN10NDFileHDF513calcNumFramesEv
```

Running demangle on that symbol yields this:
```
NDFileHDF5::calcNumFrames()
```
I think the problem with Windows is that because NDFileHDF5::calcNumFrames was declared private it is not in the symbol table, and hence the linker error.

ajgdls:
OK I think I understand the confusion but I don't understand why it works on Linux.

I believe that at some point in the past the classes (including HDF5PluginWrapper and NDFileHDF5) were expected to be compiled directly into the product plugin-test.  Then when used within the unit test product all methods would be seen as public.

The plugin library should have the same methods compiled only as private.

I've tried to go back to find the original Makefile but I can't see the commit, and so either it got changed and squashed locally or was never actually written as expected, and it wasn't noticed due to this current odd behaviour.

The fact that it works currently I do not understand at all!  I do not think the symbols should be visible in the library.

I think this problem should be rectified by first removing the ugly define and checks for the define within the code.  Then the specific unit tests should be revisited to see if they can be re-worked or the relevant methods considered public instead of private.  @MarkRivers what do you think to this course of action?  I think chasing the reason for the current behaviour is probably not the right way to go, I would prefer to remove the define statements.

MarkRivers:
I have removed the `_UNITTEST_HDF5_` conditionals from NDFileHDF5.h.
```
corvette:ADCore/lib/linux-x86_64>git diff ../../ADApp/pluginSrc/
diff --git a/ADApp/pluginSrc/NDFileHDF5.h b/ADApp/pluginSrc/NDFileHDF5.h
index f6396fc..a0430fb 100644
--- a/ADApp/pluginSrc/NDFileHDF5.h
+++ b/ADApp/pluginSrc/NDFileHDF5.h
@@ -119,9 +119,7 @@ class epicsShareClass NDFileHDF5 : public NDPluginFile
     std::map<std::string, hdf5::Element *>     onCloseMap;  // Map of handles to elements with onClose ndattributes, indexed by fullname
     Codec_t                                    codec;       // Definition of codec used to compress the data.

-#ifndef _UNITTEST_HDF5_
   protected:
-#endif
     /* plugin parameters */
     int NDFileHDF5_nRowChunks;
     #define FIRST_NDFILE_HDF5_PARAM NDFileHDF5_nRowChunks
@@ -162,9 +160,7 @@ class epicsShareClass NDFileHDF5 : public NDPluginFile
     int NDFileHDF5_SWMRMode;
     int NDFileHDF5_SWMRRunning;

-#ifndef _UNITTEST_HDF5_
   private:
-#endif
     /* private helper functions */
     inline bool IsPrime(int number)
     {
```
I rebuilt, libNDPluginLib.so built OK.
```
corvette:ADCore/lib/linux-x86_64>ls -l
total 31856
-r--r--r-- 1 epics domain users  2321836 Feb 26 08:20 libADBase.a
-r-xr-xr-x 1 epics domain users   959368 Feb 26 08:20 libADBase.so
-r--r--r-- 1 epics domain users  4142088 Feb 15 11:12 libADTestUtility.a
-r-xr-xr-x 1 epics domain users  1837600 Feb 15 11:12 libADTestUtility.so
-r--r--r-- 1 epics domain users 14218260 Feb 26 08:20 libNDPlugin.a
-r-xr-xr-x 1 epics domain users  5849032 Feb 26 08:20 libNDPlugin.so
-r--r--r-- 1 epics domain users  2202276 Feb 26 08:20 libntndArrayConverter.a
-r-xr-xr-x 1 epics domain users  1076800 Feb 26 08:20 libntndArrayConverter.so
```
Note however, that it still exports NDFileHDF5::calcNumFrames :
```
corvette:ADCore/lib/linux-x86_64>nm libNDPlugin.so | grep Frames
00000000000978e0 T _ZN10NDFileHDF513calcNumFramesEv
```
pluginTests now fails to build, as expected because calcNumFrames is now private.
```
In file included from ../HDF5PluginWrapper.h:11:0,
                 from ../test_NDFileHDF5ExtraDimensions.cpp:32:
../../../include/NDFileHDF5.h: In member function 'void test_PluginExtraDimensions::test_method()':
../../../include/NDFileHDF5.h:192:10: error: 'void NDFileHDF5::calcNumFrames()' is private
     void calcNumFrames();
          ^
../test_NDFileHDF5ExtraDimensions.cpp:419:23: error: within this context
   hdf5->calcNumFrames();
                       ^
In file included from ../HDF5PluginWrapper.h:11:0,
                 from ../test_NDFileHDF5ExtraDimensions.cpp:32:
../../../include/NDFileHDF5.h:192:10: error: 'void NDFileHDF5::calcNumFrames()' is private
     void calcNumFrames();
          ^
```
I conclude that gcc has a strange behavior.  Even if a method is declared private it is exported in the symbol table, and if an application modifies the header file to declare it public then it can call the private methods.  This seems like a dangerous feature, and does not occur with Visual Studio.

ajgdls:
I'll take a look at the unit tests and see what can be done with them.

stuwilkins:
Hi @MarkRivers @ajgdls 

Thanks! Success! The tests all run on windows now!

```
NDArrayPool:
  numBuffers=1, numFree=1
  memorySize=60000, maxMemory=60000
  freeList: (index, dataSize, pArray)
    0 60000 00000246E2EE1780

NDArrayPool:
  numBuffers=1, numFree=1
  memorySize=30000, maxMemory=60000
  freeList: (index, dataSize, pArray)
    0 30000 00000246E2EE1780
2019/02/26 09:58:04.058 NDArrayPool::alloc:: error: reached limit of 60000 memory (1 buffers)

NDArrayPool:
  numBuffers=0, numFree=0
  memorySize=0, maxMemory=60000
  freeList: (index, dataSize, pArray)

*** No errors detected

Z:\EPICS\epics-modules-win\areaDetector\ADCore\ADApp>
```
What I did was to place `USR_CXXFLAGS_WIN32 += -D_UNITTEST_HDF5_` in `areaDetector/ADCore/ADApp/pluginSrc/Makefile`

MarkRivers:
> What I did was to place `USR_CXXFLAGS_WIN32 += -D_UNITTEST_HDF5_` in `areaDetector/ADCore/ADApp/pluginSrc/Makefile`

OK, glad to hear that worked, but it is definitely not the best solution since it forces things to be public that are intended to be private.  Hopefully @ajgdls will find a fix for the unit tests.

stuwilkins:
@MarkRivers Yes I completely agree this should not be the solution! 

Given the changes that I made here, I wonder if we should merge now, or @ajgdls can work off this branch. I worry that we could have a number of collisions if not.

stuwilkins:
@ajgdls Thanks for the previous work! Would you mind having a look at these changes, they are mostly so this will compile with MSVC. I have ran it through that and it works OK. If its OK could someone merge it. Thanks!

MarkRivers:
By changing from boost and tr1 to using std::shared_ptr you have forced a dependence on C++11.

What gcc and Visual Studio compilers support C++11?  Are we disallowing older compilers, even if only for the unit tests?

MarkRivers:
I just checked the gcc 4.4.7 compiler that comes with Centos/RHEL 6.  It does not support the -std=c++11 flag, only -std=c++0x and gnu++0x.  I don't know if those support std::shared_ptr or not.

gcc 4.4.7 is a requirement for areaDetector since that is what Dectris ships on its DCU computers.

stuwilkins:
Thanks, sorry I was not aware of the dependence from Dectris. I suppose there are 2 issues here. 

1) Do people want to always compile unit tests for all systems or only on CI/CD ?
2) We also have a requirement on BOOST, which is AFAK only for the unit tests at the moment, its not used for the rest of AD. That also puts the unit tests in a slightly different position. 

I had a quick look and using `—std=c++0x` may work.

The other option is to use boost to do the stared pointers:

https://www.boost.org/doc/libs/1_61_0/libs/smart_ptr/shared_ptr.htm <https://www.boost.org/doc/libs/1_61_0/libs/smart_ptr/shared_ptr.htm>

Given boost is a dependence anyway this is perhaps the pagmatic way forward? From what I read, boost should handle the differences between windows and linux and the different compilers. 

Does someone with more C++ experience have comments or (quite possibly!) a better idea?

MarkRivers:
> The other option is to use boost to do the stared pointers:

That's actually what is currently done on the master branch.  The only exception is test_NDFileHDF5AttributeDataset.cpp which is using  std::tr1::shared_ptr instead.  I am not sure why?

Since boost is a requirement anyway for the boost_unit_test_framework why not just continue to use boost?  It should work on a larger variety of systems than C++11?

stuwilkins:
I just tested and all seems to be ok on linux having changed all to boost. I will see if this works on windows ….

MarkRivers:
Are you ready for me to merge this?  It looks OK to me.

stuwilkins:
I want to test build on windows… doing that now. If it builds then yes I would be ok. Thanks!

stuwilkins:
@MarkRivers I just did a test with MSVC 2017 and it works fine. This is ready for merge now! Thanks!

PR #382, opened 2019-02-27
Improved unit tests to remove the need for define statements.

ajgdls:
Removed check for _UNITTEST_HDF5_.
Moved some private methods required for testing into protected scope.
Added testing methods in wrapper class.
Provided getters for internal dimensions.
Updated unit tests.

This PR is raised to address issues found within #381.  The defines have been removed, private methods that are required have been made protected and exposed through the test wrapper class, and getters have been created for all internal dimensions that need to be checked.


stuwilkins:
Given @MarkRivers says OK, I am going to merge as it looks good to me too. Thanks!

PR #384, opened 2019-03-06
re-order blosc shuffle enum

xiaoqiangwang:
It should be according to this definition in *blosc.h*
```
#define BLOSC_NOSHUFFLE   0  /* no shuffle */
#define BLOSC_SHUFFLE     1  /* byte-wise shuffle */
#define BLOSC_BITSHUFFLE  2  /* bit-wise shuffle */
```


ulrikpedersen:
Whoops - yes this looks like it could be a bug actually. Thanks @xiaoqiangwang 

The HDF5 file writer assumes the shuffle enum numerical values to match the blosc.h. The value is just passed directly to the HDF5 filter parameters [here](https://github.com/areaDetector/ADCore/blob/79909f5fdf9c75d154c74262c9d2a35f8a50339e/ADApp/pluginSrc/NDFileHDF5.cpp#L3280)

I think this means that when you use the NDCodec plugin and select "Byte" you actually get "Bit" shuffle in the file - and vice versa...

@GDYendell could you please look into this and check if this is indeed the correct fix? (after shutdown of course)

Issue #385, opened 2019-03-06
memory leak in hdf5 writer (?)


xiaoqiangwang:
The writer is in single mode, without filters, no lazy open and no SWMR. I see the memory usage ramps up constantly in writing hdf5 files, in the step of ~10-60 KB.

I am not sure whether this leak comes from libhdf5 or NDFileHDF5.

MarkRivers:
Do you see a similar leak if you test with the NDFileNetCDF or TIFF drivers?  Can you send a screen shot of your detector and HDF5 screens so we can try to reproduce the problem?

xiaoqiangwang:
The memory usage stays constant for NDFileNetCDF and NDFileTIFF.

Here is the NDFileHDF5 panel when it runs on Windows.
![ndfilehdf5](https://user-images.githubusercontent.com/3283924/53896208-d86c5980-4033-11e9-98ae-fd7266abfe7f.png)

MarkRivers:
Is this issue still occurring?

MarkRivers:
I just answered my own question.  Using mode=Single and Autosave=Yes then I see a significant memory consumption when saving HDF5 files, but not when saving TIFF or netCDF.

MarkRivers:
I just ran the simDetector under valgrind looking for memory leaks when saving HDF5 files in Single mode.  When it exits I see the following, which looks like it could be the problem.
```
==109347== 6,048 bytes in 378 blocks are definitely lost in loss record 12,126 of 12,702
==109347==    at 0x4C2B9B5: calloc (vg_replace_malloc.c:711)
==109347==    by 0x653C0E: NDFileHDF5AttributeDataset::configureDims(int) (NDFileHDF5AttributeDataset.cpp:231)
==109347==    by 0x6540E9: NDFileHDF5AttributeDataset::createDataset(int) (NDFileHDF5AttributeDataset.cpp:73)
==109347==    by 0x64C35D: NDFileHDF5::createAttributeDataset(NDArray*) (NDFileHDF5.cpp:2735)
==109347==    by 0x64E248: NDFileHDF5::openFile(char const*, int, NDArray*) (NDFileHDF5.cpp:282)
==109347==    by 0x66BB1C: NDPluginFile::openFileBase(int, NDArray*) (NDPluginFile.cpp:73)
==109347==    by 0x66C6D3: NDPluginFile::writeFileBase() (NDPluginFile.cpp:254)
[...]

PR #387, opened 2019-03-20
Fix memory leaks

MarkRivers:
This fixes the HDF5 leaks that I have found using valgrind.

1) The NDFileHDF5Dataset objects were being allocated with "new" into this->detDataMap, but were never being deleted.
2) epicsStrDup was being used which allocates memory that is never freed.  I don't think it was necessary to use epicsStrDup.
3) The NDFileHDF5AttrDataset destructor was not freeing memory that it explicitly allocated.
4) There was no explicit destructor for NDFileHD5Dataset.  It is needed because it needs to free memory it explicitly allocated.


MarkRivers:
These problems were found when looking at #385

MarkRivers:
> Did you run valgrind again after the fixes? Could you share the command invocation? I'd like us to start doing that on a regular basis as well, possibly in CI.

Yes, I did and put the output in attachment in #385.

PR #391, opened 2019-03-28
Travis: run unittest through valgrind

ulrikpedersen:
Running unittest on Travis CI through valgrind for automatic memory checks.

This is a bit of a crude attempt to get memory/leak checks into the CI - consider this PR a discussion point for now.

Initial attempt does:
* Runs the unittest through valgrind, outputting report to a local text file.
* Prints (with tail) the last few lines of the valgrind report (i.e. the summary)

This obviously isn't the same as running an IOC - but is it good enough to get an idea if a new change (i.e. a PR) leaks?

Is the valgrind report useful in its current state? Should we do more to parse the resulting report? Add any other options to the valgrind command to show other details?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,10 @@ before_install:
 install:
 - make
 script:
-- "./bin/linux-x86_64/plugin-test --log_level=test_suite"
+- "valgrind --leak-check=yes --log-file=valgrind_output.txt ./bin/linux-x86_64/plugin-test --log_level=test_suite"
 after_success:
 - sh ci/coverage.sh
 - coveralls-lcov --source-encoding=ISO-8859-1 ci/coverage.info
+- tail --lines 15 valgrind_output.txt
 notifications:
   slack: epicsareadetector:xf0ac8PfwqcSqi2gE0tUcUkP
--- a/ci/install-packages.sh
+++ b/ci/install-packages.sh
@@ -101,6 +101,7 @@ if [[ $XML2_EXTERNAL == "YES" ]]; then
   sudo apt-get install libxml2-dev
 fi
 
+sudo apt-get install valgrind
 # TO DO: Install ZLIB, SZIP, JPEG, NEXUS, NETCDF if we want to use package versions
 
 # Installing latest version of code coverage tool lcov (because the ubuntu package is very old)


ulrikpedersen:
This came out of a discussion on #385 with @hinxx

hinxx:
I will have couple of valgrind outputs to compare, soon.

I have no idea if there are any valgrind parsers out there to help us out, maybe something to look into if the simple method here proves inefficient.

I'm thinking of starting with making today's attempt at #385 fix a baseline. Compare that with #385 pre-fix state and possibly with #392 - if deemed plausible. Not sure what to expect between two runs of valgrind, event against the same compiled code.

hinxx:
I am down to this:

```
==25327== 
==25327== LEAK SUMMARY:
==25327==    definitely lost: 1,376 bytes in 39 blocks
==25327==    indirectly lost: 0 bytes in 0 blocks
==25327==      possibly lost: 1,227,774 bytes in 9,124 blocks
==25327==    still reachable: 37,398,861 bytes in 216,031 blocks
==25327==         suppressed: 0 bytes in 0 blocks
==25327== Reachable blocks (those to which a pointer was found) are not shown.
==25327== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==25327== 
==25327== For counts of detected and suppressed errors, rerun with: -v
==25327== ERROR SUMMARY: 157 errors from 157 contexts (suppressed: 0 from 0)

```

hinxx:
The `definitely lost` count is probably more or less what we have to live with - they come from EPICS base. There should be no ADCore ones left.
The `still reachable` high count is (probably?) due to not having destructors in plugins and similar. I guess this is OK as it gets destroyed at IOC exit.

hinxx:
Here is a collection of valgrind outputs. [valgrind_files_and_script.tar.gz](https://github.com/areaDetector/ADCore/files/3019652/valgrind_files_and_script.tar.gz). These were created during the day of leak plugging. Several of them exists numbered by sequence, for the same valgrind memory check (1-4 are from the same code, 5-8 are from the same code).

For example in the fix-#392/ folder of the attached archive you will find valgrind outputs from several stages. number 8 and 9 differ in a single loss record that is missing in output 9.

Included script can be used like so:
```
hinxx@obzen ~/bde/R3.15.5/simioc-master/tests $ python parser-valgrind-output.py fix-#392/valgrind_output.txt-fix-8
nr records 158
nr GOOD records 23
hinxx@obzen ~/bde/R3.15.5/simioc-master/tests $ python parser-valgrind-output.py fix-#392/valgrind_output.txt-fix-9
nr records 157
nr GOOD records 22
```

You can see that one loss record was solved in 9 vs. 8. The output of the script is also saved in files with appendix -all and -good (same folder as the input valgrind output file). The script will also remove the leading PID and loss record count/number to make comparison easier on the eyes.

Not really useful without human intervention since diff will not show us what we want if applied to two script outputs.

PR #395, opened 2019-04-02
Lz4 filter

MarkRivers:
This adds support for the lz4 filter (without bitshuffle or Blosc).  

I have tested both direct chunk write of pre-compressed NDArrays from NDPluginCodec, and compressing arrays in the HDF5 library (uncompressed arrays into NDFileHDF5).


MarkRivers:
@xiaoqiangwang can you also take a look at this?

MarkRivers:
Do

MarkRivers:
> These enums are inconsistent with NDPluginCodec. Is there a standard that should/could be followed?

Do you mean the enum strings or values?

GDYendell:
Well the values don't match up, but that might not be avoidable, as some exist in one and not the other. This is probably OK as long as it does the right thing underneath.

I meant the enum strings; it would be nice if they matched between the two so that is obvious it is configured correctly and so that higher level control and scripts can set the same string value on both PVs.

MarkRivers:
I agree that would be nice. I think we should change the ones in NDFileHDF5.template since those are newer? It will break backwards compatibility for Blosc?

GDYendell:
> I think we should change the ones in NDFileHDF5.template since those are newer?
I think so, yes. It would be the smallest change overall.

PR #406, opened 2019-05-09
Fix occasional segfault when opening TIFF files

andrej-m:
The first `for` loop in openFile() was writing one element index past
the tiffFieldInfo array bounds, which sometimes caused a segmentation
fault.


MarkRivers:
I have made this fix, but in a different way in commit ef7aa9e8e317970dd43d940be387d641427bb1d0
```
commit ef7aa9e8e317970dd43d940be387d641427bb1d0
Author: Mark Rivers <rivers@cars.uchicago.edu>
Date:   Thu May 9 11:14:54 2019 -0500

    Increased TIFFTAG_LAST_ATTRIBUTE from 65500 to 65535 since that is what the TIFF standard allows.
    Allow use of TIFFTAG_LAST_ATTRIBUTE tag.  Previous it only allowed up to TIFFTAG_LAST_ATTRIBUTE-1.
    Fixed error in definition of NUM_CUSTOM_TIFF_TAGS which could lead to access violation.
```
I believe this fixes the allocation error that caused the seg-fault, while also increasing the number of user-defined tags by 36.  

@andrej-m please test this if you can.

andrej-m:
I tested ef7aa9e on my Windows setup, the problem is solved. Array indexes on line 99 look OK.

Thanks Mark!

PR #413, opened 2019-06-26
Adding opencv_video and opencv_videoio as dependancy libs for ADCompVision

jwlodek:
I have been working on the R1-2 release of ADCompVision which adds support for saving videos from areaDetector cameras (timelapses, recordings, etc.) This requires opencv_video and opencv_videoio libraries to be linked during build, so these libs have been added to the ADCompVision call in commonDriverMakefile

### DIFF
--- a/ADApp/commonDriverMakefile
+++ b/ADApp/commonDriverMakefile
@@ -245,9 +245,9 @@ ifdef ADCOMPVISION
   PROD_LIBS     += NDPluginCV
   ifdef OPENCV_LIB
     opencv_core_DIR += $(OPENCV_LIB)
-    PROD_LIBS       += opencv_core opencv_imgproc opencv_highgui
+    PROD_LIBS       += opencv_core opencv_imgproc opencv_highgui opencv_video opencv_videoio
   else
-    PROD_SYS_LIBS   += opencv_core opencv_imgproc opencv_highgui
+    PROD_SYS_LIBS   += opencv_core opencv_imgproc opencv_highgui opencv_video opencv_videoio
   endif 
 endif
 


MarkRivers:
After merging this I discovered a problem.  I cannot find a package containing libopencv-videoio for Centos 7.   

The opencv-devel-2.4.5 package contains these libraries:
```
opencv-devel-2.4.5-3.el7.x86_64 : Development files for using the OpenCV library
Repo        : @base
Matched from:
Filename    : /usr/lib64/libopencv_contrib.so
Filename    : /usr/lib64/libopencv_highgui.so
Filename    : /usr/lib64/libopencv_calib3d.so
Filename    : /usr/lib64/libopencv_stitching.so
Filename    : /usr/lib64/libopencv_ts.so
Filename    : /usr/lib64/libopencv_imgproc.so
Filename    : /usr/lib64/libopencv_superres.so
Filename    : /usr/lib64/libopencv_flann.so
Filename    : /usr/lib64/libopencv_features2d.so
Filename    : /usr/lib64/libopencv_video.so
Filename    : /usr/lib64/libopencv_photo.so
Filename    : /usr/lib64/libopencv_objdetect.so
Filename    : /usr/lib64/libopencv_core.so
Filename    : /usr/lib64/libopencv_ml.so
Filename    : /usr/lib64/libopencv_legacy.so
Filename    : /usr/lib64/libopencv_videostab.so
```
Many sites are using RHEL 7 or Centos 7.  This means they will now get build errors if they set WITH_OPENCV=YES.

jwlodek:
Wouldn't they only be affected by having `ADCOMPVISION` defined in the release file, not just with `WITH_OPENCV=YES`? I will check tomorrow to make sure but it is possible that that dependency is no longer required as I changed the way the videowriter writes files, and if that is the case we can simply remove it. I have also almost finished the other changes to ADCompVision which I'll make a pull request for tomorrow.

MarkRivers:
> Wouldn't they only be affected by having ADCOMPVISION defined in the release file, not just with WITH_OPENCV=YES? 

Sorry, yes that's correct.  In fact WITH_OPENCV is not used any Makefiles, so it should be removed.

Issue #420, opened 2019-10-23
Debug assertion in NDFileHDFLayoutXML.cpp, when built with debug DLL configuration


pheest:
This problem appears to be caused by the lines
// Also a problem in Visual Studio.  The warning is:
// warning LNK4217: locally defined symbol xmlFree imported in function "private: int __cdecl hdf5::LayoutXML::parse_root(void)" 
#ifdef _MSC_VER
  #define xmlFree free
#endif

I think the assertion signifies that memory is allocated in one DLL (through xmlMalloc), but is freed in another (through free directly).

I commented out the preprocessor override, and the issue went away.
I believe it can be removed.

I can't say I've noticed the locally defined symbol warning.

MarkRivers:
What have you tested building?
- VS version?
- Static/dynamic?
- Debug/Release?

I was definitely getting those warnings. I believe they occurred with VS2015, Static Release build.

pheest:
Hi Mark,
I'm using VS2017.
I've built debug and release /x86 dynamic.
[areaDetector.win32-x86.log](https://github.com/areaDetector/ADCore/files/3771921/areaDetector.win32-x86.log)
[areaDetector.win32-x86-debug.log](https://github.com/areaDetector/ADCore/files/3771922/areaDetector.win32-x86-debug.log)

I don't see any "locally defined symbol" warnings here.

pheest:
OK, the linker warning is present in static builds.
The fix is to add something like:
ifneq ($(SHARED_LIBRARIES),YES)
  USR_CFLAGS_WIN32 += -DLIBXML_STATIC
  USR_CXXFLAGS_WIN32 += -DLIBXML_STATIC
endif
to the pluginSrc makefile.

pheest:
My pull request includes this.
But I left the useage for migw 'as was' since I don't have the capability for a build within that environment.
I would encorage you to verify whether adding -DLIBXML_STATIC correctly fixes the issue for that environment also.

BenBradnick:
The removal of:
```
#ifdef _MSC_VER
  #define xmlFree free
#endif
```
results in the following linker error when we try and link against the library in VS2015: `NDPlugin.lib(NDFileHDF5LayoutXML.obj) : error LNK2001: unresolved external symbol xmlFree`

MarkRivers:
@BenBradnick what architecture are you building for when you see that error message?

BenBradnick:
@MarkRivers this is for windows-x64.

[...]

PR #424, opened 2019-11-20
Added capability for CCD multi-track operation.

pheest:
This is typically used for spectroscopy applications.

Also, addressed issue #420, when used with VS build.


pheest:
Multi-track spectrocopy is a capability of several CCD camera types.
We have been using this at JET for many years with my exisitng (non-EPICS) software with the following types:
Andor, Princeton, Photometrics. Also some obsolete types, e.g. Wright Instruments.
There is a clear intent on my part to use this also for (at least) Princeton and Photometrics. For this reason I have placed it here.

I have not written a GUI, to date. I have tested the software by means of batch scripts. 
Clearly a GUI will be needeed, by users, when this goes into service.

PR #425, opened 2019-11-28
Added the creation of NDAttributes that describe how the multi-track …

pheest:
…was configured.

I prefer to add this as a method in ADCore/CCDMultiTrack (invoked from the camera class), since the attribute storage itself is not camera-type specific.

### DIFF
--- a/ADApp/ADSrc/CCDMultiTrack.cpp
+++ b/ADApp/ADSrc/CCDMultiTrack.cpp
@@ -42,7 +42,8 @@ void CCDMultiTrack::storeTrackAttributes(NDAttributeList* pAttributeList)
         for (size_t TrackNum = 0; TrackNum < size(); TrackNum++)
         {
             // Add new attributes listing.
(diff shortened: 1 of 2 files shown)


MarkRivers:
Note that your pull request fails to build on Travis because std::string::to_string does not exist in that system.

pheest:
Oh dear. Sorry about that. I’ve replaced the usage with _itoa.

MarkRivers:
_itoa does not fix the problem. Please see the Travis log.

pheest:
The Microsoft compiler tells me I should use _itoa - and not itoa - because it is posix and ISO C++ compliant…
Should have checked that.

OK, snprintf will work.

MarkRivers:
This merge had 2 problems.

- It uses snprintf.  That function is not supported on older versions of vxWorks, and so the Jenkins build at APS failed.  I replaced it with epicsSnprintf.
- It has Windows line endings.  All other files in areaDetector has Linux line endings.  I changed them to Linux line endings.

Issue #416, opened 2019-07-04
Throttling usage of callbacks for 10kHz detector


LeeHudsonDLS:
@MarkRivers We have a detector (panda) producing NDArrays at 10kHz which I am trying to optimise (cpu usage is through the roof). Profiling the IOC i can see the vast majority of the cpu usage is being taken up with callbacks. It would be good to have method that calls callParamCallbacks every n miliseconds rather than for every array but also keeps track of arrays in between so as not to lose them. I was thinking of doing this in ADCore even though this callParamCallbacks is from asyn, can you think of a better way to get around this?

MarkRivers:
Can you post the profile output? What are the callbacks doing?

I don’t understand what you mean by keeping track of arrays in between. What does it do with those intermediate arrays? 

How large are the arrays?

thomascobb:
> Can you post the profile output? What are the callbacks doing?

@LeeHudsonDLS will post the output shortly. The callbacks are updating the uniqueID and timeStamp parameters on all the plugins in the chains at 10kHz, causing the db to deliver camonitor updates at 10kHz to CA clients.

> I don’t understand what you mean by keeping track of arrays in between. What does it do with those intermediate arrays?

I think this can be better worded as "if we are only calling callParamCallbacks every n milliseconds, then a final NDArray comes in to update the uniqueID and timeStamp parameters without calling callParamCallbacks, how can we ensure the pvs will not show old information if nothing else calls callParamCallbacks?"

Here's an example, let's assume arrays come in at 5kHz but we only want to update every 1 ms:

```
UID  1 2 3 4 5 6 7                             8 9
CB?  Y N N N N Y N       Y                     Y N
```

With the current system, we would get a callParamCallbacks() at every UID, but that kills performance. If we throttle it to only actually call callParamCallbacks() if we haven't done so in the last millisecond we will get the updates at 1, 6 and 8, but not the one a short time after 7. This is obviously not ideal as we would spend a good amount of time with incorrect information on the screen. A possible solution is pushing the callParamCallbacks to a background thread:

```python
# Pseudo-code that looks an awful lot like Python...
class asynNDArrayDriver:
    min_time_between_cpc = 0.001  # seconds
    need_cpc = False

    def callParamCallbacks(self):
        self.need_cpc = True

    def thread(self):
        while True:
            self.lock():
            if self.need_cpc:
                super().callParamCallbacks()
                self.need_cpc = False
            self.unlock()
[...]

PR #426, opened 2020-01-02
Set scan rate

tomtrafford:
After finding a high CPU rate when running a detector at 10KHz, it was found that changing the SCAN value for certain PVs would lower the CPU rate considerably. We have then added a macro for these PVs, so they could be easily changed, with a default value of "I/O Intr" so these changes would have no effect on others.

Refs #416

### DIFF
--- a/ADApp/Db/ADBase.template
+++ b/ADApp/Db/ADBase.template
@@ -181,7 +181,7 @@ record(ai, "$(P)$(R)AcquireTime_RBV")
    field(DTYP, "asynFloat64")
    field(INP,  "@asyn($(PORT),$(ADDR=0),$(TIMEOUT=1))ACQ_TIME")
    field(PREC, "3")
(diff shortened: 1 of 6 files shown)


tomtrafford:
@thomascobb Here are the changes we were trying

tomtrafford:
Using these changes to set the SCAN values to .1 second, I observed a cpu usage for the IOC drop from about 300% to around 180%.

thomascobb:
I think those issues will be fine for us, but I guess we should add it to the documentation. Which file should we put it in?

MarkRivers:
> I think those issues will be fine for us, but I guess we should add it to the documentation. Which file should we put it in?

I think in the ADCore RELEASE.md.  For the ArrayCounter_RBV put a comment in that database at the top about the potential impact on ImageJ.

thomascobb:
@MarkRivers do the docs additions look ok too?

PR #428, opened 2020-02-28
Adding support for maxThreads > 1 for NDPluginProcess

jwlodek:
Beamline at NSLS2 uses plugin for background subtraction, but CPU was not able to process frames fast enough causing dropped frames. Added support for `maxThreads` > 1, tested with ADSimDetector, ADUVC, and ADDexela


jwlodek:
Is there a way we could allow multiple threads for background and flat field calculations, but not Filter calculations? Something like call `setIntegerParam(ADMaxThreads, 1)` When enableFilter is set, but set it back to whatever it was previously when this is disabled? I believe the beamline is only using the background subtraction feature here, and form looking at the code
```
for (i=0; i<nElements; i++) {
        value = data[i];
        if (autoOffsetScale) {
            if (data[i] < minValue) minValue = data[i];
            if (data[i] > maxValue) maxValue = data[i];
        }
        if (background) value -= background[i];
        if (flatField) {
            if (flatField[i] != 0.) 
                value *= scaleFlatField / flatField[i];
            else
                value = scaleFlatField;
        }
        if (enableOffsetScale) value = (value + offset)*scale;
        if (enableHighClip && (value > highClip)) value = highClip;
        if (enableLowClip  && (value < lowClip))  value = lowClip;
        data[i] = value;
    }
```
from what I can see, the section with the background subtraction doesn't write to any shared memory, though the filter does.

jwlodek:
I did something similar for `ADCompVision`, where for the image subtraction function, which similarly needed images to process in order I set the thread count to 1, but allowed it to be higher for other functions, since those are thread safe

MarkRivers:
I actually think I see a problem even with one thread in the background subtraction.  The hard work in the processPlugins code is done with the lock released.  However, it is accessing the memory in pBackground and pFlatField.  But if the user presses SaveBackground that releases pBackground and allocates a new one.  This seems like it could lead to memory access errors?

jwlodek:
I would say that that is the case, but I just tested this on the live IOC, and I can't get it to crash.

If I have background subtraction enabled, and click save background, It seems to just stop doing the background subtraction, and then if I press save again, it starts the subtraction again. The behavior is the same with 1 thread and with multiple threads

MarkRivers:
I think that if you click save background while background subtraction is enabled, the background it uses is the previous background subtracted image, which is small.  So it will look like there is no background subtraction because what it is subtracting is small.

I probably does not crash because the *pBackgound NDArray has just been returned to the NDArrayPool, the memory has not actually be released.  But that NDArray could be reused while the background subtraction is in process, and so the computed data would be wrong.

MarkRivers:
I am curious what frame size and frame rate is the detector that cannot keep up with background subtraction in a single thread?

jwlodek:
The frame is 3K by 3K and the rate is only 2Hz, but the machine it is running on is very weak (from 2010, and only 4 cores), and I believe it is shared with another detector. I have recommended that the long term solution is to simply upgrade the machine.

PR #434, opened 2020-04-13
Fix problem with not closing constant datasets

### LINKED ISSUE #433: HDF5 file plugin does not properly close constant datasets
If an HDF5 layout XML file specifies a dataset with source="constant" then that dataset is not closed when the file is closed. This leads to error messages from NDFileHDF5 and presumably to memory leaks.

Here is an simple example HDF5 layout XML that demonstrates the problem.
```
<hdf5_layout>
  <group name="detector">
    <dataset name="data1" source="detector" det_default="true" />
    <dataset name="det_name" source="constant" value="Simulated detector" type="string" when="OnFileClose" />
  </group>
</hdf5_layout>
```
Note that the dataset called "det_name" has source="constant".  When saving HDF5 files with the simDetector with this layout file the following messages are observed:

```
2020/04/13 15:27:43.178 NDFileHDF5::closeFile Closing file not totally clean.  Other remaining=2
```
If that line with the second dataset is removed or commented out the error messages do not occur.

I believe this bug was introduced in commit 9efe84144035e9082e9f899f0c4009eef16970f9.  Notice that on line 615 this call to close the dataset was commented out:
```
  //H5Dclose (hdfdset);
```

Presumably this was done in order to support defining attributes for constant datasets, since the dataset must not be closed to add attributes later.

I believe the fix is to maintain a map of constant datasets as is done for detector datasets and attribute datasets, and then call H5Dclose for each of them in NDFileHDF5::closeFile().

MarkRivers:
This fixes #433.

I discovered that attDataMap was actually not being used.  Rather this list was used instead:
```
 std::list<NDFileHDF5AttributeDataset*> attrList;
```
I renamed attDataMap to constDsetMap.  It contains all the hid_t for each of the constant datasets.  All of these datasets are now closed in CloseFile().

This eliminates the error messages.

### DIFF
--- a/ADApp/pluginSrc/NDFileHDF5.cpp
+++ b/ADApp/pluginSrc/NDFileHDF5.cpp
@@ -1067,6 +1067,10 @@ hid_t NDFileHDF5::createDataset(hid_t group, hdf5::Dataset *dset)
   }
   else if(dset->data_source().is_src_constant()) {
       retcode = this->writeHdfConstDataset(group,  dset);
+      if (retcode != -1) {
+          // Store the dataset into the constant dataset map
+          this->constDsetMap[dset->get_full_name()] = retcode;
+      }
   }
   else {
     retcode = -1;
@@ -1699,8 +1703,8 @@ asynStatus NDFileHDF5::closeFile()
     H5Dclose(it_dset->second->getHandle());
   }
   std::map<std::string, hid_t>::iterator it_hid;
-  // Iterate over the stored attribute data sets and close them
-  for (it_hid = this->attDataMap.begin(); it_hid != this->attDataMap.end(); ++it_hid){
+  // Iterate over the stored constant data sets and close them
+  for (it_hid = this->constDsetMap.begin(); it_hid != this->constDsetMap.end(); ++it_hid){
     H5Dclose(it_hid->second);
   }
 
@@ -1746,7 +1750,7 @@ asynStatus NDFileHDF5::closeFile()
     delete it_dset->second;
   }
   detDataMap.clear();
-  attDataMap.clear();
+  constDsetMap.clear();
   defDsetName = "";
 
   epicsTimeGetCurrent(&now);
--- a/ADApp/pluginSrc/NDFileHDF5.cpp
+++ b/ADApp/pluginSrc/NDFileHDF5.cpp
@@ -869,7 +869,6 @@ hid_t NDFileHDF5::writeH5dsetStr(hid_t element, const std::string &name, const s
     return -1;
   }
 
-  //H5Dclose (hdfdset);
   H5Tclose(hdfdatatype);
   H5Sclose(hdfdataspace);
 
@@ -956,7 +955,6 @@ hid_t NDFileHDF5::writeH5dsetInt32(hid_t element, const std::string &name, const
       return -1;
     }
   }
-  //H5Dclose (hdfdset);
   H5Sclose(hdfdataspace);
   H5Tclose(hdfdatatype);
   return hdfdset;
@@ -1043,7 +1041,6 @@ hid_t NDFileHDF5::writeH5dsetFloat64(hid_t element, const std::string &name, con
       return -1;
     }
   }
-  //H5Dclose (hdfdset);
   H5Sclose(hdfdataspace);
   H5Tclose(hdfdatatype);
   return hdfdset;
--- a/ADApp/pluginSrc/NDFileHDF5.h
+++ b/ADApp/pluginSrc/NDFileHDF5.h
@@ -122,7 +122,7 @@ class epicsShareClass NDFileHDF5 : public NDPluginFile
     hsize_t getVirtualDim(int index);
 
     std::map<std::string, NDFileHDF5Dataset *> detDataMap;  // Map of handles to detector datasets, indexed by name
-    std::map<std::string, hid_t>               attDataMap;  // Map of handles to attribute datasets, indexed by name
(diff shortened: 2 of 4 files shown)


ajgdls:
@GDYendell don't these have to be held open until CloseFile is called?  I don't see how it would work otherwise.

GDYendell:
The datasets just need to be closed sometime between being created and the file being closed, right? There is no other dependency?

ajgdls:
Is the constant value not written for every frame?  So if you have 0 frames (continuous) then you would write the constant each frame until the acquisition is stopped.

ajgdls:
I think this was added by another facility, it's not something we use at DLS.

GDYendell:
I think this is a value defined in the xml file layout, which will only be created once. @MarkRivers is that correct?

MarkRivers:
> Presumably, five years later, still nothing makes use of the fact that these datasets aren't closed straight away? I wonder if it would be better to just revert to the old behavior and uncomment the H5DClose calls. As you have demonstrated here, it is trivial to implement this properly when and if it is required.

No, I don't think that is true.  My understanding of the reason for commit 9efe84144035e9082e9f899f0c4009eef16970f9 was that there were errors adding HDF5 attributes to constant datasets.  Presumably this was because the attributes were being added after the dataset had been closed, so the lines to close the datasets were commented out.  However, the datasets were then never closed.  This problem was hidden until fairly recently, when we started testing that there were still open objects in CloseFile().

I know that people have been using constant datasets.  Here is an example from an APS tomography setup:
https://github.com/data-exchange/dxfile/blob/master/doc/demo/areadetector/2-BM/flir2bmaLayout.xml
```
    <group name="measurement">
      <group name="instrument">
        <dataset name="name" source="constant" value="2-BM-A Fast micro-tomography system using PG cameras" type="string" when="OnFileClose" />   
```

> Is the constant value not written for every frame? So if you have 0 frames (continuous) then you would write the constant each frame until the acquisition is stopped.

The above example specifies when="OnFileClose".  I am not sure if that is actually honored for constant datasets.  I will try testing with when="OnFileWrite" and see if it is actually written more than once.

> I wonder if it would be better to just revert to the old behavior and uncomment the H5DClose calls. 

I suspect that if that is done, and the constant dataset has an attribute then it will crash or give an error.  That seems to be the reason for the commit that removed the calls to H5Dclose().  I will test this.

MarkRivers:
> I will try testing with when="OnFileWrite" and see if it is actually written more than once.

I just tested the simDetector with this simple layout XML file
```
<hdf5_layout>
  <group name="detector">
    <dataset name="data1" source="detector" det_default="true">
        <attribute name="AcquireTime" source="ndattribute" ndattribute="AcquireTime" when="OnFileWrite"></attribute>
   </dataset>
    <dataset name="det_name" source="constant" value="Simulated detector" type="string" when="OnFileWrite">
        <attribute name="AcquireTime" source="ndattribute" ndattribute="AcquireTime" when="OnFileWrite"></attribute>
   </dataset>
   <dataset name="NumCaptured" source="ndattribute" ndattribute="NumCaptured" when="OnFileWrite"/>
  </group>
</hdf5_layout>
```
I collected a stream of 10 arrays with simDetector.

The constant dataset det_name is only written once, it is not written 10 times.

The dataset NumCaptured, on the other hand only has 1 value if when="OnFileOpen" or "OnFileClose", but has 10 values if it is "OnFileWrite".  

>> I wonder if it would be better to just revert to the old behavior and uncomment the H5DClose calls.

> I suspect that if that is done, and the constant dataset has an attribute then it will crash or give an error. That seems to be the reason for the commit that removed the calls to H5Dclose(). I will test this.

I tested that with the above XML file.  The AcquireTime attribute of the det_name constant dataset is written correctly with this pull request.  I also tested uncommenting the  H5Dclose() calls when writing constant datasets and commenting the H5Dclose() call in CloseFile() for constant datasets.  To my surprise this worked fine, there was no crash or HDF5 error messages.  

So I am not sure why those H5Dclose() calls were originally commented out.  The related issue #65 does not describe the problem.

GDYendell:
A lot (+1500 / -500 lines) has changed since that commit, so it is possible any issues were fixed.

I guess the more important question is, is this the behavior you expect from constant datasets? I.e. they are defined statically in the layout xml and written once either on open or close?

MarkRivers:
> I guess the more important question is, is this the behavior you expect from constant datasets? I.e. they are defined statically in the layout xml and written once either on open or close?

I think that behavior is OK.  I don't see any need for OnFileWrite to write the same thing more than once.  For the constant dataset itself there is no way to determine if it was written when the file was opened or closed.  However, there is a way to tell when an associated attribute was written.

I used this test layout file:
```
<hdf5_layout>
  <group name="detector">
    <dataset name="data1" source="detector" det_default="true">
        <attribute name="AcquireTime" source="ndattribute" ndattribute="AcquireTime" when="OnFileClose"></attribute>
        <attribute name="NumCaptured" source="ndattribute" ndattribute="NumCaptured" when="OnFileOpen"/>
   </dataset>
    <dataset name="det_name" source="constant" value="Simulated detector" type="string" when="OnFileOpen">
        <attribute name="NumCaptured" source="ndattribute" ndattribute="NumCaptured" when="OnFileClose"></attribute>
   </dataset>
   <dataset name="NumCaptured" source="ndattribute" ndattribute="NumCaptured"/>
  </group>
</hdf5_layout>
```

Note that the NumCaptured attribute of det_name can specify either on OnFileOpen or OnFileClose, and the values will be different for the 2 cases.  I determined empirically that the value written to the file depends on the value of "when" for the attribute, and is independent of the value of "when" for the constant dataset.

PR #436, opened 2020-04-15
re-order header inclusion to have correct exports

xiaoqiangwang:
the principle is to only include class header after
epicsExport.h.

parent class header is included once more before
epicsExport.h, so the linkage gets correct.

### DIFF
--- a/ADApp/pluginSrc/NDFileHDF5.cpp
+++ b/ADApp/pluginSrc/NDFileHDF5.cpp
@@ -34,6 +34,7 @@
 #include <osiSock.h>
 
 #include <asynDriver.h>
+#include "NDPluginFile.h"
 
 #include <epicsExport.h>
 #include "NDFileHDF5.h"
--- a/ADApp/pluginSrc/NDFileHDF5.h
+++ b/ADApp/pluginSrc/NDFileHDF5.h
@@ -9,9 +9,7 @@
 
 #include <list>
 #include <hdf5.h>
-#include <asynDriver.h>
-#include <NDPluginFile.h>
-#include <NDArray.h>
+#include "NDPluginFile.h"
 #include "NDFileHDF5Layout.h"
 #include "NDFileHDF5Dataset.h"
 #include "NDFileHDF5LayoutXML.h"
--- a/ADApp/pluginSrc/NDFileJPEG.cpp
+++ b/ADApp/pluginSrc/NDFileJPEG.cpp
@@ -17,9 +17,9 @@
 #include <iocsh.h>
 
 #include <asynDriver.h>
+#include "NDPluginFile.h"
 
 #include <epicsExport.h>
-#include "NDPluginFile.h"
 #include "NDFileJPEG.h"
 
 
--- a/ADApp/pluginSrc/NDFileMagick.cpp
+++ b/ADApp/pluginSrc/NDFileMagick.cpp
@@ -17,9 +17,9 @@
 #include <iocsh.h>
 
 #include <asynDriver.h>
+#include "NDPluginFile.h"
 
 #include <epicsExport.h>
-#include "NDPluginFile.h"
 #include "NDFileMagick.h"
 
 static const char *driverName = "NDFileMagick";
--- a/ADApp/pluginSrc/NDFileNetCDF.cpp
+++ b/ADApp/pluginSrc/NDFileNetCDF.cpp
@@ -18,9 +18,9 @@
 #include <iocsh.h>
 
 #include <asynDriver.h>
+#include "NDPluginFile.h"
 
 #include <epicsExport.h>
-#include "NDPluginFile.h"
 #include "NDFileNetCDF.h"
 
 #define MAX_ATTRIBUTE_STRING_SIZE 256
--- a/ADApp/pluginSrc/NDFileNexus.cpp
+++ b/ADApp/pluginSrc/NDFileNexus.cpp
@@ -22,9 +22,9 @@
 #include <string.h>
 
 #include <asynDriver.h>
+#include "NDPluginFile.h"
 
 #include <epicsExport.h>
-#include "NDPluginFile.h"
 #include "NDFileNexus.h"
 
 static const char *driverName="NDFileNexus";
--- a/ADApp/pluginSrc/NDFileNull.cpp
+++ b/ADApp/pluginSrc/NDFileNull.cpp
@@ -18,6 +18,7 @@
 #include <iocsh.h>
 
 #include <asynDriver.h>
+#include "NDPluginFile.h"
 
 #include <epicsExport.h>
 #include "NDFileNull.h"
--- a/ADApp/pluginSrc/NDFileTIFF.cpp
+++ b/ADApp/pluginSrc/NDFileTIFF.cpp
@@ -31,10 +31,9 @@
 #include <iocsh.h>
 
 #include <asynDriver.h>
+#include "NDPluginFile.h"
 
 #include <epicsExport.h>
-#include "NDPluginFile.h"
-#include "tiffio.h"
 #include "NDFileTIFF.h"
 
 #define STRING_BUFFER_SIZE 2048
--- a/ADApp/pluginSrc/NDPluginAttrPlot.cpp
+++ b/ADApp/pluginSrc/NDPluginAttrPlot.cpp
@@ -8,6 +8,8 @@
 #include <epicsMath.h>
 #include <epicsThread.h>
 
+#include "NDPluginDriver.h"
+
 #include <epicsExport.h>
 #include "NDPluginAttrPlot.h"
 #include "CircularBuffer.h"
--- a/ADApp/pluginSrc/NDPluginAttribute.cpp
+++ b/ADApp/pluginSrc/NDPluginAttribute.cpp
@@ -21,9 +21,9 @@
 #include <iocsh.h>
 
 #include <asynDriver.h>
+#include "NDPluginDriver.h"
 
 #include <epicsExport.h>
-#include "NDPluginDriver.h"
 #include "NDPluginAttribute.h"
 
 const epicsInt32 NDPluginAttribute::MAX_ATTR_NAME_      = 256;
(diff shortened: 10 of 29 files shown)


MarkRivers:
I believe the principle that Andrew Johnson explained to me is that any header files that define classes that are in the current DLL should be included after epicsExports.h.  

Consider NDFileJPEG.h. NDPluginFile.h defines classes that will be in the current DLL (NDPlugin.dll) so it should come after epicsExports.h.  

In this case it is actually moot, and NDPluginFile.h does not actually need to be included in NDFileJPEG.cpp at all, because it is included in NDFileJPEG.h.  But I think that if it is included in NDFileJPEG.cpp it is logically more correct, and clearer code, to included it after epicsExport.h because that makes clear that it is defining things that are included in the current DLL.

@anjohnson can you provide advice?

xiaoqiangwang:
The different order affects the parent class declaration. In my view, only the implementation cpp file should include the header after "epicsExport.h".

In this case `NDFileJPEG.cpp` does not implement class NDPluginFile, so it is actually _importing_ a class definition.
```
#include "NDPluginFile.h"
#include <epicsExport.h>
#include "NDFileJPEG.h"
```
In the above code, NDPluginFile expands to
```
class __declspec(dllimport) NDPluginFile : public NDPluginDriver
```

If `NDPluginFile.h` comes after `epicsExport.h` or not included at all, the NDPluginFile class will be
```
class __declspec(dllexport) NDPluginFile : public NDPluginDriver
```
which is clearly not what it meant.

MarkRivers:
I think that 

```
class __declspec(dllexport) NDPluginFile : public NDPluginDriver
```

is what is meant. The DLL we are building with this Makefile exports the NDPluginFile class, it does not import it.

xiaoqiangwang:
It expands to the same when compiler compiles `NDPluginFile.cpp`
```
class __declspec(dllexport) NDPluginFile : public NDPluginDriver
```

The linker is smart enough to figure out where is the real exports is. But logically I see that a class should only be exported once.

anjohnson:
The MinGW linker is generally more strict than the MSVC one when it comes to these decorations, so trying to build for the target `windows-x64-mingw` will usually fail if you have them wrong.

I don't know this specific code-base very well but it looks like all plugins get linked into the same library (there is only one LIBRARY built by this Makefile), so all the local plugin header files should appear _after_ the epicsExport.h in these source files.

@xiaoqiangwang are you proposing these changes because you are seeing build errors? If so please post the command-line and the actual error messages.

xiaoqiangwang:
A bit background is needed here: At PSI we use Dirk Zimoch's makefile to build epics modules. In that process, one shared library is created out of one epics module. For ADCore, I build the source of ADSrc and pluginSrc together to create an ADCore shared library.

In such a scenario, the linker would create a rather cryptic error,
```
NDPluginDriver.obj : error LNK2005: "public: __cdecl asynPortDriver::asynPortDriver(class asynPortDriver const &)" (??0asynPortDriver@@QEAA@AEBV0@@Z) already defined in asyn.lib(asyn.dll)
NDPluginDriver.obj : error LNK2005: "public: class asynPortDriver & __cdecl asynPortDriver::operator=(class asynPortDriver const &)" (??4asynPortDriver@@QEAAAEAV0@AEBV0@@Z) already defined in asyn.lib(asyn.dll)
ADCore.dll : fatal error LNK1169: one or more multiply defined symbols found
```

Initially I thought the default copy constructors of asynPortDriver are the trouble, I manually disabled them (putting them to _private_), and that makes the linker error disappear. 

But later on by inspecting the orders of header inclusion, I figure it could be the side effect of the symbol exports in the chain of asynPortDriver -> asynNDArrayDriver -> NDPluginDriver ->NDPluginFile. I did the re-order of the inclusion so that only the implementation cpp file gets to do its own export. That have solved the linker error.

MarkRivers:
If the plugin library is built with this PR then NDPluginFile.obj will define
```
class __declspec(dllexport) NDPluginFile : public NDPluginDriver
```
but NDFileJPEG.obj will define:
```
class __declspec(dllimport) NDPluginFile : public NDPluginDriver
```
The linker will then be putting 2 conflicting definition of the same class in the same library.  @xiaoqiangwang says the linker does not complain, but I don't think that is guaranteed for all linkers, or even for future versions of Microsoft's linker.

More importantly if you use Dirk's method to build a single library for ADCore, rather than 2 libraries, you compound this problem.  For example, ADApp\ADSrc\asynNDArrayDriver.obj will define 
```
class __declspec(dllexport) asynNDArrayDriver : public asynPortDriver 
```
But ADApp\pluginSrc\NDPluginDriver.obj will define
```
class __declspec(dllimport) asynNDArrayDriver : public asynPortDriver 
```
That is normally fine, because we are putting those 2 object files in different libraries.  But with Dirk's method those are now being put the same library, which strikes me as an ill-advised thing to do which may not work on all linkers.

xiaoqiangwang:
I assume the linker chooses dllexport over dllimport, if both are present for the same symbol., although  I couldn't find the official statement about this.

Back to the original error message. It complains about repeated copy constructors duplicated. I inspected generated NDPlugin.dll (using normal synApps building system), 
```
# Note: This runs on msvc toolchain on Linux via WINE 
$ /opt/wine-msvc-2017/bin/x64/dumpbin /exports NDPlugin.dll  | grep asynPortDriver
         76   4B 00003380 ??0asynPortDriver@@QEAA@AEBV0@@Z
        156   9B 000034C0 ??4asynPortDriver@@QEAAAEAV0@AEBV0@@Z
        227   E2 000A84E8 ??_7asynPortDriver@@6B@
```
The first two are the copy constructors and the last is the virtual function table. 

Also symbols about asynNDArrayDriver(ADBase.dll) are also (re-)exported.
```
$ /opt/wine-msvc-2017/bin/x64/dumpbin /exports bin/windows-x64/NDPlugin.dll  | grep asynNDArrayDriver
         75   4A 00003B90 ??0asynNDArrayDriver@@QEAA@AEBV0@@Z
        153   98 00003F30 ??4asynNDArrayDriver@@QEAAAEAV0@AEBV0@@Z
        ...
```

With this PR, only locally defined classes are exported.

MarkRivers:
Please do the same search, but grep for epicsTime instead.  

When I run dumpbin exports on NDPlugin.dll on Windows I the output includes the following:
```
        155   9A 00005060 ??4asynNDArrayDriver@@QEAAAEAV0@AEBV0@@Z
        156   9B 000043D0 ??4asynPortDriver@@QEAAAEAV0@AEBV0@@Z
        157   9C 00047F90 ??4epicsTime@@QEAAAEAV0@$$QEAV0@@Z
        158   9D 00048180 ??4epicsTime@@QEAAAEAV0@AEBUepicsTimeStamp@@@Z
        159   9E 00048120 ??4epicsTime@@QEAAAEAV0@AEBUgm_tm_nano_sec@@@Z
        160   9F 000481B0 ??4epicsTime@@QEAAAEAV0@AEBUl_fp@@@Z
        161   A0 000480F0 ??4epicsTime@@QEAAAEAV0@AEBUlocal_tm_nano_sec@@@Z
        162   A1 000481E0 ??4epicsTime@@QEAAAEAV0@AEBUtime_t_wrapper@@@Z
        163   A2 00048150 ??4epicsTime@@QEAAAEAV0@AEBUtimespec@@@Z
        164   A3 00047F80 ??4epicsTime@@QEAAAEAV0@AEBV0@@Z
        165   A4 00047F70 ??4epicsTimeEvent@@QEAAAEAV0@$$QEAV0@@Z
        166   A5 00047F70 ??4epicsTimeEvent@@QEAAAEAV0@AEBV0@@Z
        167   A6 00048070 ??8epicsTime@@QEBA_NAEBV0@@Z
        168   A7 00048090 ??9epicsTime@@QEBA_NAEBV0@@Z
        169   A8 00047FA0 ??BepicsTimeEvent@@QEBAHXZ
        170   A9 00047FB0 ??GepicsTime@@QEBA?AV0@AEBN@Z
        171   AA 000480D0 ??OepicsTime@@QEBA_NAEBV0@@Z
        172   AB 000480B0 ??PepicsTime@@QEBA_NAEBV0@@Z
        173   AC 00047FE0 ??YepicsTime@@QEAA?AV0@AEBN@Z
        174   AD 00048020 ??ZepicsTime@@QEAA?AV0@AEBN@Z
        175   AE 000A8DC0 ??_7NDArray@@6B@
        176   AF 000A8DD0 ??_7NDArrayPool@@6B@
        177   B0 000AF338 ??_7NDFileHDF5@@6BasynNDArrayDriver@@@
```
The symbols for epicsTime demangle into things like this:
```
public: class epicsTime & __ptr64 __cdecl epicsTime::operator=(struct l_fp const & __ptr64) __ptr64
```
So NDPlugin.dll is also exporting things from EPICS base, but that does not seem to be a problem.

Why is it a problem for asynPortDriver, but not a problem for epicsTime?

xiaoqiangwang:
Of course I am wrong in saying this
> I assume the linker chooses dllexport over dllimport, if both are present for the same symbol., although I couldn't find the official statement about this.

The linker actually will complain if I build passing `OPT_WHOLE_PROGRAM=NO` to `make`.
```
NDPosPlugin.obj : warning LNK4049: locally defined symbol ??0NDPluginDriver@@QAE@PBDHH0HHHIHHHHHHH_N@Z (public: __thiscall NDPluginDriver::NDPluginDriver(char const *,int,int,cha  const *,int,int,int,unsigned int,int,int,int,int,int,int,int,bool)) imported
NDPluginFile.obj : warning LNK4049: locally defined symbol ??0NDPluginDriver@@QAE@PBDHH0HHHIHHHHHHH_N@Z (public: __thiscall NDPluginDriver::NDPluginDriver(char const *,int,int,ch r const *,int,int,int,unsigned int,int,int,int,int,int,int,int,bool)) imported
NDPluginPva.obj : warning LNK4049: locally defined symbol ??0NDPluginDriver@@QAE@PBDHH0HHHIHHHHHHH_N@Z (public: __thiscall NDPluginDriver::NDPluginDriver(char const *,int,int,cha  const *,int,int,int,unsigned int,int,int,int,int,int,int,int,bool)) imported
...
```
So this PR is not a fix. I will take it back.

But as you said, the question why the other re-exported symbols are not a problem to the linker is still to be answered.

PR #438, opened 2020-04-20
Clean includes

xiaoqiangwang:
Somehow related to (retracted) #436, it is to address those symbols exported by ADBase and NDPlugin, but originally exported by libCom and asyn.

It is tested with windows-x64 arch.

### DIFF
--- a/ADApp/ADSrc/NDArray.h
+++ b/ADApp/ADSrc/NDArray.h
@@ -13,9 +13,20 @@
 #define NDArray_H
 
 #include <set>
+#include <stdio.h>
+
+#ifdef epicsExportSharedSymbols
+#   define epicsExportSharedSymbols_NDArrayH
+#   undef epicsExportSharedSymbols
+#endif
+
 #include <epicsMutex.h>
 #include <epicsTime.h>
-#include <stdio.h>
+
+#ifdef epicsExportSharedSymbols_NDArrayH
+#   define epicsExportSharedSymbols
+#   include <shareLib.h>
+#endif
 
 #include "NDAttribute.h"
 #include "NDAttributeList.h"
--- a/ADApp/ADSrc/NDAttribute.h
+++ b/ADApp/ADSrc/NDAttribute.h
@@ -14,11 +14,21 @@
 #include <stdio.h>
 #include <string.h>
 
+#ifdef epicsExportSharedSymbols
+#   define epicsExportSharedSymbols_NDAttributeH
+#   undef epicsExportSharedSymbols
+#endif
+
 #include <ellLib.h>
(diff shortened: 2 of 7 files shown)


MarkRivers:
@xiaoqiangwang  I agree with your changes to the .cpp files.  I have fixed these files on the master branch, and pushed to Github.  You will need to merge the master branch into yours because there are now conflicts.

I don't understand your changes to the .h files.  Can you please explain what errors you are getting that these changes are fixing.  I don't see any problems locally, on the Travis builds, or on the Jenkins builds.

xiaoqiangwang:
NDArray.h
```
 #include <epicsMutex.h>
 #include <epicsTime.h>
 #include <stdio.h>
```
It is included by NDArray.cpp
```
#include <epicsExport.h>
#include "NDArray.h"
```
The macro _epicsShareFunc_  expands to  ___declspec(dllexport)_ for all headers included by NDArray.h. That would be wrong for functions from epicsTime.h, which are not exported here. The extra block undefines _epicsExportSharedSymbols_ so that _epicsShareFunc_ will expand to ___declspec(dllimport)_.
```
#ifdef epicsExportSharedSymbols
 #   define epicsExportSharedSymbols_NDArrayH
 #   undef epicsExportSharedSymbols
 #endif
```

I saw the this usage in the base until 7.0.4
https://github.com/epics-base/epics-base/blob/R3.15.5/src/ca/client/db_access.h#L20

From base 7.0.4 however, it starts to use a unique macro per module to export symbols, see https://github.com/epics-base/epics-base/blob/951b6acbbc102a5865592f671d2b410e96c5a4d3/documentation/RELEASE_NOTES.md#_api-macros-in-epics-headers.

If all modules adopt the new mechanism, this change is then not needed.

MarkRivers:
The alternate way to handle this is just to make sure that the .cpp file includes <epicsTime.h> before epicsExport.h.  That is the way I have been handling this previously.

Did you actually see a warning or error with the existing code?  I have not, so I am wondering how you found the issue?

xiaoqiangwang:
I think this is what you have seen, the epicsTime* symbols are seen from NDPlugin.dll.
https://github.com/areaDetector/ADCore/pull/436#issuecomment-615295506

MarkRivers:
I just pushed a commit that duplicates all of the includes in NDArray.h in the .cpp files before epicsExport.h is included or epicsExportSharedSymbols is defined.  This fixes the problem with epicsTime symbols being defined in NDPlugin.dll.

This may not be the cleanest way to do it, since it requires changing .cpp files if a new file is included in NDArray.h.  

I have created a new issue #455 for this.  It should be done in all of ADCore and driver repositories.

Issue #446, opened 2020-06-16
PVI Proposal


GDYendell:
At DLS we have been working on an idea of Tom Cobb's making: PVI (PV Interface).

There is an outline hosted on readthedocs: https://pvi.readthedocs.io/en/latest/contents.html

As a brief summary, the goal of PVI is to consolidate the duplicated boiler plate of an EPICS driver (templates, low-level screens and the parameter creation in the c++ code) into one place, allowing these pieces to be generated instead. After an initial YAML file generation from the existing template file, these files can all be generated consistently from a single source.

To begin with I have made 3 PRs:

- asyn - https://github.com/epics-modules/asyn/pull/117
- ADCore - https://github.com/areaDetector/ADCore/pull/447
- ADSimDetector - https://github.com/areaDetector/ADSimDetector/pull/18

as this represents the full stack to get an areaDetector driver working under the PVI framework.

**asyn**
Creates new asynParam and asynParamSet classes to store the parameter definitions separately from the driver class and updates asynPortDriver to have an asynParamSet and call createParam on all the parameters stored within it from child classes. There is a new asynPortDriver constructor that takes a asynParamSet and calls `createParams`.

**ADCore**
Create new asynNDArrayDriverParamSet and ADDriverParamSet classes and update asynNDArrayDriver and ADDriver constructors to take an instance of their specific asynParamSet derived class. The parameter definitions are split out into the param sets and plugins and drivers are then able to inherit these classes to add their own parameter definitions on top and then pass the resulting param set class to the parent driver class for creation. This is It also updates NDPluginDriver to inherit from ADDriverParamSet, so that no changes are required for its children so that they can still call createParam and access parameters directly rather than via paramSet->, for now.

**ADSimDetector**
Update simDetector to make use of the updated classes in ADCore and asyn. Split the parameter definitions into simDetectorParamSet and pass the instance to ADDriver to have the parameters created. Parameters access updated to use its paramSet pointer.

The proposed class hierarchy for these 3 modules is as follows:
![image](https://user-images.githubusercontent.com/15689569/84792891-57e52880-afec-11ea-9276-d7d90cb72462.png)

PR #447, opened 2020-06-16
PVI

GDYendell:
Create new asynNDArrayDriverParamSet and ADDriverParamSet classes and update asynNDArrayDriver and ADDriver constructors to take an instance of their specific asynParamSet derived class. The parameter definitions are split out into the param sets and plugins and drivers are then able to inherit these classes to add their own parameter definitions on top and then pass the resulting param set class to the parent driver class for creation. This is It also updates NDPluginDriver to inherit from ADDriverParamSet, so that no changes are required for its children so that they can still call createParam and access parameters directly rather than via paramSet->, for now.

For full context see: areaDetector/ADCore#446

---
**Update 31/01/23**

I have updated this branch and it is now a reasonable demonstration of what pvi can do. 

As it stands, this PR converts asynNDArrayDriver, ADDriver, NDPluginDriver, NDPluginFile and NDFileHDF5 to use pvi and half converts all the other plugins so that they work unchanged (via multiple inheritance of the parent ParamSet). The plugin tests have been updated and pass. I have been testing via GitHub Actions [here](https://github.com/dls-controls/ADCore/actions). It does a build with and without pvi available to make sure checked in files are valid and that the generated files are correct, e.g. [here](https://github.com/dls-controls/ADCore/runs/7862931368). I also did a build [here](https://github.com/dls-controls/ADCore/runs/7862930703) to test the CI works without any existing pvi generated files checked in (only the pvi=true build works).

The process of creating this branch was
- Run `pvi convert ...` for each driver
- Manually half convert NDPluginAttrPlot - the autoconvert failed because there was more than one class in the file.
- Manually fix NDFileHDF5 - Convert does not handle parameters in arrays, PVs missing
- Update Makefiles
- Update tests to create ParamSets for drivers
- Run `pvi regroup ...` for each driver with a manually selected set of adl files
- Run `make` and check in results

UIs (adl, edl and bob) are generated as part of the build in a `pvi` subdirectory next to `autoconvert`. These are low level screens containing every PV in the yaml file, organised into groups. Here are some generated UIs for NDFileHDF5 (including all parent parameters).

adl
edl
bob

Note:
- The layouts can be configured, such as the max height, the widths of the labels and widgets. For example in these screenshots for adl the widget width is wider to better match the existing hand created screens.
- These have been grouped automatically based on the adl screens, but can be adjusted by passing different adl files to the conversion or by editing the yaml files.

These are some of the things pvi could allow us to do:
- Have ophyd read the PVs of each driver in an IOC at runtime and propagate them into members and methods of ophyd devices These would match the generated screens.
- Ability to define parameters to load/save for scans
- Live introspection of PandA parameters, which are only created dynamically based on hardware during IOC boot. This will then generate and serve phoebus screens over http and populate ophyd devices at runtime.
- It could be possible to do a similar thing for GigE cameras to create PVs exactly matching the parameters the camera provides

Example python:

```{python}
Out: param_tree = await caget("TESTSIMDETECTOR:CAM:PVI", datatype=DBR_CHAR_STR)
In: param_tree_dict = json.loads(param_tree)
In: [(param["name"], param["type"]) for group in param_tree_dict['parameters'] for param in group["children"]]
Out: 
[('PortName', 'SignalR'),
 ('Manufacturer', 'SignalR'),
 ('Model', 'SignalR'),
 ('SerialNumber', 'SignalR'),
 ('FirmwareVersion', 'SignalR'),
...
 ('Noise', 'SignalRW'),
 ('PeakWidthY', 'SignalRW'),
 ('XSineOperation', 'SignalRW'),
 ('YSineOperation', 'SignalRW'),
 ('PeakVariation', 'SignalRW'),
 ('Reset', 'SignalRW'),
 ('SimMode', 'SignalRW')]
In: param_tree_dict['parameters'][-1]['children'][-1]
Out: 
{'type': 'SignalRW',
 'name': 'SimMode',
 'label': '',
 'pv': 'SimMode',
 'widget': {'type': 'ComboBox'},
 'read_pv': 'SimMode_RBV',
 'read_widget': {'type': 'TextRead', 'lines': 1}}
```

CSV for embed in docs:

These currently live in the `pvi` directory, but should be generated in place somewhere as part of the docs build.
csv

Outstanding Issues:
- Currently the records for NDFile have been moved down into asynNDArrayDriver.pvi.producer.yaml and NDFile.pvi.producer.yaml is empty.
    - The effect of this is that the file writing PVs are available in all plugins and drivers and appear on the generated UIs
    - This is because pvi does not support the way this is done currently (asyn parameters defined in asynNDArrayDriver source, but records in NDFile.template)
    - I imagine it would be possible to support this through pvi, but this may be an opportunity to rework this logic.
      - NDFile.pvi.producer.yaml could become a mixin that defines a ParamSet.h and Parameters.template without creating a driver that is used explicitly. This can then be used by pilatusDetector and the file writing plugins through multiple inheritance. This way the file writing asyn parameters will not be available in all drivers and plugins, as they are currently.
- There is some tidying to do of lines inserted into the source and templates in the simplest way that was functional, e.g. template includes at the very top of the file, initialiser list out of order. These should be fixed, but they don't cause any problems so I have left until things are finalised.
- It would be good to use the automatic make include magic used by sncseq, but I found that it included RULES_PVI too late, so the compile step came before pvi regenerate. This meant a subsequent make was required to see the updated header files and recompile. This is probably because the prerequisites are not correct. For now the Makefiles have to include RULES_PVI explicitly, and before RULES.
- There are some limitations in the UI generation. These are probably all solvable, but whether it is worthwhile depends on what they are used for.
  - Tables - e.g. SizeX and SizeY on the same line
  - Full width widgets for long strings
  - Widgets for soft records, e.g. ArrayRate_RBV
  - Widgets for individual fields
  - Override default widget type for a record type - e.g. Start/Stop buttons for bi rather than a toggle switch

Note this is based on #483 to make use of the CI build. The first commit will be removed once that is merged to master.


MarkRivers:
This PR has lots of conflicts.  I believe these are coming because you removed trailing whitespace before issuing the PR, but subsequently I have removed trailing whitespace in all files on the master branch.  This also makes it very difficult to see just what files you really changed.

Can you merge master into the dls-controls:pvi branch so these conflicts and apparent changes go away?

MarkRivers:
I think you should remove the .dls-build-master file.

GDYendell:
Thanks for making the whitespace changes, that makes things easier. The diff should be clearer now.

GDYendell:
The PVI PV inserted into the generated templates was to be used to introspect all of the PVs a driver has. We intend to rmove this pV and the json in the ParamSet.h and replace it with info tags on each record, which will create a V4 PV with a flat dictionary of the PVs in the driver. The grouping is not needed at runtime and this is simpler.

GDYendell:
We are now converting templates to a yaml representation for generating UIs and an API for ophyd, so we have what we need. We could revisit this in future.

Issue #448, opened 2020-07-07
Improve support for disconnected drivers


MarkRivers:
Most areaDetector drivers do not do a good job of indicating when the underlying hardware is not accessible.  The user is often left looking at an OPI screen with no clear indication of a problem.  They are forced to look at the IOC output to see that there were errors.

The best way to handle this is for the areaDetector driver to signal to asynManager that it is not "connected".   If that is done then all existing OPI screens will show that the driver is Disconnected in large red letters.  If the user attempts to modify any widget in the OPI screen they should get an error because the caput fails.

MarkRivers:
Closed via #449

PR #449, opened 2020-07-07
Improve disconnect

MarkRivers:
This PR addresses issue #448.

It does the following:
- Adds a new bool member variable `deviceIsConnected` to ADDriver.
- Initializes `deviceIsConnected` to `true` in the ADDriver constructor.
- Adds the `ADDriver::connect()` method.  This method looks at the value of `deviceIsConnected`.  If it is `true` then it calls `asynNDArrayDriver::connect()`, which ultimately that calls `asynPortDriver::connect()`.  That method calls `pasynManager->exceptionConnect()`.  This method is needed to prevent asynManager from autoconnecting to the driver when `deviceIsConnected` is `false`.

Real drivers can now signal that they have failed to connect to hardware, and thus set the driver to the `Disconnected` state by doing the following if hardware initialization fails, typically in their constructor:
- Set `this->deviceIsConnected = false`
- Call `this->disconnect(pasynUserSelf)`

The constructor in ADSimDetector was changed with commented out lines as follows to take advantage of this new feature:
```
corvette:ADSimDetector/simDetectorApp/src>git diff  de317041980e95752ab2e31de35a6d55368306a3 simDetector.cpp
diff --git a/simDetectorApp/src/simDetector.cpp b/simDetectorApp/src/simDetector.cpp
index 4e64182..17687b7 100644
--- a/simDetectorApp/src/simDetector.cpp
+++ b/simDetectorApp/src/simDetector.cpp
@@ -1113,6 +1113,11 @@ simDetector::simDetector(const char *portName, int maxSizeX, int maxSizeY, NDDat
         return;
     }

+    // Test setting the device connected to false.
+    // Real drivers should do this if they fail to connect to the hardware
+    //this->deviceIsConnected = false;
+    //this->disconnect(pasynUserSelf);
+
     /* Create the thread that updates the images */
     status = (epicsThreadCreate("SimDetTask",
                                 epicsThreadPriorityMedium,
```

With those 2 lines uncommented the OPI screen looks like this when the IOC is run:

simdisconnect

The constructor in ADPerkinElmer has had the following change:

```
diff --git a/perkinElmerApp/src/PerkinElmer.cpp b/perkinElmerApp/src/PerkinElmer.cpp
index 4f3aa64..4990ed6 100755
--- a/perkinElmerApp/src/PerkinElmer.cpp
+++ b/perkinElmerApp/src/PerkinElmer.cpp
@@ -184,7 +184,10 @@ PerkinElmer::PerkinElmer(const char *portName,  int IDType, const char *IDValue,
                               epicsThreadGetStackSize(epicsThreadStackMedium),
                               (EPICSTHREADFUNC)acquireStopTaskC,
                               this) == NULL);
-  initializeDetector();
+  if (!initializeDetector()) {
+    this->deviceIsConnected = false;
+    this->disconnect(pasynUserSelf);
+  }

   // Set exit handler to clean up
   epicsAtExit(exitCallbackC, this);
```

With this change the OPI screen now shows the following when the IOC is run with the detector disconnected.

image

### DIFF
--- a/ADApp/ADSrc/ADDriver.cpp
+++ b/ADApp/ADSrc/ADDriver.cpp
@@ -54,6 +54,24 @@ void ADDriver::setShutter(int open)
     }
 }
 
+/** Connects driver to device;
+  * This method uses the class variable deviceIsConnected to determine whether to call
+  * the base class method, which in turn calls pasynManager::exceptionConnect().
+  * \param[in] pasynUser The pasynUser structure which contains information about the port and address */
+asynStatus ADDriver::connect(asynUser *pasynUser)
+{
+    static const char *functionName = "connect";
+
+    if (this->deviceIsConnected) {
+        return asynNDArrayDriver::connect(pasynUser);
+    }
+    asynPrint(pasynUserSelf, ASYN_TRACE_WARNING,
+        "%s::%s warning attempt to connect but deviceIsConnected is false\n",
+        driverName, functionName);
+    return asynSuccess;
+}
+
+
 /** Sets an int32 parameter.
   * \param[in] pasynUser asynUser structure that contains the function code in pasynUser->reason.
   * \param[in] value The value for this parameter
@@ -110,7 +128,8 @@ ADDriver::ADDriver(const char *portName, int maxAddr, int numParams, int maxBuff
     : asynNDArrayDriver(portName, maxAddr, maxBuffers, maxMemory,
           interfaceMask | asynInt32Mask | asynFloat64Mask | asynOctetMask | asynGenericPointerMask | asynDrvUserMask,
           interruptMask | asynInt32Mask | asynFloat64Mask | asynOctetMask | asynGenericPointerMask,
-          asynFlags, autoConnect, priority, stackSize)
+          asynFlags, autoConnect, priority, stackSize),
+    deviceIsConnected(true)
 
 {
     //char *functionName = "ADDriver";
--- a/ADApp/ADSrc/ADDriver.cpp
+++ b/ADApp/ADSrc/ADDriver.cpp
@@ -55,18 +55,23 @@ void ADDriver::setShutter(int open)
 }
 
 /** Connects driver to device;
-  * This method uses the class variable deviceIsConnected to determine whether to call
-  * the base class method, which in turn calls pasynManager::exceptionConnect().
+  * This method is called when the driver's pasynCommon->connect() function is called.
+  * It uses the class variable deviceIsReachable to determine whether to call
+  * asynPortDriver::connect(), which in turn calls pasynManager::exceptionConnect()
+  * to signal that the driver is connected to the underlying hardware.
+  * Derived classes can override this method if they need to handle connect() calls in a more
+  * complex way.  For example, with a network camera that can be temporarily unreachable
+  * the driver could attempt to connect to the camera each time that connect() is called.
   * \param[in] pasynUser The pasynUser structure which contains information about the port and address */
 asynStatus ADDriver::connect(asynUser *pasynUser)
 {
     static const char *functionName = "connect";
 
-    if (this->deviceIsConnected) {
-        return asynNDArrayDriver::connect(pasynUser);
+    if (this->deviceIsReachable) {
+        return asynPortDriver::connect(pasynUser);
     }
     asynPrint(pasynUserSelf, ASYN_TRACE_WARNING,
-        "%s::%s warning attempt to connect but deviceIsConnected is false\n",
+        "%s::%s warning attempt to connect but deviceIsReachable is false\n",
         driverName, functionName);
     return asynSuccess;
 }
@@ -129,7 +134,7 @@ ADDriver::ADDriver(const char *portName, int maxAddr, int numParams, int maxBuff
           interfaceMask | asynInt32Mask | asynFloat64Mask | asynOctetMask | asynGenericPointerMask | asynDrvUserMask,
           interruptMask | asynInt32Mask | asynFloat64Mask | asynOctetMask | asynGenericPointerMask,
           asynFlags, autoConnect, priority, stackSize),
-    deviceIsConnected(true)
+    deviceIsReachable(true)
 
 {
     //char *functionName = "ADDriver";
--- a/ADApp/ADSrc/ADDriver.h
+++ b/ADApp/ADSrc/ADDriver.h
@@ -139,6 +139,7 @@ public:
 
     /* These are the methods that we override from asynPortDriver */
     virtual asynStatus writeInt32(asynUser *pasynUser, epicsInt32 value);
+    virtual asynStatus connect(asynUser *pasynUser);
 
     /* These are the methods that are new to this class */
     virtual void setShutter(int open);
@@ -179,6 +180,8 @@ protected:
     int ADStatusMessage;
     int ADStringToServer;
     int ADStringFromServer;
+    
+    bool deviceIsConnected;
 };
 
 #endif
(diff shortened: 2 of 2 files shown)


GDYendell:
This looks like a good idea - it is a common issue.

I haven't yet had an in detail look at the logic flow of this through asyn and ADSimDetector yet, so maybe I am missing something. These are my initial thoughts:

- Where is ADDriver::connect meant to be called?

- What PV is that Disconnected status on the screen above? Is that controlled by calling `exceptionConnect`?

- Maybe it is just the naming, but this snippet
```
if (this->deviceIsConnected) {
        return asynNDArrayDriver::connect(pasynUser);
}
```
is counter-intuitive to me. If it is already connected, why do you then call connect? exceptionConnect looks like it just does some housekeeping to register the fact a connect took place. I guess you know this so it is obvious to you, but as this is in a relatively high level class a comment explaining this distinction would be useful. In the same way the error

`attempt to connect but deviceIsConnected is false`

seems strange. The state `!deviceIsConnected` seems to me the exact state in which you would call `connect`, because you are not connected and you want to be.

- Am I right in thinking that asynNDArrayPortDriver::connect just calls directly into asynPortDriver::connect, because it is not overridden? If so would it be clearer to either call into asynPortDriver::connect explicitly or create an asynNDArrayPortDriver::connect and just explicitly call the base class method?

MarkRivers:
> Where is ADDriver::connect meant to be called?

It will be called by any code that calls pasynCommon->connect() on that port. It will be called by asynManager automatically if:
- The port is currently not connected
- The port is configured with autoConnect = true.

> What PV is that Disconnected status on the screen above? Is that controlled by calling exceptionConnect?

It is asynIO.CNCT PV, which is the .CNCT field of the asyn record that is automatically loaded for each driver.  Its value is the status returned by asynManager->isConnected() for that port.  Drivers signal asynManager that they are connected or disconnected by calling pasynManager->exceptionConnect() and pasynManager->exceptionDisconnect().  The asyn record registers for callbacks on those exceptions, and updates the .CNCT field appropriately.

> Maybe it is just the naming, but this snippet is counter-intuitive to me. If it is already connected, why do you then call connect? exceptionConnect looks like it just does some housekeeping to register the fact a connect took place. 
```
if (this->deviceIsConnected) {
        return asynNDArrayDriver::connect(pasynUser);
}
```
The way asyn connection works is that asynManager (or other code) calls pasynCommon->connect() in the driver.  In the case of asynPortDriver/ADDriver that results in a call to ADDriver::connect().  The pasynCommon->connect() does the actual connection to the device (for example creating a socket in the case of the TCP driver).  If it succeeds it calls pasynManager->exceptionConnect() to signal that it succeeded.  If it fails it does not call pasynManager->exceptionConnect().  asynPortDriver::connect() always calls pasynManager->exceptionConnect().  In most cases asynPortDriver::connect() is not overridden, and calls to pasynCommon->connect() always succeed.  What this PR does is add the ability for drivers to override the method and therefore not call pasynManager->exceptionConnect() if deviceIsConnected is false.

deviceIsConnected is a flag that is used to allow a call to pasynCommon->connect() to succeed or fail.

> The state !deviceIsConnected seems to me the exact state in which you would call connect, because you are not connected and you want to be.

There are 2 different types of state flags:
- asynManager maintains a connection state for each asyn port.  If it is not connected and autoConnect is true it will call the pasynCommon->connect() method to try to connect.  asyn clients can call pasynManager->isConnected() to query this state.
- deviceIsConnected is a flag in ADDriver that means the hardware is reachable.  If that flag is false then a call to pasynCommon->connect() should fail.

This PR is really intended to make it easy to handle devices that failed to connect in the constructor, and will never be able to connect, for example PCI devices, etc. 

For devices like GigE cameras, the driver (e.g. ADAravis) should itself override the connect() method so that it can keep trying to connect even if it failed in the constructor.  In that case the code in this PR will not be used, because ADDriver::connect() will be overridden.  

>Am I right in thinking that asynNDArrayPortDriver::connect just calls directly into asynPortDriver::connect, because it is not overridden? If so would it be clearer to either call into asynPortDriver::connect explicitly or create an asynNDArrayPortDriver::connect and just explicitly call the base class method?

I thought about that.  But it seems more future-proof to just call down to the next base class in case we decide there is a good reason to add a connect() method there.  We already do this with other methods (readInt32(), writeInt32(), etc.) that may or may not be implemented in asynNDArrayDriver.

jwlodek:
Hi, just a question about this because I have been trying to allow for something similar myself for the `ADUVC` driver. As I understand, this PR is mainly aimed at failed connections during the initial startup/constructor call. I think this is definitely a valuable improvement, but I also have a slightly different though similar case. 

I was also not sure how to handle the case where the camera connects OK, but then is subsequently unplugged. Is there a way to have some kind of periodic check to see if the camera remains connected? Or even to check if the camera is still connected when we call `writeInt32` etc? I have had cases where cheap USB cameras running with `ADUVC` went into standby mode without the IOC realizing, making the device handle in the driver stale, and causing the IOC to crash when any camera operation was performed, and this could have been prevented by pressing `Connect` again on the OPI screen.

This could be something like a base function in `ADDriver` that is overridden by each driver that uses whatever helper library to poll for a camera connection, and could be called at some pre-set interval (maybe set as another PV)?

MarkRivers:
I have updated the ADDexela driver to set its state to disconnected if it cannot connect to the hardware in the constructor.

image

MarkRivers:
> I was also not sure how to handle the case where the camera connects OK, but then is subsequently unplugged. Is there a way to have some kind of periodic check to see if the camera remains connected? Or even to check if the camera is still connected when we call writeInt32 etc? I have had cases where cheap USB cameras running with ADUVC went into standby mode without the IOC realizing, making the device handle in the driver stale, and causing the IOC to crash when any camera operation was performed, and this could have been prevented by pressing Connect again on the OPI screen.

Do you have a way to detect that the USB camera is still connected without crashing the IOC?

As soon as your driver detects that the camera is not available it should do the following:

- Set this->deviceIsConnected=false
- Call this->disconnect()

The ADDriver base class has a parameter called ReadStatus and a ReadStatus record that can be processed periodically.  Several drivers check that parameter in readInt32 and call the readStatus() function.  We should add that check to ADDriver::readInt32() and add an ADDriver::readStatus method which might do nothing.  Derived classes can implement readStatus and do a check there to see if the device is available.  They may also detect that the device is disconnected at other times.  They should then do the 2 above steps.

If the driver detects that the camera comes back online it should do the following:
- Set deviceIsConnected=true
- Call asynPortDriver::connect()

Checking that the device is alive cannot be done in readStatus() if that is triggered by a periodic write of the ReadStatus record.  The reason is that if the driver is disconnected then records cannot write to it, asynManager will refuse the queueRequest.  So it would need to be an internal polling thread.

GDYendell:
> The way asyn connection works is that asynManager (or other code) calls pasynCommon->connect() in the driver. In the case of asynPortDriver/ADDriver that results in a call to ADDriver::connect(). The pasynCommon->connect() does the actual connection to the device (for example creating a socket in the case of the TCP driver). If it succeeds it calls pasynManager->exceptionConnect() to signal that it succeeded.

Ah OK that makes sense. So it is basically interrupting the call stack to verify if it actually managed to connect to the device and establish communications (however the driver happens to determine that) vs just succeeding to connect the port?

MarkRivers:
I have done the following:
- Renamed `ADDriver::deviceIsConnected` to `ADDriver::deviceIsReachable` to avoid multiple meanings of "connected".
- Added more comments in ADDriver::connect().
- Call asynPortDriver::connect() rather than asynNDArrayDriver::connect(), which does not currently exist.

GDYendell:
That's great, the new doc string makes it really clear what is going on. Thanks Mark!

PR #452, opened 2020-07-21
remove destroy

mrkraimer:
Starting with the next epics 7 release pvDatabase.cpp will no longer have any destroy methods.
With this change NDPluginPva will not build unless it's destroy methods are removed.

Note that for at least the last couple of epics7 releases pvDatabase.cpp has only had empty destroy methods.
Thus even with this change to NDPluginPva, ADCore will still build for at least the last couple of epics7 releases.

### DIFF
--- a/ADApp/pluginSrc/NDPluginPva.cpp
+++ b/ADApp/pluginSrc/NDPluginPva.cpp
@@ -43,7 +43,6 @@ public:
     virtual ~NTNDArrayRecord () {}
     static NTNDArrayRecordPtr create (string const & name);
     virtual bool init ();
-    virtual void destroy ();
     virtual void process () {}
     void update (NDArray *pArray);
 };
@@ -68,11 +67,6 @@ bool NTNDArrayRecord::init ()
     m_ntndArray = NTNDArray::wrap(getPVStructure());
     m_converter.reset(new NTNDArrayConverter(m_ntndArray));
     return true;
-}
-
-void NTNDArrayRecord::destroy()
-{
-    PVRecord::destroy();
 }
 
 void NTNDArrayRecord::update(NDArray *pArray)


MarkRivers:
We need to keep ADCore compatible with 3.14, 3.15, and 7.x.  So if needed we can use ifdefs for other versions.  Will this code break when using the V4 code that is not part of base?

mrkraimer:
I think that it should be compatible with any 3.14, 3.15, and 7.x release that has been generated in the last couple of years.
Let me check further.

With regard to:

``Will this code break when using the V4 code that is not part of base?``

Note that I have been working with  epics-base, all V4 modules, and areaDetector cloned from gitHub.

mrkraimer:
I have spent some time looking at compatibility with previous releases.
First note that the change to NTNDArrayRecord is to remove:

    void NTNDArrayRecord::destroy()
    {
        PVRecord::destroy();
    }

It was only declared because it modeled itself after PVRecord, which in the early days of V4 actually had real destroy code.
I looked and in Oct, 2017 pvDatabase pull request 28 stated

    All destroy methods except PVRecord::destroy either deprecated or removed.

And it should also have said that all deprecated methods are empty methods.
pvDatabaseCPP now  uses C++ shared pointers for memory management.

pvDatabaseCPP now has a pull request to remove all destroy methods.
It does this because I do not want any code that uses pvDatabaseCPP to implement  destroy methods that do something.
I just merged this pull request.

Removing destroy from PVRecord  will cause NTNDArrayRecord to have a compilation error.

    ../NDPluginPva.cpp: In member function ‘virtual void NTNDArrayRecord::destroy()’:
    ../NDPluginPva.cpp:75:15: error: ‘destroy’ is not a member of ‘epics::pvDatabase::PVRecord’
       75 |     PVRecord::destroy();



3.14, 3.15, and 3.16 releases are not an issue because they do not have any V4 modules.

Any epics 7 release newer than Oct,2017 will already have the version of pvDatabaseCPP with a PVRecord::destroy() that does nothing.

The next version of epics7 will no longer have this method
In my working area I am building areaDetector, cloned from github, with  NTNDArrayRecord::destroy() removed.

MarkRivers:
> 3.14, 3.15, and 3.16 releases are not an issue because they do not have any V4 modules.

I am not sure I understand this comment.  areaDetector supports pvAccess on 3.14, 3.15, and 3.16 using the separate V4 package.  We need to continue to support those versions of base with pvAccess.

mrkraimer:
'''areaDetector supports pvAccess on 3.14, 3.15, and 3.16 using the separate V4 package.'''

Look at the aps home page
https://epics.anl.gov

It appears to me that NO V4 release is mentioned.
It looks to me like V4 releases have been abandoned, since what was in the V4 releases is now part of epics7.

brunoseivam:
Hi Marty, a bit of context: at FRIB we are on 3.15 and deploy separate packages for V4 modules. I know this is perhaps not "official" but it has been the case since we don't have the bandwidth, currently, to upgrade our stuff to 7.0 entirely. We don't make much use of the V4 modules yet, though, so this breakage probably won't affect us, but might affect other sites.

Maybe the best way to avoid breakage is to wrap `PVRecord::destroy();` in an `ifdef` checking for EPICS base version?

mrkraimer:
Bruno or Mark.
Do you know what the idef should say?
If either of you want to make the changes on this branch just let me know.

Bruno,
When you do start making more use of V4 modules I strongly encourage you upgrade to a recent epics 7 release.
The early V4 modules had many issues that have been fixed in recent epics 7 releases.

MarkRivers:
Since each of the V4 modules is a separate repository in github.com/epics-base can't people with older versions of base (3.14, 3.15, 3.16) just update those modules?  Or are more recent versions of those modules tied to base 7 changes?

mrkraimer:
Bruno,

You said:

     we are on 3.15 and deploy separate packages for V4 modules

Does this mean that you are not using V4 releases.?
Do  You just clone the V4 modules you need from github?
ADCore requires  pvDatabaseCPP 
In this case if you have cloned pvDatabaseCPP any time after Oct 2017 then ADCore should build without any ifdefs  in NDPlugIn.

MarkRivers:
Does pvDatabaseCPP have macros in its header files that define what version it is?  If so then the #ifdef should just be based on those.

brunoseivam:
> Does this mean that you are not using V4 releases.?

Yes, unfortunately. We're on a fairly old version, but no one is really using it in production, so it's fine. We will have it fixed in the near future. I just wanted to provide some context about some installations out there

mrkraimer:
I maintain that no ifdefs are necessary.
Let me explain:
The code

    void NTNDArrayRecord::destroy()
    {
        PVRecord::destroy();
    }

Is and never was needed.
NTNDArrayRecord itself does not have any code for destroy. It  only asks PVRecord to execute it's destroy method
But even if the above code is not present, when the destroy is called for NTNDArrayRecord the PVRecord destroy will be called

A derived class calling base::method is useful when all of the following is true:
1) method is virtual
2) base has code for method
3) derived has code for method
4) when the derived method is called it also wants the base method code executed,

brunoseivam:
If `destroy` was never required by `PVRecord` to be present in its child classes, then your change won't do any harm indeed, and no ifdefs are necessary, I believe

oksanagit:
Hi Marty and all,
For the last couple days I was trying to understand why the code (pulled and checkout a specific branch) , which worked 2 month ago, now gives the error: 
 *../NDPluginPva.cpp: In member function ‘virtual void NTNDArrayRecord::destroy()’:
../NDPluginPva.cpp:75:15: error: ‘destroy’ is not a member of ‘epics::pvDatabase::PVRecord’
     PVRecord::destroy();
* 
I freshly clone and checkout a selected version using an ansible script, anb build. This worked two month ago, and now fails with the above error. 

When I build the base I checkout R7.0.3.1, however, it turned out that I end up with the master from every submodule in modules/. I realized that until now I was taking the master branches of every submodule every time I built the base regardless which tag I was checking out.  Yet, I though, it  should be possible to do some recursive checkouts.  But it looks like the tagging of the modules is inconsistent and has very little in common with epics-base tagging. For example,  pvaClient and PVAccess have *…., 7.0.0, 7.1.0, 7.1.1, 7.1.2*; pvaDatabase has * ... 4.2.1, 4.3.0, 4.4.0, 4.4.1, 4.4.2, 4.5.0, 4.5.1, 4.5.2*; pvDatabase has *... 7.1.1, 7.1.2, 7.1.3, 8.0.0, 8.0.1, 8.0.2, 8.0.3*; normativeTypes has *5.1.1, 5.1.2, 5.2.0, 5.2.1, 5.2.2, 6.0.0, 6.0.1*. It would be nice to have a way to figure out which modules' version are meant to be together, and in an ideal case to do the recursive checkout.  

What I ended up doing is looking at the commit dates, checking out pvaDatabase 4.5.2, rebuilding everything just in case and my ADCore got compiled.

MarkRivers:
@oksanagit you need to do `git submodule update`.  Here is an example.

See what version of base I have checked out:
```
corvette:local/epics-devel/base-7.0.4>git status
# HEAD detached at R7.0.4
...
```
So I have version R7.0.4 checked out.  See what version of each submodule I have:
```
corvette:local/epics-devel/base-7.0.4>git submodule status
-55038b731516f22584b7e1cede4ef0a05b8b647e .ci
 1250a3c236f0aa92e0b5bd73647fd71d8a09360d modules/normativeTypes (6.0.1)
 1641bd18b4243e5c281dc955c628c33925cab8c2 modules/pvAccess (7.1.1)
 d9b3f98e3571b77995a048af5457eb583673433a modules/pvData (8.0.2)
 3f5bfd067f7ffc414335a3e948be30c9be664231 modules/pvDatabase (4.5.2)
 cb13435d152a1247ac0a6dcc8ea6f5c7d8a5fa80 modules/pva2pva (1.2.3)
 bc9ac8422cb94a38c27385c3c6781aea30c2c8b8 modules/pvaClient (4.7.1)
```
That shows which version of each submodule I have.

Now checkout R7.0.3.1 instead:
```
corvette:local/epics-devel/base-7.0.4>git checkout R7.0.3.1
M       configure/CONFIG_SITE
M       configure/os/CONFIG_SITE.Common.vxWorksCommon
M       modules/pvAccess
M       modules/pvData
M       modules/pvDatabase
M       modules/pva2pva
M       modules/pvaClient
Previous HEAD position was 745d275... Update version numbers for release
HEAD is now at 34834df... Clear development flags for 7.0.3.1
```
Now update the submodules for 7.0.3.1.
```
corvette:local/epics-devel/base-7.0.4>git submodule update
Submodule path 'modules/pvAccess': checked out '47e6f61f049c273ee74dac7a6142ed2734b79140'
Submodule path 'modules/pvData': checked out '828506720c5204301f71078652ab799ef2d7b2b9'
Submodule path 'modules/pvDatabase': checked out '0c92f077498aee1b3d4340f60398e88e1054dee2'
Submodule path 'modules/pva2pva': checked out '3024f9fb0cabee6469ad2f1d685a1a38b8c43cbf'
Submodule path 'modules/pvaClient': checked out 'b724b726249f297f800ec0649da987036da68fce'
corvette:local/epics-devel/base-7.0.4>git submodule status
 1250a3c236f0aa92e0b5bd73647fd71d8a09360d modules/normativeTypes (6.0.1)
 47e6f61f049c273ee74dac7a6142ed2734b79140 modules/pvAccess (7.1.0)
 828506720c5204301f71078652ab799ef2d7b2b9 modules/pvData (8.0.1)
 0c92f077498aee1b3d4340f60398e88e1054dee2 modules/pvDatabase (4.5.1)
 3024f9fb0cabee6469ad2f1d685a1a38b8c43cbf modules/pva2pva (1.2.2)
 b724b726249f297f800ec0649da987036da68fce modules/pvaClient (4.7.0)
```

Now go back to 7.0.4.
```
corvette:local/epics-devel/base-7.0.4>git checkout R7.0.4
M       configure/CONFIG_SITE
M       configure/os/CONFIG_SITE.Common.vxWorksCommon
M       modules/pvAccess
M       modules/pvData
M       modules/pvDatabase
M       modules/pva2pva
M       modules/pvaClient
Previous HEAD position was 34834df... Clear development flags for 7.0.3.1
HEAD is now at 745d275... Update version numbers for release
```

Now update the submodules for 7.0.4.1.
```
corvette:local/epics-devel/base-7.0.4>git submodule update
Submodule path 'modules/pvAccess': checked out '1641bd18b4243e5c281dc955c628c33925cab8c2'
Submodule path 'modules/pvData': checked out 'd9b3f98e3571b77995a048af5457eb583673433a'
Submodule path 'modules/pvDatabase': checked out '3f5bfd067f7ffc414335a3e948be30c9be664231'
Submodule path 'modules/pva2pva': checked out 'cb13435d152a1247ac0a6dcc8ea6f5c7d8a5fa80'
Submodule path 'modules/pvaClient': checked out 'bc9ac8422cb94a38c27385c3c6781aea30c2c8b8'
corvette:local/epics-devel/base-7.0.4>git submodule status
-55038b731516f22584b7e1cede4ef0a05b8b647e .ci
 1250a3c236f0aa92e0b5bd73647fd71d8a09360d modules/normativeTypes (6.0.1)
 1641bd18b4243e5c281dc955c628c33925cab8c2 modules/pvAccess (7.1.1)
 d9b3f98e3571b77995a048af5457eb583673433a modules/pvData (8.0.2)
 3f5bfd067f7ffc414335a3e948be30c9be664231 modules/pvDatabase (4.5.2)
 cb13435d152a1247ac0a6dcc8ea6f5c7d8a5fa80 modules/pva2pva (1.2.3)
 bc9ac8422cb94a38c27385c3c6781aea30c2c8b8 modules/pvaClient (4.7.1)
```

oksanagit:
Thank you @MarkRivers! Scratch my complain, worked magic indeed.

PR #453, opened 2020-07-24
Fix/420 xmlfree

BenBradnick:
Resolves the LNK2001 unresolved symbol issue in #420 when building ADSupport dynamically and ADCore statically.

This was tested on Windows with EPICS BASE 3.14.12.7 and also tested on RHEL7 (although ADCore is dynamically in this case).


BenBradnick:
Hi @MarkRivers and @pheest would you be able to take a quick look at this PR? I believe it was what was discussed to do in the issue.

PR #454, opened 2020-07-29
Added cursor value computation to ND Stats plugin.

gregoryrd:
Provides access to image pixel value at cursor position.

This enhancement and pull request suggested by colleague and EPICS mentor, Matt Pearson.

Added forward links to CursorX and CursorY to assure cursor value is updated when cursor changes.

### DIFF
--- a/ADApp/Db/NDStats.template
+++ b/ADApp/Db/NDStats.template
@@ -642,6 +642,7 @@ record(longout, "$(P)$(R)CursorX")
    field(DTYP, "asynInt32")
    field(OUT,  "@asyn($(PORT),$(ADDR=0),$(TIMEOUT=1))CURSOR_X")
    field(VAL,  "256")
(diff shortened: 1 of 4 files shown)


MarkRivers:
@gregoryrd this looks good, thanks.

The change to the documentation was for the wrong file.  The HTML files in ADCore/documentation are obsolete and will be removed in a future release.  The current documentation is in docs/ADCore/*.rst.   I will make the change there.

I will also add this new PV to NDPluginStats.adl and update the other OPI files.

MarkRivers:
After merging this I discovered a minor issue.  The database is forcing the plugin to process whenever CursorX or Y changes.  That happens at iocInit, before the plugin has run, so it has no previous NDArray cached, and there were error messages about processing the plugin with no previous array.

I changed NDPluginDriver.cpp to make those messages be ASYN_TRACE_WARNING rather than ASYN_TRACE_ERROR, and to return asynSuccess rather than asynError.

PR #458, opened 2020-09-13
Create AWS S3 Interface for TIFF Files

stuwilkins:
@MarkRivers .. This is more of an exploratory "DEMO" PR, I wanted to test using the AWS S3 storage directly from AD that we talked about over e-mail. 

I have verified that this works at the moment writing to a MinIO object store at SDCC here at BNL but I have no reason it would not work directly to Amazon S3 object storage. 

MinIO is here:

https://min.io/

This basically gets rid of the requirement to have a file system to write images to. 

This is "super hacky" and is so brute force and ugly it might even be elegant. 

With some changes to the base NDFileTIFF it would only be a small shim over the files. 

Comments from all welcome... 

P.S. I realize that there is copied code in there. I didn't want to play too much with the NDFileTIFF at this stage. However, one could separate this out a little bit and make the NSFileTIFFS3 really simple.

### DIFF
--- a/ADApp/Db/Makefile
+++ b/ADApp/Db/Makefile
@@ -52,6 +52,7 @@ DB += NDAttrPlotAttr.template
 DB += NDAttrPlotData.template
 DB += NDAttrPlot.template
 DB += NDCodec.template
+DB += NDFileTIFFS3.template
 
 #----------------------------------------------------
 # If <anyname>.db template is not named <anyname>*.template add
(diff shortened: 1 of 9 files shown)


tacaswell:
https://github.com/constantinpape/z5 may also be of interest to target more cloud-native data formats.

stuwilkins:
Thanks @tacaswell This is just a test, not sure I am married to TIFF (as you know!). But, by using the AWS S3 async upload I can get line rate out of my office computer.........

Issue #460, opened 2020-09-22
Possible error in NDFileHDF5.cpp when writing HDF5 string attributes and datasets


MarkRivers:
The following error message was reported when using ADCore R3-9.
```
HDF5-DIAG: Error detected in HDF5 (1.10.1) thread 139745761855232:
  #000: ../H5T.c line 2153 in H5Tset_size(): size must be positive
    major: Invalid arguments to routine
    minor: Bad value
```

`H5Tset_size()` is called 3 times in NDFileHDF5.cpp:
```
corvette:ADCore/ADApp/pluginSrc>grep -n -U3 H5Tset_size NDFileHDF5.cpp
840-
841-  hdfdataspace     = H5Screate_simple(rank, &dims, NULL);
842-  hdfdatatype      = H5Tcopy(H5T_C_S1);
843:  hdfstatus        = H5Tset_size(hdfdatatype, str_value.size());
844-  hdfstatus        = H5Tset_strpad(hdfdatatype, H5T_STR_NULLTERM);
845-  hdfdset          = H5Dcreate2(element, name.c_str(), hdfdatatype, hdfdataspace, H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
846-  if (hdfdset < 0) {
--
1084-
1085-  hdfattrdataspace = H5Screate(H5S_SCALAR);
1086-  hdfdatatype      = H5Tcopy(H5T_C_S1);
1087:  hdfstatus        = H5Tset_size(hdfdatatype, str_attr_value.size());
1088-  hdfstatus        = H5Tset_strpad(hdfdatatype, H5T_STR_NULLTERM);
1089-  hdfattr          = H5Acreate2(element, attr_name.c_str(), hdfdatatype, hdfattrdataspace, H5P_DEFAULT, H5P_DEFAULT);
1090-  if (hdfattr < 0) {
--
3017-  if (strlen(attrStrValue) > 0){
3018-    hdfattrdataspace = H5Screate(H5S_SCALAR);
3019-    hdfdatatype      = H5Tcopy(H5T_C_S1);
3020:    H5Tset_size(hdfdatatype, strlen(attrStrValue));
3021-    H5Tset_strpad(hdfdatatype, H5T_STR_NULLTERM);
3022-    hdfattr          = H5Acreate2(element, attrName,
3023-                                  hdfdatatype, hdfattrdataspace,
```

[...]

PR #461, opened 2020-09-22
Include trailing null in size passed to H5Tset_size()

MarkRivers:
The size passed to H5Tset_size() is supposed to include the trailing null character.  The current code is not doing that.  This results in error messages from the HDF5 library if a zero length string is passed.  The file structure is also not correct, the string data should include the trailing null character.

This fixes issue #460.

### DIFF
--- a/ADApp/pluginSrc/NDFileHDF5.cpp
+++ b/ADApp/pluginSrc/NDFileHDF5.cpp
@@ -840,7 +840,7 @@ hid_t NDFileHDF5::writeH5dsetStr(hid_t element, const std::string &name, const s
 
   hdfdataspace     = H5Screate_simple(rank, &dims, NULL);
   hdfdatatype      = H5Tcopy(H5T_C_S1);
-  hdfstatus        = H5Tset_size(hdfdatatype, str_value.size());
+  hdfstatus        = H5Tset_size(hdfdatatype, str_value.size()+1);
   hdfstatus        = H5Tset_strpad(hdfdatatype, H5T_STR_NULLTERM);
   hdfdset          = H5Dcreate2(element, name.c_str(), hdfdatatype, hdfdataspace, H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
   if (hdfdset < 0) {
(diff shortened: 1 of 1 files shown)


MarkRivers:
I have verified that with this PR:
- There is no longer an error message when writing constant string dataset of length zero
- The size of the dataset now includes the trailing null character.

This is the output of h5dump with a non-zero length constant string dataset written before the fix:
```
(base) corvette:~/scratch>h5dump -p -d /detector/det_name hdf5_string_test_001.h5
HDF5 "hdf5_string_test_001.h5" {
DATASET "/detector/det_name" {
   DATATYPE  H5T_STRING {
      STRSIZE 18;
      STRPAD H5T_STR_NULLTERM;
      CSET H5T_CSET_ASCII;
      CTYPE H5T_C_S1;
   }
   DATASPACE  SIMPLE { ( 1 ) / ( 1 ) }
   STORAGE_LAYOUT {
      CONTIGUOUS
      SIZE 18
      OFFSET 2968
   }
   FILTERS {
      NONE
   }
   FILLVALUE {
      FILL_TIME H5D_FILL_TIME_IFSET
      VALUE  H5D_FILL_VALUE_DEFAULT
   }
   ALLOCATION_TIME {
      H5D_ALLOC_TIME_LATE
   }
   DATA {
   (0): "Simulated detector"
   }
}
}
```
This is the output of h5dump on that dataset after the fix:
```
(base) corvette:~/scratch>h5dump -p -d /detector/det_name hdf5_string_test_012.h5
HDF5 "hdf5_string_test_012.h5" {
DATASET "/detector/det_name" {
   DATATYPE  H5T_STRING {
      STRSIZE 19;
      STRPAD H5T_STR_NULLTERM;
      CSET H5T_CSET_ASCII;
      CTYPE H5T_C_S1;
   }
   DATASPACE  SIMPLE { ( 1 ) / ( 1 ) }
   STORAGE_LAYOUT {
      CONTIGUOUS
      SIZE 19
      OFFSET 2968
   }
   FILTERS {
      NONE
   }
   FILLVALUE {
      FILL_TIME H5D_FILL_TIME_IFSET
      VALUE  H5D_FILL_VALUE_DEFAULT
   }
   ALLOCATION_TIME {
      H5D_ALLOC_TIME_LATE
   }
   DATA {
   (0): "Simulated detector"
   }
}
}
```
In both cases the string is "Simulated detector" which is 18 bytes long.  Before the fix the SIZE reported by h5dump is 18, after the fix it is 19 because it includes the trailing null.  But the DATA printed is the same.

GDYendell (approved):
This seems reasonable to me.

PR #465, opened 2020-12-09
Add GitHubActions build with ci-scripts

### LINKED ISSUE #464: GitHub Actions
Has any consideration been made on moving CI to GitHub Actions? I have been working on the PVI PR and the travis builds were (and still are) incredibly slow to start, so I setup the Github Actions build and it seems to work really nicely. You can see them here: https://github.com/dls-controls/ADCore/actions

GDYendell:
This is towards getting #447 building on Github Actions to demonstrate how PVI works. I am still having a lot of trouble with travis builds and I am failing to get the build to use python 3.7, rather than 3.5.

I have tried to follow the docs and examples to set this up in the correct way, but there are a few outstanding points

- I had to add the gem install for coveralls-lcov directly to the build yaml. I don't know if there is a better way to do this with ci-scripts like the APT list of packages. Also, I had to add sudo to the command otherwise it fails with a permission error. The coveralls-lcov README specifically says not to use sudo, but maybe this is just a travis thing.
- The apt installs are now not coupled to the CONFIG_SITE generation as I they are being installed by the ci-scripts runner via the APT variable, so it will just install all of them always. Maybe we don't want to do this via ci-scripts and should just call the install-packages.sh as it was before
- I added the running and reporting of tests directly to the build yaml because I didn't see in the docs or in example repositories using ci-scripts how to do this with make tapfiles or make runtests.
- The build defines specific versions of asyn and ADSupport, but possibly it should build against master.
- I have only used one build definition while getting this working, but there could be a matrix of epics base versions, ubuntu versions, compilers etc. A windows build matrix would also be easy to add, but the pre-install and setup would have to be figured out

The builds are visible here: https://github.com/dls-controls/ADCore/actions

Closes #464


GDYendell:
Closing in favour of #483 which I have rebased my work for #447 on.

PR #477, opened 2022-02-07
Uncommented NDPluginPva plugin

antoninomiceli:
Uncommented NDPluginPva plugin. This is should be used routinely instead of the Image plugin.


MarkRivers:
This change is reasonable, because the default setting in areaDetector/configure/EXAMPLE_CONFIG_SITE.local is WITH_PVA=YES.
https://github.com/areaDetector/areaDetector/blob/b3879d7078b810f5a47b8b4682cd9a39c20f63e5/configure/EXAMPLE_CONFIG_SITE.local#L37

PR #483, opened 2022-07-01
Add github actions build w/ ci-scripts

mdavidsaver:
Adds GHA builds on Linux and Windows.  Some using OS packages, and some ADSupport.

The windows DLL job is commented out for now.  If fails to correctly find `xml2.lib` from ADSupport.

The `plugin-test` executable is only run with the OS package jobs as I'm unsure how to get boost test on Windows.

I think this is a superset of #465


mdavidsaver:
I think this PR is ready.  Now that @ralphlange has merged https://github.com/epics-base/ci-scripts/pull/70, I've updated the `.ci` submodule added by this PR to point to the main [ci-scripts](https://github.com/epics-base/ci-scripts) repository.

PR #490, opened 2023-05-12
Fix possible memory leak when using NDFileModeCapture in NDPluginFile

flowln:
Previously, when using the Capture mode, if you started a new capture without saving the previously captured frames to a file, the `pCapture` array holding the frames wouldn't be free'd properly, causing a memory leak. Another way of causing such leak was by changing the capture mode to stream without saving the frames to a file.

With this commit, when resetting `NDFileNumCaptured` before starting a new capture, we make sure to also empty the capture buffer if it is not empty yet, so that no buffer is leaked.

I used Valgrind to try to make sure there'd be no other way of reaching those `NDArray`s held in `pCapture` after overriding it or setting `NDFileModeCapture` to zero, but sorry if I missed something! :slightly_smiling_face:

### DIFF
--- a/ADApp/pluginSrc/NDPluginFile.cpp
+++ b/ADApp/pluginSrc/NDPluginFile.cpp
@@ -401,6 +401,18 @@ void NDPluginFile::freeCaptureBuffer(int numCapture)
     this->pCapture = NULL;
 }
 
+/** Resets the capture buffer so that it looks like there hasn't been any capture.
+ *  This ensures no memory is leaked when setting NDFileNumCaptured to zero. */
+void NDPluginFile::resetCaptureBuffer()
+{
+    int prevNumCaptures = 0;
+    getIntegerParam(NDFileNumCaptured, &prevNumCaptures);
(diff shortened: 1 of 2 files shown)


MarkRivers:
I had made similar fixes to this PR but with some cleanup and additional features.  These were done in these commits:

86e30c22232fe9a09179473aa7f96cc4cd94bef8
- Change capture buffer to use std::vector which is simpler and cleaner.
- Free capture buffer when capture or streaming start to prevent memory leak.
- Free capture buffer when the FreeCapture record is processed.  Allows manually freeing without starting another file save operation.

78781ddeacd8700b428950d5a815d230c619aa71
- Add FreeCapture record to NDFile.template

48e14a63b662d6c2c1b2cd3e860bbadd30342e08
- Add FreeCapture record to NDFileBase.adl

bb49af432db3a70963df3bd7adcca7786e3bf62d
-  Add new NDFileFreeCapture parameter to free the capture buffer in asynNDArrayDriver

PR #499, opened 2023-12-13
adding video compression (h264)

bsobhani:
ADCore changes for https://github.com/areaDetector/ADSupport/pull/44

### DIFF
--- a/ADApp/Db/NDCodec.template
+++ b/ADApp/Db/NDCodec.template
@@ -44,6 +44,8 @@ record(mbbo, "$(P)$(R)Compressor")
     field(THVL, "3")
     field(FRST, "BSLZ4")
     field(FRVL, "4")
+    field(FVST, "H264")
+    field(FVVL, "5")
     info(autosaveFields, "VAL")
 }
 
@@ -61,6 +63,8 @@ record(mbbi, "$(P)$(R)Compressor_RBV")
     field(THVL, "3")
     field(FRST, "BSLZ4")
     field(FRVL, "4")
+    field(FVST, "H264")
+    field(FVVL, "5")
     field(SCAN, "I/O Intr")
 }
 
@@ -217,3 +221,19 @@ record(waveform, "$(P)$(R)CodecError")
     field(FTVL, "CHAR")
     field(NELM, "256")
 }
+
+# Sets group of picture size. A value of 1 is effectively frame-level compression. Set to higher values for true video compression.
+record(ao, "$(P)$(R)GOPSize")
+{
+    field(DTYP, "asynInt32")
+    field(OUT,  "@asyn($(PORT),$(ADDR=0),$(TIMEOUT=1))GOP_SIZE")
+}
+
+# Sets the quantizer (qmin and qmax) to value written to PV.
+# One can effectively adjust quality this way, with 1 being the least lossy, and higher numbers lossier
+record(ao, "$(P)$(R)QMinMax")
+{
+    field(DTYP, "asynInt32")
+    field(OUT,  "@asyn($(PORT),$(ADDR=0),$(TIMEOUT=1))QMINMAX")
+}
+
--- a/ADApp/pluginSrc/NDPluginCodec.cpp
+++ b/ADApp/pluginSrc/NDPluginCodec.cpp
@@ -491,7 +491,6 @@ NDArray *compressLZ4(NDArray *input, NDCodecStatus_t *status, char *errorMessage
     return output;
 }
 
-
 NDArray *decompressLZ4(NDArray *input, NDCodecStatus_t *status, char *errorMessage)
 {
     // Sanity check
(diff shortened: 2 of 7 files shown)


ADApp/pluginSrc/NDPluginCodec.cpp
```diff
@@ -635,6 +634,96 @@ NDArray *decompressBSLZ4(NDArray *input, NDCodecStatus_t *status, char *errorMes
...
+}
+
+
+void set_gop_size(int value){
+    printf("unable to set gop size to %d\n", value);
+}
+
+void set_q_min_max(int value){
+    printf("unable to set q min max to %d\n", value);
+}
```
ericonr, line 721: These functions are implemented only for the case where `HAVE_VC` isn't defined.

ADApp/pluginSrc/NDPluginCodec.cpp
```diff
@@ -877,6 +983,8 @@ NDPluginCodec::NDPluginCodec(const char *portName, int queueSize, int blockingCa
     createParam(NDCodecBloscCLevelString,     asynParamInt32,   &NDCodecBloscCLevel);
     createParam(NDCodecBloscShuffleString,    asynParamInt32,   &NDCodecBloscShuffle);
     createParam(NDCodecBloscNumThreadsString, asynParamInt32,   &NDCodecBloscNumThreads);
+    createParam(NDCodecGOPSizeString, asynParamInt32,   &NDCodecGOPSize);
+    createParam(NDCodecQMinMaxString, asynParamInt32,   &NDCodecQMinMax);
```
ericonr, line 987: This isn't following the existing formatting.

ADApp/pluginSrc/NDPluginCodec.h
```diff
@@ -13,6 +13,8 @@
 #define NDCodecBloscCLevelString      "BLOSC_CLEVEL"     /* (int r/w) Blosc compression level */
 #define NDCodecBloscShuffleString     "BLOSC_SHUFFLE"    /* (bool r/w) Should Blosc apply shuffling? */
 #define NDCodecBloscNumThreadsString  "BLOSC_NUMTHREADS" /* (int r/w) Number of threads to be used by Blosc */
+#define NDCodecGOPSizeString  "GOP_SIZE" /* (int r/w) Group of pictures size for H264 */
+#define NDCodecQMinMaxString  "QMINMAX" /* (int r/w) Sets min and max values for quantizer for H264 */
```
ericonr, line 17: This isn't following the existing formatting either.

ericonr (commented):
The existing code has a lot of leftover whitespace changes, commented out code, and debugging printfs. Are you planning on cleaning these up after initial review?

ADApp/pluginSrc/NDPluginCodec.cpp
ericonr: I see now that `set_gop_size` and `set_q_min_max` are defined in the ADSupport functions. They are operating on global status; is it safe to operate on them during compression, for example? And this means we can't have more than one compression stream on the same IOC.

ADApp/Db/NDCodec.template
```diff
@@ -217,3 +221,19 @@ record(waveform, "$(P)$(R)CodecError")
...
+    field(OUT,  "@asyn($(PORT),$(ADDR=0),$(TIMEOUT=1))GOP_SIZE")
+}
+
+# Sets the quantizer (qmin and qmax) to value written to PV.
+# One can effectively adjust quality this way, with 1 being the least lossy, and higher numbers lossier
+record(ao, "$(P)$(R)QMinMax")
+{
+    field(DTYP, "asynInt32")
+    field(OUT,  "@asyn($(PORT),$(ADDR=0),$(TIMEOUT=1))QMINMAX")
+}
```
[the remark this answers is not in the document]
bsobhani: It is my understanding that QMin and QMax set a range for acceptable "Q" (which stands for quantizer and roughly corresponds to quality), and then the codec selects a "Q" within this range based on various other factors. When I was testing this I found it useful to have a single PV that changes the compression quality, rather than separate min/max that must be set every time. It would be more complete to have separate QMin/QMax PVs, and then perhaps a PV that controls a "QCenter" which shifts QMin and QMax by the same amount as a convenience, although there is a little bit of complexity in implementing "QCenter", and I tried to keep things simple for the initial implementation. I agree that in principle QMin and QMax should be exposed.

ADApp/pluginSrc/NDPluginCodec.cpp
```diff
@@ -635,6 +634,96 @@ NDArray *decompressBSLZ4(NDArray *input, NDCodecStatus_t *status, char *errorMes
...
+            *status = NDCODEC_ERROR;
+            //goto failure;
+            return NULL;
+    }
+
+
+
+    NDArrayInfo_t info;
+    input->getInfo(&info);
+    int outputSize = LZ4_compressBound((int)info.totalBytes);
```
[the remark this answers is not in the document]
bsobhani: LZ4_compressBound appears to be a simple function that calculates a number slightly larger than the value that it is given. I found this sufficient when testing this plugin, but I will think about if another calculation (or perhaps simply using info.totalBytes) would be more suitable.

ADApp/pluginSrc/NDPluginCodec.cpp
bsobhani: Yes I think a global variable is not ideal because I believe it is possible to have two separate cameras in a single IOC, and hence multiple instances of the plugin. I did not think of this. In practice, there is usually only camera per IOC. I don't think it is likely that the global variable is the reason for the unlock issues since I did my testing with only one instance of the plugin.

bsobhani:
Erico, thank you for the feedback. I have responded to some of your comments. When I have time I will take a more thorough look at the pull request and probably make some additional commits.

ADApp/Db/NDCodec.template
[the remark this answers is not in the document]
ericonr: Fair enough, that's a more complicated interface.

ADApp/pluginSrc/NDPluginCodec.cpp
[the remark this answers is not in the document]
ericonr: But `LZ4_compressBound` is also specific to LZ4 compression. That it works for video compression is more a coincidence than something that should be relied upon. That's why I think passing the buffer size to the compression function would at least allow us to catch issues with buffer size and avoid memory corruption. FFMpeg might also provide a function to help us determine the necessary buffer size, based on the configured bit-rate and quantization.

ADApp/pluginSrc/NDPluginCodec.cpp
ericonr: > I don't think it is likely that the global variable is the reason for the unlock issues since I did my testing with only one instance of the plugin.

Even with a single instance of the plugin, but without the additional locks you added, if you write into the QMinMax or GOPSize PVs while compression is happening, you will reset the codec state while the struct is being used. That's not safe. These functions should be implemented in NDPluginCodec, which can track the current encoding state, and block writes to these values if encoding is ongoing.

bsobhani:
I did some testing and was able to confirm a couple of things: 1) If I create a mutex to prevent video compression parameters from being written while compressing then I can safely leave H264_compress unlocked in ADCore. 2) I tried running two video streams in the same IOC and confirm there were issues - if I make each instance of NDPluginCodec have its own video compression context pointer and pass this pointer to the video compression functions then the issues seem to go away.

I have made changes locally that seem to resolve these things, but not ready to push yet. Next week I should have some more time to work on this and then I will push my changes.

bsobhani:
I have pushed changes which fix those two major issues. I plan to clean some things up and address some of the other smaller issues later.

Also, I seem to have done something to mutilate something in this pull request. All I was trying to do was fix a typo in my email address for a single commit. But in the process it seems to have changed the hashes of all the commits in the repo, which I think may be why it thinks there are 994 new commits. I will think about what to do about this.

PR #501, opened 2024-01-10
add BOB screen for ADSimDetector PVA image

prjemian:
This BOB screen (for CSS Studio phoebus) could support the PVA1:Image from any area detector.  The _more ..._ pop-up menu has the `simDetector.bob` screen as the first item.  (Can this be generalized?)

Call this screen with two macros:

macro  | description  | example
---- | ---- | ----
`P`      | IOC prefix   | `ad:`
`R`      | cam module   | `cam1: `

example:

sim_cam_image


prjemian:
Compare with existing, similar screen for CaQtDM: https://github.com/areaDetector/ADCore/blob/master/ADApp/op/ui/ad_cam_image.ui

PR #503, opened 2024-02-08
Support XML string and file path for NDAttributesFile

### LINKED ISSUE #502: Support XML string as well as file path in `NDAttributesFile`
In bluesky we would like to configure an areaDetector without having to have a shared filesystem with the detector. This is almost possible at the moment, but if we need to set `NDAttributesFile` then we need to pass a filesystem path which is read here:
https://github.com/areaDetector/ADCore/blob/9d2c32a210408cdc353331a4353a0b416d95444a/ADApp/ADSrc/asynNDArrayDriver.cpp#L329-L338

The pos plugin has a similar XML reading routine, but it supports the PV being either a blob of XML or a filename:
https://github.com/areaDetector/ADCore/blob/9d2c32a210408cdc353331a4353a0b416d95444a/ADApp/pluginSrc/NDPosPluginFileReader.cpp#L65-L69

I propose we do something similar for `NDAttributesFile`:
```cpp
if (fileName.find("<Attributes>") != std::string::npos){ 
    buffer = fileName;
} else {
    // The existing code
} 
```

Is this a reasonable thing to do? If so, then we can make a PR for it.

marcelldls:
closes #502 @coretl Tested against the example XML from ADCore/ADApp/ADSrc/asynNDArrayDriver.cpp

### DIFF
--- a/ADApp/ADSrc/asynNDArrayDriver.cpp
+++ b/ADApp/ADSrc/asynNDArrayDriver.cpp
@@ -326,16 +326,21 @@ asynStatus asynNDArrayDriver::readNDAttributesFile()
     this->pAttributeList->clear();
     if (fileName.length() == 0) return asynSuccess;
 
-    infile.open(fileName.c_str());
-    if (infile.fail()) {
-        asynPrint(pasynUserSelf, ASYN_TRACE_ERROR,
-            "%s::%s error opening file %s\n",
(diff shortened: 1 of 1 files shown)


coretl (approved):

coretl:
@MarkRivers are we ok to merge this?

MarkRivers:
Yes, doing it now.

PR #505, opened 2024-04-01
Remove erroneous variable type information.

ericonr:
Even though the shutter open and close delay times are related to the shutter mode, that isn't their type.


ericonr:
@MarkRivers

EmilioPeJu (approved):

PR #518, opened 2024-11-11
ADApp/pluginSrc/NDFileTIFF.cpp: restore dedicated handling of 32-bit …

mp49:
…integers. Having the 32-bit case fall through to the 64-bit case resulted in corrupted data most of the time, and this restores previous behavior without affecting the new 64-bit support.

### DIFF
--- a/ADApp/pluginSrc/NDFileTIFF.cpp
+++ b/ADApp/pluginSrc/NDFileTIFF.cpp
@@ -300,7 +300,11 @@ asynStatus NDFileTIFF::openFile(const char *fileName, NDFileOpenMode_t openMode,
             case NDAttrInt16:
             case NDAttrUInt16:
             case NDAttrInt32:
-            case NDAttrUInt32:
+            case NDAttrUInt32: {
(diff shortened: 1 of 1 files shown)


mp49:
fixes #519

MarkRivers (commented):
I think the original logic and this logic are both incorrect.

pAttribute->getValue() should be passed a datatype that matches the pointer it is passed.  pAttribute->getValue() does the type conversion from the actual type of the attribute to datatype passed to it.

So the calls should be
```
pAttribute->getValue(NDAttrInt32, &value.i32);
```
and
```
pAttribute->getValue(NDAttrInt64, &value.i64);
```
The existing code will probably work OK on little-endian machines, but not on big-endian machines.

My modified code is still not perfect because UInt32 and UInt64 will be printed as if they are signed.

MarkRivers:
I have create a new PR that fixes the issues with this one, and adds testing. #520

mp49:
Closing this because #520 is a better fix

PR #530, opened 2025-04-22
Fix asyn build in CI.

gustavosr8:


gustavosr8:
I forgot to mention that the best way to ensure compatibility with older platforms is to use containers, because this way we are not limited to the availability of github images, but this leads to a more complex development on CI.

PR #532, opened 2025-04-28
Convert NDPluginPva to use PVXS

### LINKED ISSUE #528: Convert NDPluginPva to use PVXS
In [epics-containers/epics-base](https://github.com/epics-containers/epics-base/blob/main/Dockerfile) we are building EPICS base with PVXS rather than pvAccess. This solves a number of issues for us, including the addition of `PVA_NAME_SERVERS` which means we can use TCP only allowing use of the network address translation needed in Kubernetes without exposing host network. (@gilesknap please correct me if I have got this wrong)

The only remaining component for us that is using the old pvAccess stack is the NDPluginPVA.

Would it be possible to convert it to use PVXS? Should it be an in place conversion, or should there be a Makefile switch to build an NDPluginPva against pvAccess or PVXS?

I noticed that #485 was commented as a stepping stone to PVXS by @mdavidsaver but is still marked as draft.

DLS is happy to do this work, but would need some guidance on the best approach, and someone to review the PR!

jsouter:
Closes #528 

Building with WITH_PVXS = YES (also require WITH_PVA = YES for time being) allows us to use pvxs, removing dependency on normativeTypesCPP, pvDataCPP, pvDatabaseCPP, pvAccessCPP.

WIP; currently testing against modified versions of ADSimDetector and pvaDriver, note that packages depending on this using PVXS will also need to be built with WITH_PVXS=YES in their config files. It may be that ultimately it would be better to move all the PVXS logic out of the separate *_pvxs.cpp files, but this was the simplest and most readable way to do it during development. 

There are a few TODOs I will address, so making this a draft PR for now.


ADApp/ntndArrayConverterSrc/ntndArrayConverter_pvxs.cpp
```diff
@@ -225,25 +225,25 @@ template <typename valueType>
 void NTNDArrayConverter::toAttribute (NDArray *dest, pvxs::Value attribute, NDAttrDataType_t dataType)
 {
     // TODO, can we make dataType a template parameter?
-    const char *name = attribute["name"].as<std::string>().c_str();
-    const char *desc = attribute["descriptor"].as<std::string>().c_str();
-    const char *source = attribute["source"].as<std::string>().c_str();
+    std::string name = attribute["name"].as<std::string>();
+    std::string desc = attribute["descriptor"].as<std::string>();
+    std::string source = attribute["source"].as<std::string>();
```
ericonr, line 230: Can't you use `auto` here?

ericonr (commented):
The first commit breaks the build, because it uses the copied (pvAccess) versions of the files, but links everything against PVXS. Maybe it would be best to do "copy files" -> "convert into PVXS" -> "hook up build system"?

Updating headers before implementation also breaks the build... I think it's easier to understand to have a single commit for implementation and header that can be verified at once, instead of splitting it.

What do you think?

ADApp/ntndArrayConverterSrc/Makefile
```diff
@@ -6,14 +6,26 @@ include $(TOP)/configure/CONFIG
...
 
 USR_CPPFLAGS += -DBUILDING_ntndArrayConverter_API
 
 LIB_LIBS              += ADBase
-LIB_LIBS              += pvData
-LIB_LIBS              += nt
+
+ifeq ($(WITH_PVXS),YES)
+	LIB_LIBS 			  += pvxs
+	LIB_LIBS 			  += Com
```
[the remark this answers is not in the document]
jsouter: Per this guide https://epics-base.github.io/pvxs/building.html
I have not tested without, I will see if it's required.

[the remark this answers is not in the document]
ericonr: This makefile already adds the common bits elsewhere, that's why it shouldn't be needed here. It's needed in the generic case, though, so the documentation has to include it.

ADApp/ntndArrayConverterSrc/ntndArrayConverter_pvxs.cpp
```diff
@@ -0,0 +1,447 @@
...
+    info.ndims = (int) dims.size();
+    info.nElements = 1;
+
+    for(int i = 0; i < info.ndims; ++i)
+    {
+        info.dims[i]    = dims[i]["size"].as<size_t>();
+        info.nElements *= info.dims[i];
+    }
+
+    info.codec = m_value["codec.name"].as<std::string>();
```
mdavidsaver, line 83: More relevant with pvaDriver, so fyi. here.  Be aware that `as<>()` will throw an exception if the `Value` is empty.  So this line will fail hard if `codec` is missing.

In situations where the structure definition is not provided by local code, eg. client code working with a server provided Value, I would recommend a more relaxed test of any field which are not essential.

There are two overloads of `Value::as()` to assist with this:

```c++
// if info.codec is already std::string
(void)m_value["codec.name"].as<std::string>(info.codec); // returns true if info.codec is updated

m_value["codec.name"].as<std::string>([&info](const std::string& v) {
    info.codec = v;
});
```

Or alternately

```c++
if(auto name = m_value["codec.name"])
    info.codec = name.as<std::string>();
```

In the first two cases both absence and type mis-match are handled softly.  In the last, only absense.

ADApp/ntndArrayConverterSrc/ntndArrayConverter_pvxs.cpp
```diff
@@ -0,0 +1,447 @@
...
+template <typename dataType>
+void NTNDArrayConverter::fromValue(NDArray *src) {
+    NDArrayInfo_t arrayInfo;
+    src->getInfo(&arrayInfo);
+
+    m_value["compressedSize"] = src->compressedSize;
+    m_value["uncompressedSize"] = arrayInfo.totalBytes;
+    std::string fieldName = m_fieldNameMap[typeid(dataType)];
+    auto arrayType = m_arrayTypeMap[typeid(dataType)];
+    const auto val = pvxs::detail::copyAs(
```
mdavidsaver, line 335: The `pvxs::detail` namespace is not part of the public API.  Please do not use members directly.

In the `dtype==stype` case, and in the short term, I think you would do better to `pvxs::allocArray()` and then `memcpy()`.

In the longer term, I think it would be interesting to see if this copy could be avoided by building a `pvxs::shared_array` around an `NDArray` reference.

ADApp/ntndArrayConverterSrc/ntndArrayConverter_pvxs.cpp
```diff
@@ -0,0 +1,447 @@
...
+    attrs.resize(src->pAttributeList->count());
+    while((attr = srcList->next(attr)))
+    {
+        NDAttrSource_t sourceType;
+        attr->getSourceInfo(&sourceType);
+        attrs[i] = m_value["attribute"].allocMember();
+        attrs[i].update("name", attr->getName());
+        attrs[i].update("descriptor", attr->getDescription());
+        attrs[i].update("source", attr->getSource());
+        attrs[i].update("sourceType", sourceType);
```
mdavidsaver, line 422: fyi. `Value::update()` is define so that it can be chained.

```c++
attrs[i] = m_value["attribute"].allocMember()
.update("name", attr->getName())
.update("descriptor", attr->getDescription())
.update("source", attr->getSource())
.update("sourceType", sourceType);
```

ADApp/pluginSrc/NDPluginPva_pvxs.cpp
```diff
@@ -0,0 +1,201 @@
...
+    
+    if(!pvRecord->init())
+        pvRecord.reset();
+    
+    return pvRecord;
+}
+
+bool NTNDArrayRecord::init ()
+{
+    m_pv = pvxs::server::SharedPV(pvxs::server::SharedPV::buildMailbox());
```
mdavidsaver, line 53: You may want to consider `buildReadonly()` instead, unless you also provide an `onPut()` handler.

ADApp/commonDriverMakefile
```diff
@@ -23,7 +23,11 @@ ifeq ($(WITH_QSRV),YES)
   PROD_LIBS += qsrv
 endif
 
-ifeq ($(WITH_PVA),YES)
+ifeq ($(WITH_PVXS),YES)
```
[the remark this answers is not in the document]
jsouter: Thanks for the review!
This is good to know, though maybe we still want to be explicit about selecting the PVXS backend if both PVA and PVXS are provided, considering that it changes the public signature of the NTNDArrayConverter constructor. What do you think @ericonr?
I suppose with how the build works right now that any other modules using ADCore as a dependency (like pvaDriver) will only be able to see the libraries for one implementation at a time even though it can see both versions of the headers, probably I should fix it so that we build both libraries.

ADApp/pluginSrc/NDPluginPva_pvxs.cpp
```diff
@@ -0,0 +1,201 @@
...
+bool NTNDArrayRecord::init ()
+{
+    m_pv = pvxs::server::SharedPV(pvxs::server::SharedPV::buildReadonly());
+    m_pv.open(m_value);
+    m_server = pvxs::server::Server::fromEnv();
+    m_server.addPV(m_name, m_pv);
+    m_server.start(); // start is not blocking
+    m_converter.reset(new NTNDArrayConverter(m_value));
+    return true;
+}
```
jsouter, line 60: To discuss: do we want to open a server per plugin instance?

ADApp/ntndArrayConverterSrc/Makefile
```diff
@@ -6,14 +6,25 @@ include $(TOP)/configure/CONFIG
...
 LIBRARY_IOC += ntndArrayConverter
 INC += ntndArrayConverterAPI.h
+ifeq ($(WITH_PVXS), YES)
+INC += ntndArrayConverter.h
+LIB_SRCS += ntndArrayConverter_pvxs.cpp
+USR_CPPFLAGS += -DHAVE_PVXS
+else
 INC += ntndArrayConverter.h
 LIB_SRCS += ntndArrayConverter.cpp
+endif
```
jsouter, line 16: Currently this builds only one implementation, should we be building both versions to allow users to switch between pvAccess and pvxs without rebuilding ADCore?

ADApp/pluginSrc/NDPluginPva_pvxs.cpp
mdavidsaver: You should not need to create a `Server` at all.  Instead use [`pvxs::ioc::server()`](https://epics-base.github.io/pvxs/ioc.html#adding-custom-pvs-to-server).

ADApp/commonDriverMakefile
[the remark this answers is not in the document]
mdavidsaver: The form I would encourage is:

```make
# Link QSRV
ifdef PVXS_1_3_0 # prefer v2
myioc_DBD += pvxsIoc.dbd
myioc_LIBS += pvxsIoc pvxs
else
ifdef EPICS_QSRV_MAJOR_VERSION # fallback to v1
myioc_LIBS += qsrv
myioc_LIBS += $(EPICS_BASE_PVA_CORE_LIBS)
myioc_DBD += PVAServerRegister.dbd
myioc_DBD += qsrv.dbd
endif
endif
```

Which detects either PVXS or pva2pva.

ADApp/pluginSrc/NDPluginPva_pvxs.cpp
```diff
@@ -0,0 +1,201 @@
...
+  *            This value should also be used for any other threads this object creates.
+  */
+NDPluginPva::NDPluginPva(const char *portName, int queueSize,
+        int blockingCallbacks, const char *NDArrayPort, int NDArrayAddr,
+        const char *pvName, int maxBuffers, size_t maxMemory, int priority, int stackSize)
+    /* Invoke the base class constructor */
+    : NDPluginDriver(portName, queueSize, blockingCallbacks,
+            NDArrayPort, NDArrayAddr, 1, maxBuffers, maxMemory, 0, 0,
+            0, 1, priority, stackSize, 1, true),
+            m_record(NTNDArrayRecord::create(pvName))
```
ericonr, line 140: Wrong indentation for `m_record` here

ericonr (changes requested):

ADApp/pluginSrc/NDPluginPva_pvxs.cpp
ericonr: @mdavidsaver how does that interact with my comment about `pvxsIoc.dbd` https://github.com/areaDetector/ADCore/pull/532#discussion_r2063797791 ? Is including `pvxsIoc.dbd` our responsibility, or the final integrator's when building the IOC?

And the documentation for the plugin should therefore state that the IOC needs to be built with QSRV2 and it shouldn't be disabled at runtime, right? Or should there be a fallback where if `pvxs::ioc::server()` throws, we create our own server?

Right now, wouldn't this throw anyway? The documentation states

> This Server instance is created during a registrar function, started by the initHookAfterCaServerRunning phase of iocInit().

But `NTNDArrayRecord::init` is called by `NTNDArrayRecord::create`, which is called by the `NDPluginPva` initializer, which should always happen before `iocInit`.

ADApp/ntndArrayConverterSrc/Makefile
ericonr: I feel like that's too much flexibility without clear differences between versions. IMO choosing between one or the other should be left to the EPICS "distributor", to simplify things.

But I am biased, because I didn't even think it was necessary to keep supporting the old PVA libraries, since PVXS can always be installed as a separate module.

mp49: I would prefer compatibility with the old libraries to be retained, at least until PVXS is made part of base. We have local areaDetector applications that make use of PVAccess/PVData libraries as well as the pvaDriver and pvaPlugin modules packaged with areaDetector. So we'd need to port our local applications to PVXS in order to update areaDetector, although I suppose using both at the same time is possible. Ideally we would port our local applications, and it doesn't seem difficult to do this. It's just we have at least 60+ IOC applications spread across 30 beamlines, and it would be a major upgrade.

jsouter: I think this may also be the first required use of C++11 in ADCore, before moving my test build to a containerised environment I was seeing linker errors trying to compile this against Diamond's released support modules (asyn and busy iirc) which were compiled using an older version of GCC - I don't think pvAccessCPP et al. use any C++11 features.

MarkRivers: @sveseli can you comment on this?  I seem to recall that PVAccess/PVData has some features that are lacking in PVXS with respect to serving multiple clients?

MarkRivers:
> I think this may also be the first required use of C++11 in ADCore (if a PVA plugin is needed), 

NDPluginBasePixel in ADCore requires C++11.

These are all of the modules in my areaDetector tree that have c++ flags:
```
(base) [epics@corvette areaDetector]$ find . -name Makefile -exec grep -H 'c++' {} \;
./ADPICam/PICamApp/src/Makefile:USR_CXXFLAGS_Linux += -std=c++1y
./ADCore/ADApp/pluginSrc/Makefile:NDPluginBadPixel_CXXFLAGS_Linux += -std=c++11
./ADCore/ADApp/pluginSrc/Makefile:NDPluginBadPixel_CXXFLAGS_Darwin += -std=c++11
./ADCore/ADApp/pluginTests/Makefile:  USR_CXXFLAGS_Linux += -std=c++11
./ADUVC/uvcApp/src/Makefile:USR_CPPFLAGS += -std=c++11
./ADGenICam/GenICamApp/src/Makefile:USR_CXXFLAGS_Linux += -std=c++11
./ADLambda/LambdaApp/src/Makefile:USR_CXXFLAGS += -std=c++17
./ADAravis/aravisApp/src/Makefile:USR_CXXFLAGS_Linux += -std=c++11
./ADPluginBar/barApp/barSrc/Makefile:CODE_CXXFLAGS=-std=c++11
./ADTimePix3/tpx3App/src/Makefile:USR_CPPFLAGS += -std=c++11
./ADSpinnaker/spinnakerApp/src/Makefile:USR_CXXFLAGS_Linux += -std=c++11 -Wno-unknown-pragmas
./ADSpinnaker/spinnakerApp/exampleSrc/Makefile:USR_CXXFLAGS_Linux += -std=c++11 -Wno-unknown-pragmas
```

ADApp/ntndArrayConverterSrc/Makefile
sveseli: The old plugin relies on PvDatabase server, which has the ability to distribute different data to different clients (e.g., image1 to client1, image2 to client2, etc.) if clients request that.  

I think would prefer that we do not modify the existing NDPluginPVA plugin, but instead that we simply add a new plugin (e.g. NDPluginPVXS).

MarkRivers: > I think I would prefer that we do not modify the existing NDPluginPVA plugin, but instead that we simply add a new plugin (e.g. NDPluginPVXS).

I agree.  Since the 2 PVA libraries have different features it makes sense to support both.  It also does not require rebuilding ADCore when switching between them.

sveseli: > I agree. Since the 2 PVA libraries have different features it makes sense to support both. It also does not require rebuilding ADCore when switching between them.

Yes. For example, APS DAQ system still uses old libraries, and if the support for those is dropped, there would have to be lots of work required on our end if we wanted to upgrade to the new ADCore version.

ericonr: > The old plugin relies on PvDatabase server, which has the ability to distribute different data to different clients (e.g., image1 to client1, image2 to client2, etc.) if clients request that.

Is this a PVXS limitation that's being worked on, or just something nice to have that people will eventually be able to live without?

I could imagine emulating it by having a fanout to multiple NDPluginPVXS instances, but it definitely doesn't have the same ergonomics...

jsouter: Okay, so it sounds like the best course for the time being is to provide an NDPluginPVA, NDPluginPVXS and the required NTNDArrayConverter and NTNDArrayConverterPVXS (or some similar name), and maybe we have completely separate header files for each so that we can choose to provide one, both or neither simultaneously based on CONFIG flags.

jsouter: I believe this works correctly with the current state of the branch, though I will clean up the git history so that rolling back of changes doesn't change too much of the git blame.

mdavidsaver: > The old plugin relies on PvDatabase server, which has the ability to distribute different data to different clients ...

The PVXS `SharedPV` object specifically does not do this.  Thus the "shared" part of its name.  PVXS also provides [a lower level API](https://epics-base.github.io/pvxs/source.html) on which `SharedPV` is built (also `p4p.gw`).  The [`spam.cpp`](https://github.com/epics-base/pvxs/blob/master/test/spam.cpp) test program is one example usage, providing a unique deluge to each subscriber to test per-subscription flow control.

jsouter (commented):
todo

jsouter:
Docs will need updating.
Also: need to make a dev build which contains both the old pva packages and PVXS, and check build/deployments work as expected, especially with regards to the pvxs ioc server

jwlodek:
I think you can probably just link to pvxsIOC / include the pvxsIOC.dbd in the commonDriverMakefile without needing do it into the library

MarkRivers (commented):
Looks good to me.  I will merge it.

MarkRivers:
I merged this PR and built pvxs with no errors.  I modified the following files in areaDetector/configure/

CONFIG_SITE.local to have these lines:
```
# EPICS PVXS can be used for NDPluginPva, and qsrv
WITH_PVXS = YES
```

RELEASE_LIBS.local to have this line
```
PVXS=$(SUPPORT)/pvxs
```

RELEASE_PRODS.local to have this line:
```
# PXVS is an optional package for NDPluginPVA and pvaServer
PVXS=$(SUPPORT)/pvxs
```

ADCore builds fine.  However, when I build ADSimDetector or any other driver I get these errors when linking the IOC.  I have added line breaks to make the output more legible.
```
/usr/bin/g++ -o simDetectorApp -Wl,-Bstatic -L/home/epics/devel/areaDetector-3-14/ADSimDetector/lib/linux-x86_64 
-L/home/epics/devel/areaDetector-3-14/ADSimDetector/iocs/simDetectorIOC/lib/linux-x86_64 
-L/corvette/home/epics/devel/areaDetector-3-14/ADCore/lib/linux-x86_64 
-L/corvette/home/epics/devel/areaDetector-3-14/ADSupport/lib/linux-x86_64 
-L/corvette/home/epics/devel/asyn-4-45/lib/linux-x86_64 
-L/corvette/home/epics/devel/autosave-5-11/lib/linux-x86_64 
-L/corvette/home/epics/devel/busy-1-7-3/lib/linux-x86_64 
-L/corvette/home/epics/devel/calc-3-7-4/lib/linux-x86_64 
-L/corvette/home/epics/devel/devIocStats-3-1-16/lib/linux-x86_64 
-L/corvette/home/epics/devel/pvxs/lib/linux-x86_64 
-L/corvette/home/epics/devel/seq-2-2-9/lib/linux-x86_64 
-L/corvette/home/epics/devel/sscan-2-11-5/lib/linux-x86_64 
-L/corvette/usr/local/epics-devel/base-7.0.9/lib/linux-x86_64 
-Wl,-rpath,/home/epics/devel/areaDetector-3-14/ADSimDetector/lib/linux-x86_64 
-Wl,-rpath,/home/epics/devel/areaDetector-3-14/ADSimDetector/iocs/simDetectorIOC/lib/linux-x86_64 
-Wl,-rpath,/corvette/home/epics/devel/areaDetector-3-14/ADCore/lib/linux-x86_64 
-Wl,-rpath,/corvette/home/epics/devel/areaDetector-3-14/ADSupport/lib/linux-x86_64 
-Wl,-rpath,/corvette/home/epics/devel/asyn-4-45/lib/linux-x86_64 
-Wl,-rpath,/corvette/home/epics/devel/autosave-5-11/lib/linux-x86_64 
-Wl,-rpath,/corvette/home/epics/devel/busy-1-7-3/lib/linux-x86_64 
-Wl,-rpath,/corvette/home/epics/devel/calc-3-7-4/lib/linux-x86_64 
-Wl,-rpath,/corvette/home/epics/devel/devIocStats-3-1-16/lib/linux-x86_64 
-Wl,-rpath,/corvette/home/epics/devel/pvxs/lib/linux-x86_64 
-Wl,-rpath,/corvette/home/epics/devel/seq-2-2-9/lib/linux-x86_64 
-Wl,-rpath,/corvette/home/epics/devel/sscan-2-11-5/lib/linux-x86_64 
-Wl,-rpath,/corvette/usr/local/epics-devel/base-7.0.9/lib/linux-x86_64          
-rdynamic -m64          simDetectorApp_registerRecordDeviceDriver.o simDetectorAppMain.o   -lsimDetector 
-lNDPlugin -lADBase -lqsrv -lntndArrayConverter -lnt -lpvDatabase -lpvAccessIOC -lpvAccessCA -lpvAccess 
-lpvData -lpvxs -lpvxsIoc -lntndArrayConverterPvxs -lnetCDF -lMagick++ -lcoders -lMagick -ljbig -ljp2 -lbzlib 
-lpng -lwebp -llcms -lttf -lwmf -lfilters -lnanohttp_stream -ltiff -lxml2 -lNeXus -lhdf5 -lhdf5_hl -lbitshuffle -lblosc 
-lszip -lzlib -ljpeg -lasyn -lautosave -lbusy -lcalc -ldevIocStats -lsscan -lseq -lpv -ldbRecStd -ldbCore -lca -lCom 
-Wl,-Bdynamic -lboost_system -lX11 -lXext -levent_core -levent_pthreads -lpthread   -lreadline -lm -lrt -ldl -lgcc
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink.o): in function `dbpvar':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../pvalink.cpp:186: multiple definition of `dbpvar'; /corvette/usr/local/epics-devel/base-7.0.9/lib/linux-x86_64/libqsrv.a(pvalink.o):/usr/local/epics-devel/base-7.0.9/modules/pva2pva/pdbApp/O.linux-x86_64/../pvalink.cpp:203: first defined here
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink_channel.o):/home/epics/devel/pvxs/ioc/O.linux-x86_64/../pvalink_channel.cpp:20: multiple definition of `pvaLinkNWorkers'; /corvette/usr/local/epics-devel/base-7.0.9/lib/linux-x86_64/libqsrv.a(pvalink_channel.o):/usr/local/epics-devel/base-7.0.9/modules/pva2pva/pdbApp/O.linux-x86_64/../pvalink_channel.cpp:8: first defined here
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink_jlif.o):/home/epics/devel/pvxs/ioc/O.linux-x86_64/../pvalink_jlif.cpp:310: multiple definition of `pvar_jlif_lsetPVA'; /corvette/usr/local/epics-devel/base-7.0.9/lib/linux-x86_64/libqsrv.a(pvalink.o):/usr/local/epics-devel/base-7.0.9/modules/pva2pva/pdbApp/O.linux-x86_64/../pvalink.cpp:349: first defined here
/bin/ld: /corvette/home/epics/devel/areaDetector-3-14/ADSupport/lib/linux-x86_64/libjp2.a(jas_stream.o): in function `jas_stream_tmpfile':
/home/epics/devel/areaDetector-3-14/ADSupport/supportApp/GraphicsMagickSrc/jp2/src/libjasper/O.linux-x86_64/../../../../../../supportApp/GraphicsMagickSrc/jp2/src/libjasper/base/jas_stream.c:368: warning: the use of `tmpnam' is dangerous, better use `mkstemp'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(iochooks.o): in function `pvxs::ioc::pvxsi()':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../iochooks.cpp:212: undefined reference to `pvxs::target_information(std::ostream&)'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink.o): in function `pvxs::client::Config::build() const':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../../src/pvxs/client.h:1076: undefined reference to `pvxs::client::Context::Context(pvxs::client::Config const&)'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink.o): in function `pvxs::ioc::linkGlobal_t::alloc()':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../pvalink.cpp:60: undefined reference to `pvxs::client::Context::~Context()'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink.o): in function `pvxs::client::Config::build() const':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../../src/pvxs/client.h:1076: undefined reference to `pvxs::client::Context::Context(pvxs::client::Config const&)'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink.o): in function `pvxs::ioc::linkGlobal_t::alloc()':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../pvalink.cpp:62: undefined reference to `pvxs::client::Context::~Context()'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink.o): in function `pvxs::ioc::testqsrvWaitForLinkConnected(link*, bool)':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../pvalink.cpp:134: undefined reference to `pvxs::client::Context::hurryUp()'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink_channel.o): in function `pvxs::ioc::linkGlobal_t::~linkGlobal_t()':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../pvalink_channel.cpp:51: undefined reference to `pvxs::client::Context::~Context()'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink_channel.o): in function `pvxs::client::detail::CommonBuilder<pvxs::client::PutBuilder, pvxs::client::detail::PRBase>::rawRequest(pvxs::Value const&)':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../../src/pvxs/client.h:688: undefined reference to `pvxs::client::detail::CommonBase::_rawRequest(pvxs::Value const&)'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink_channel.o): in function `pvxs::ioc::pvaLinkChannel::put(bool)':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../pvalink_channel.cpp:278: undefined reference to `pvxs::client::PutBuilder::exec()'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink_channel.o): in function `pvxs::client::detail::CommonBuilder<pvxs::client::PutBuilder, pvxs::client::detail::PRBase>::~CommonBuilder()':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../../src/pvxs/client.h:649: undefined reference to `pvxs::client::detail::PRBase::~PRBase()'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink_channel.o): in function `pvxs::client::detail::CommonBuilder<pvxs::client::MonitorBuilder, pvxs::client::detail::CommonBase>::rawRequest(pvxs::Value const&)':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../../src/pvxs/client.h:688: undefined reference to `pvxs::client::detail::CommonBase::_rawRequest(pvxs::Value const&)'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink_channel.o): in function `pvxs::ioc::pvaLinkChannel::open()':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../pvalink_channel.cpp:123: undefined reference to `pvxs::client::MonitorBuilder::exec()'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink_channel.o): in function `pvxs::client::detail::CommonBuilder<pvxs::client::MonitorBuilder, pvxs::client::detail::CommonBase>::~CommonBuilder()':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../../src/pvxs/client.h:649: undefined reference to `pvxs::client::detail::CommonBase::~CommonBase()'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink_channel.o): in function `pvxs::client::detail::CommonBuilder<pvxs::client::PutBuilder, pvxs::client::detail::PRBase>::~CommonBuilder()':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../../src/pvxs/client.h:649: undefined reference to `pvxs::client::detail::PRBase::~PRBase()'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink_channel.o): in function `pvxs::client::detail::CommonBuilder<pvxs::client::MonitorBuilder, pvxs::client::detail::CommonBase>::~CommonBuilder()':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../../src/pvxs/client.h:649: undefined reference to `pvxs::client::detail::CommonBase::~CommonBase()'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink_channel.o): in function `pvxs::ioc::linkGlobal_t::linkGlobal_t()':
/home/epics/devel/pvxs/ioc/O.linux-x86_64/../pvalink_channel.cpp:47: undefined reference to `pvxs::client::Context::~Context()'
/bin/ld: /corvette/home/epics/devel/pvxs/lib/linux-x86_64/libpvxsIoc.a(pvalink_channel.o):(.data.rel.local.DW.ref._ZTIN4pvxs6client10DisconnectE[DW.ref._ZTIN4pvxs6client10DisconnectE]+0x0): undefined reference to `typeinfo for pvxs::client::Disconnect'
collect2: error: ld returned 1 exit status
make[5]: *** [/corvette/usr/local/epics-devel/base-7.0.9/configure/RULES_BUILD:232: simDetectorApp] Error 1
make[5]: Leaving directory '/home/epics/devel/areaDetector-3-14/ADSimDetector/iocs/simDetectorIOC/simDetectorApp/src/O.linux-x86_64'
make[4]: *** [/corvette/usr/local/epics-devel/base-7.0.9/configure/RULES_ARCHS:58: install.linux-x86_64] Error 2
make[4]: Leaving directory '/home/epics/devel/areaDetector-3-14/ADSimDetector/iocs/simDetectorIOC/simDetectorApp/src'
make[3]: *** [/corvette/usr/local/epics-devel/base-7.0.9/configure/RULES_DIRS:85: src.install] Error 2
make[3]: Leaving directory '/home/epics/devel/areaDetector-3-14/ADSimDetector/iocs/simDetectorIOC/simDetectorApp'
make[2]: *** [/corvette/usr/local/epics-devel/base-7.0.9/configure/RULES_DIRS:85: simDetectorApp.install] Error 2
make[2]: Leaving directory '/home/epics/devel/areaDetector-3-14/ADSimDetector/iocs/simDetectorIOC'
make[1]: *** [/corvette/usr/local/epics-devel/base-7.0.9/configure/RULES_DIRS:85: simDetectorIOC.install] Error 2
make[1]: Leaving directory '/home/epics/devel/areaDetector-3-14/ADSimDetector/iocs'
make: *** [/corvette/usr/local/epics-devel/base-7.0.9/configure/RULES_DIRS:85: iocs.install] Error 2
```

Am I doing something wrong?

MarkRivers:
@mdavidsaver thanks for that.  I pulled the master branch, and the link errors are fixed.

I modified commonPlugins.cmd to add NDPluginPvxs and tried it with the simDetector.  This is the output from the startup script when running with gdb:
```
# Optional: load NDPluginPva and/or NDPluginPvxs plugins
#NDPvaConfigure("PVA1", $(QSIZE), 0, "$(PORT)", 0, $(PREFIX)Pva1:Image, 0, 0, 0)
NDPvxsConfigure("PVA1", 20, 0, "SIM1", 0, 13SIM1:Pva1:Image, 0, 0, 0)
[New Thread 0x7fff66615640 (LWP 1369994)]
[New Thread 0x7fff66514640 (LWP 1369995)]
[New Thread 0x7fff66413640 (LWP 1369996)]
2025/12/31 11:23:47.568 NDPluginDriver::deleteCallbackThreads timeout waiting for plugin thread 0 exit message
[Thread 0x7fff66413640 (LWP 1369996) exited]
[Thread 0x7fff66514640 (LWP 1369995) exited]
c++ error: No Instance
```
It appears that it fails to create the plugin.  It crashes later, but the backtrace is not very useful.

mdavidsaver:
Looking back to https://github.com/areaDetector/ADCore/pull/532#discussion_r2071696614, I recommend using some Makefile `ifdef` to detect the presence of PVXS and/or pva2pva when linking an executable.

```make
ifdef PVXS_1_3_0
# QSRV2 present
else
# no QSRV2, test for QSRV1
ifdef EPICS_QSRV_MAJOR_VERSION # fallback to v1
# QSRV1 present
endif
endif
```

mdavidsaver:
> ... when linking an executable.

This might be done in `commonDriverMakefile` with some logic like:

```make
# detect presence of QSRV1 or 2.  Prefer QSRV2.
ifdef PVXS_1_3_0
HAVE_PVXS = YES
else
ifdef EPICS_QSRV_MAJOR_VERSION
HAVE_PVA2PVA = YES
endif
endif

ifeq ($(HAVE_PVA2PVA)$(WITH_QSRV),YESYES)
# ... existing
endif

ifeq ($(HAVE_PVXS)$(WITH_PVXS),YESYES)
# ... existing
endif

jsouter:
Thanks both, apologies that I did not reply over the christmas period - I am no longer working at Diamond (or indeed with EPICS) so I was not set up to do any testing - but for clarification, I did most of my development in an epics-containers environment that Giles Knap is working on at DLS, which uses some ansible magic to handle configure/RELEASE logic across the included modules, so it's possible that that obscured some build errors for me.

PR #535, opened 2025-06-12
Fix inconsistencies in the documentation of some records

henriquesimoes:
While taking a look at the available PVs on the ADDriver, NDArray, and HDF5 plugin pages, I found some inconsistencies in the documentation, mostly typos in PV names, and some missing records.

For the missing records, I've written a quite generic description. Feel free to propose a better description line for each of them.

I haven't gone through all pages, so there might be other inconsistencies around.

As always, I recommend reviewing this patch series commit-wise.

### DIFF
--- a/docs/ADCore/NDArray.rst
+++ b/docs/ADCore/NDArray.rst
@@ -614,8 +614,8 @@ loading ADBase.template.
       The value of this record must be set with caution, because too large a value could
       use all memory on the system, resulting serious performance degradation.
     - POOL_NUM_PRE_ALLOC_BUFFERS
-    - $(P)$(R)NumPreAllocBuffers
-    - longout
+    - $(P)$(R)NumPreAllocBuffers, $(P)$(R)NumPreAllocBuffers_RBV
+    - longout, longin
   * - NDPoolPreAllocBuffers
     - asynInt32
     - r/w
--- a/docs/ADCore/NDArray.rst
+++ b/docs/ADCore/NDArray.rst
@@ -479,7 +479,7 @@ loading ADBase.template.
     - Compressed size of the array data in bytes. Only meaningful if NDCodec is not empty
       string.
     - COMPRESSED_SIZE
-    - $(P)$(R)Compressed_RBV
+    - $(P)$(R)CompressedSize_RBV
     - longin
   * -
     -
--- a/docs/ADCore/NDArray.rst
+++ b/docs/ADCore/NDArray.rst
@@ -465,13 +465,14 @@ loading ADBase.template.
   * - N.A.
     - N.A
     - r/o
-    - Size of each array dimension, extracted from the $(P)$(R)Dimensions and $(P)$(R)Dimensions_RBV
+    - Size of each array dimension K in [0..9], extracted from the $(P)$(R)Dimensions and $(P)$(R)Dimensions_RBV
       waveform records. Note that these are both longin record, i.e. readonly values using
       subarray records. In the future longout records may be added to write to the individual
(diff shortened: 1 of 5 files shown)


docs/ADCore/NDArray.rst
```diff
@@ -409,6 +409,41 @@ loading ADBase.template.
...
+    - longin
+  * - NDEpicsTSNsec
+    - asynInt32
+    - r/o
+    - Nanoseconds time stamp fraction of the EPICS time stamp. It should be
+      composed with $(P)$(R)EpicsTSSec_RBV to get the final time stamp with
+      nanosecond resolution.
+    - EPICS_TS_NSEC
+    - $(P)$(R)EpicsTSNsec_RBV
+    - longin
```
ericonr, line 446: You mentioned having to explain the meaning of each parameter in order to differentiate them. Could this documentation simply point to the (ported to rst version of the) areaDetectorTimeStampSupport page? See #537

docs/ADCore/NDFileHDF5.rst
```diff
@@ -751,45 +751,9 @@ Parameters and Records
...
-    - $(P)$(R)ExtraDimSize8, $(P)$(R)ExtraDimSize8_RBV
-    - longout, longin
-  * - asynInt32
-    - r/w
-    - Size of extra dimension 9
-    - HDF5_extraDimSize9
-    - $(P)$(R)ExtraDimSize9, $(P)$(R)ExtraDimSize9_RBV
+    - Size of the N'th extra dimension, for N in [3..9]
+    - HDF5_extraDimSize[N]
+    - $(P)$(R)ExtraDimSize[N], $(P)$(R)ExtraDimSize[N]_RBV
```
ericonr, line 786: Maybe there should be an example? It's not clear from this how the `[N]` substitution works. Is it `[3]` or `3`? I'd like to use `$N`, but I think that's confusing because `$(P)$(R)` is already in use here; maybe `<N>` would be better, to avoid confusion with the array notation from above?

ericonr (changes requested):
Nits:

> These records have been implemented a long time ago [1],

"were implemented a long time ago" (use simple past if specifying when something happened) or "have been implemented for a long time"

> reflecting the structured used in the source database template.

structure

> Document it under the metadata section as it is basically a metadata

"is basically metadata" (data is uncountable)

> the default value the readers should take when no value is
directly.

I think some word is missing

> Document it under the metadata section as it is basically a metadata
regarding the default value the readers should take when no value is
directly. This has been documented based on the corresponding HDF5
library function call that uses this value.
>
> [1] https://support.hdfgroup.org/documentation/hdf5/latest/group___d_c_p_l.html#ga4335bb45b35386daa837b4ff1b9cd4a4

The message is missing the `[1]` reference.

docs/ADCore/NDArray.rst
henriquesimoes: I find it reasonable to deduplicate the explanation about such things. However, I did find the [areaDetectorSupport page](http://htmlpreview.github.io/?https://github.com/areaDetector/ADCore/blob/74716326a0cbdf2907d90d4eeb18f426623edf54/documentation/areaDetectorTimeStampSupport.html) very IOC developer-oriented.

From my perspective, this documentation we are updating has a broader audience. End-users should be able to understand what a PV does without having to dig to much into details. Maybe that's a misconception of my part.

Well, if we are going to reference that page (or an adapted version of it), we still need to restore it. Are you planning to submit a PR for this? Or do you think we should include in the scope of this patch series?

ericonr (changes requested):

PR #542, opened 2025-07-23
Fix NDPluginStats dimension assignment

### LINKED ISSUE #541: NDStats plugin "Profiles" is broken for colour cameras
NDStats plugin uses Dim 0 and Dim 1 for the Profiles tool. For mono cameras this is fine, however for colour cameras Dim0 is used for the RGB channels, causing the Profiles tool to break.

Image

OCopping:
This PR aims to fix NDPluginStats and allow for using the "Profiles" tool for colour cameras. Currently the code manually assigns Dim 0 and Dim 1 to sizeX and sizeY respectively, but this is only the case for mono cameras. For RGB cameras, Dim 0 is used for RGB channels.

Instead I propose we use a the same logic as use in NDPluginROI, where `getInfo` determines the correct variables, based on the number of dims and the selected colour mode, and assigns it to an `NDArrayInfo_t` structure that is returned to the script for better assignment.

Fixes #541

### DIFF
--- a/ADApp/pluginSrc/NDPluginStats.cpp
+++ b/ADApp/pluginSrc/NDPluginStats.cpp
@@ -431,7 +431,7 @@ void NDPluginStats::processCallbacks(NDArray *pArray)
     double bgdCounts, avgBgd;
     NDArray *pBgdArray=NULL;
     int computeStatistics, computeCentroid, computeProfiles, computeHistogram;
-    size_t sizeX=0, sizeY=0;
+    size_t sizeX=0, sizeY=0, colorSize=0;
     int i;
     int itemp;
     NDArrayInfo arrayInfo;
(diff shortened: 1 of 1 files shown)


MarkRivers:
I don't think this is going to help with your goal of producing profiles for color cameras.  NDArrays for RGB color cameras have 3 dimensions, [NX, NY, 3].  Note that the documentation here:
https://areadetector.github.io/areaDetector/ADCore/NDPluginStats.html
says this:

>Each calculation can be independently enabled and disabled. Calculations 1 and 4 can be performed on arrays of any dimension. Calculations 2 and 3 are restricted to 2-D arrays.

Calculations 2 and 3 are centroids and profiles.  This is enforced here:
https://github.com/areaDetector/ADCore/blob/56eecf22d126451bd59719105c8d244792fcd192/ADApp/pluginSrc/NDPluginStats.cpp#L205
and here:
https://github.com/areaDetector/ADCore/blob/56eecf22d126451bd59719105c8d244792fcd192/ADApp/pluginSrc/NDPluginStats.cpp#L338

PR #545, opened 2025-07-31
Minor modifications to commonDriverMakefile to remove unused link for…

jwlodek:
… CV based plugins, add missing link for ffmpegServer


gustavosr8:
@jwlodek what do you think about splitting this commit in two parts?

The first could be `Remove unused link for CV based plugin` and the second `Add missing link for ffmpegServer`.

PR #550, opened 2025-08-22
Support variable character limit in NDAttributesFile

callumforrester:
Following https://github.com/areaDetector/ADCore/pull/503 the NDAttributesFile record can also directly contain an XML string, however the record is still limited to 256 characters, which may not be enough to accomodate a large blob of XML. This adds a macro, `XMLSIZE`, with a default of 256, so IOCs can be configured to accomodate larger blobs of XML.


jwlodek:
I think the macro should be used for the NELM field rather than FTVL?

callumforrester:
@MarkRivers thanks, done

PR #551, opened 2025-08-27
Avoid unnecessary indirection in xml2 usage

ronpandolfi:
My understanding is that xmlFree has always been part of the public api of libxml2. However, xmlGetGlobalState has since been removed in recent versions. This PR modifies xml2 usage according to these changes.

I made this observation from Rocky 10 and libxml2 2.12.5


MarkRivers:
Have you tested this using the version of xml2 in ADSupport?  You can just set XML2_EXTERNAL=NO in areaDetector/configure, and then it will use the version in ADSupport.

ronpandolfi:
@MarkRivers Thanks Mark! I do remember that tip and I haven't. I'm trying to adapt Jeung Han Lee's EPICS docker files for Rocky 10. He doesn't use ADSupport.

I might do that in the short term still, but I wanted to let you know so ADCore can get ahead of this update.

MarkRivers:
I just tested your changes on Linux and Windows using the version of xml2 in ADSupport.  They both compiled fine, so I am merging this.

Issue #557, opened 2026-01-02
too few arguments herr_t H5Oget_info_by_name3


mdavidsaver:
Fails to build with libhdf5 1.14 (circa Debian 13).  It looks like `H5Oget_info_by_name()` changed signature with libhdf5 1.12.

```
../HDF5FileReader.cpp: In function ‘herr_t file_info(hid_t, const char*, const H5L_info2_t*, void*)’:
../HDF5FileReader.cpp:15:25: error: too few arguments to function ‘herr_t H5Oget_info_by_name3(hid_t, const char*, H5O_info2_t*, unsigned int, hid_t)’
   15 |     H5Oget_info_by_name (loc_id, name, &infobuf, H5P_DEFAULT);
In file included from /usr/include/hdf5/serial/H5Apublic.h:21,
                 from /usr/include/hdf5/serial/hdf5.h:22,
                 from ../HDF5FileReader.h:15,
                 from ../HDF5FileReader.cpp:8:
/usr/include/hdf5/serial/H5Opublic.h:543:15: note: declared here
  543 | H5_DLL herr_t H5Oget_info_by_name3(hid_t loc_id, const char *name, H5O_info2_t *oinfo, unsigned fields,
      |               ^~~~~~~~~~~~~~~~~~~~
```

MarkRivers:
The fix for this has been merged. #558

PR #558, opened 2026-01-02
use HDF5 API circa 1.10

mdavidsaver:
Address #557 by selecting libhdf5 [API compatibility](https://support.hdfgroup.org/documentation/hdf5/latest/api-compat-macros.html) uniformly as of 1.10.


ericonr:
Could you include part of the issue description in the commit message? Something like

```
use HDF5 API circa 1.10

this fixes a build failure with libhdf5 1.14 (circa Debian 13).
H5Oget_info_by_name() changed signature with libhdf5 1.12 and added a
new argument.
```

mdavidsaver:
> Could you include part of the issue description in the commit message?

Done.

PR #563, opened 2026-01-09
Fix dll build

FreddieAkeroyd:
Set `HDF5_STATIC_BUILD` correctly for DLL builds. Also fixes the `xml2.lib` missing issue during link issue.


FreddieAkeroyd:
Though both `ADsupport` and `ADcore` include  `$(TOP)/configure/CONFIG_SITE.local` in their `CONFIG_SITE` adding to `CONFIG_SITE.local` did not work but adding to `CONFIG_SITE` did. Could  `$(TOP)` be not quite right in this case? needs more investigation possibly.

FreddieAkeroyd:
Hmm ... windows build failed with
```
Traceback (most recent call last):
  File "D:\a\ADCore\ADCore\.ci\cue.py", line 15, in <module>
    import distutils.util
ModuleNotFoundError: No module named 'distutils'
Error: Process completed with exit code 1.
```
I think ADCore needs a newer version of ci-scripts, i see the latest `cue.py` does not import distutils. I'll update PR to see if that fixes it - i'm guessing the windows build must use a more recent python than the linux builds, hence not seeing the issue there?

FreddieAkeroyd:
@mdavidsaver @MarkRivers please can you trigger a re-run of the GHA

MarkRivers:
It works!  I'm merging this.

PR #570, opened 2026-02-19
Destroyed pool fix: prevent SIGSEGV on IOC exit when pvAccess holds NDArrays after driver/pool destroyed

kgofron:
# Segmentation fault 

"fix: prevent SIGSEGV on IOC exit when pvAccess holds NDArrays after driver/pool destroyed" refers to Segmentation fault after ioc exits, when acquisition was performed (memory/pool allocated).

```
epics> auto_settings.sav: 2354 of 2354 PV's connected
ACQUIRE CHANGE: ADAcquire=1 (was 0), current ADStatus=0
PrvHst: Checking if TCP streaming should start - WritePrvHst=0
PrvHst: WritePrvHst is disabled (0) - TCP streaming not started
After acquireStart: ADStatus=1
ACQUIRE CHANGE: ADAcquire=0 (was 1), current ADStatus=1
PrvImg TCP connection closed by peer
PrvHst TCP disconnected
After acquireStop: ADStatus=0

epics> exit
PrvHst TCP disconnected
./st.cmd: line 5: 2343260 Segmentation fault      ../../bin/linux-x86_64/tpx3App st_base.cmd
```
### Fix applied to ADCore 3.14.0 master.

```
epics> exit
PrvHst TCP disconnected
```

### Problem
When an IOC exits (e.g. user types exit) after acquisition has run, the process can hit a SIGSEGV (signal 11). The crash is in NDArrayPool::release() (or equivalent use of the pool) after the detector driver and its NDArrayPool have already been destroyed.
**Cause**: Shutdown order: the detector driver destructor runs and deletes pNDArrayPoolPvt_. Later, the pvAccess ServerContext is torn down (atexit). Its MonitorElements still hold NDArray-derived data. The deleter used by ntndArrayConverter (freeNDArray) calls NDArray::release() on those arrays. By then the pool is gone, so release() runs against freed memory → SIGSEGV.
This has been seen with areaDetector IOCs (e.g. ADTimePix3) using ADCore 3.12.1 and 3.14.0. See issue areaDetector/ADTimePix3#5.

### Approach
Two parts:
**“Destroyed pool” registry**
* Before the driver deletes its pool, it registers the pool pointer in a static set.
* In NDArray::release(), we check that set using only the pool address (no dereference).
* If the pool was registered as destroying, we set pNDArrayPool = NULL and return without calling the pool.
So any late release() (from PVA or elsewhere) no-ops safely, even for NDArrays that are not the driver’s pArrays[] (e.g. copies handed to PVA).

**asynNDArrayDriver destructor**
* Store maxAddr in a member maxAddr_.
* In ~asynNDArrayDriver(): call NDArrayPool::registerDestroyingPool(pNDArrayPoolPvt_), null pNDArrayPool on each pArrays[i], then delete pNDArrayPoolPvt_.

### Changes

| File | Change |
| :---------------- | :------: |
| NDArray.h | Declare NDArrayPool::registerDestroyingPool(NDArrayPool*) and NDArrayPool::isPoolDestroyed(NDArrayPool*).|
| NDArrayPool.cpp | Implement both with a static std::set<NDArrayPool*> and a mutex. Pools are only ever added; the set is process-lifetime. |
| NDArray.cpp | At the start of NDArray::release(), if isPoolDestroyed(pNDArrayPool) then set pNDArrayPool = NULL and return ND_ERROR without calling the pool. |
| asynNDArrayDriver.h | Add private member int maxAddr_. |
| asynNDArrayDriver.cpp |  Constructor: initialize maxAddr_(maxAddr) (initializer order matches member declaration). Destructor: call registerDestroyingPool(pNDArrayPoolPvt_), then loop over pArrays[0..maxAddr_-1] and set pArrays[i]->pNDArrayPool = NULL, then delete pNDArrayPoolPvt_. |

[ADCore314_fix.md](https://github.com/user-attachments/files/25354999/ADCore314_fix.md)

### References

* areaDetector/ADTimePix3#5 – same issue and fix description (including diff-style notes).

### DIFF
--- a/ADApp/ADSrc/NDArrayPool.cpp
+++ b/ADApp/ADSrc/NDArrayPool.cpp
@@ -8,6 +8,7 @@
 
 #include <string.h>
 #include <stdlib.h>
+#include <set>
 #include <dbDefs.h>
 #include <stdint.h>
 
@@ -45,6 +46,33 @@ FreeFunc_t defaultFrameFree = free;
 
 volatile int eraseNDAttributes=0;
 extern "C" {epicsExportAddress(int, eraseNDAttributes);}
+
+/* Registry of pool addresses being destroyed so NDArray::release() can no-op (avoids SIGSEGV
+ * when pvAccess tears down after the driver). Only the address is used; the pointer is never
+ * dereferenced in isPoolDestroyed(). */
+static std::set<NDArrayPool *> *destroyedPools = NULL;
+static epicsMutexId destroyedPoolsMutex = NULL;
+
+void NDArrayPool::registerDestroyingPool(NDArrayPool *p)
+{
+    if (!p) return;
+    if (!destroyedPoolsMutex)
+        destroyedPoolsMutex = epicsMutexCreate();
+    if (!destroyedPools)
+        destroyedPools = new std::set<NDArrayPool *>();
+    epicsMutexLock(destroyedPoolsMutex);
+    destroyedPools->insert(p);
+    epicsMutexUnlock(destroyedPoolsMutex);
+}
+
+bool NDArrayPool::isPoolDestroyed(NDArrayPool *p)
+{
+    if (!p || !destroyedPoolsMutex || !destroyedPools) return false;
+    epicsMutexLock(destroyedPoolsMutex);
+    bool found = (destroyedPools->find(p) != destroyedPools->end());
+    epicsMutexUnlock(destroyedPoolsMutex);
+    return found;
+}
 
 /** NDArrayPool constructor
   * \param[in] pDriver Pointer to the asynNDArrayDriver that created this object.
--- a/ADApp/ADSrc/NDArray.cpp
+++ b/ADApp/ADSrc/NDArray.cpp
@@ -235,6 +235,11 @@ int NDArray::release()
     printf("%s: WARNING, no owner\n", functionName);
     return(ND_ERROR);
   }
+  /* No-op if pool was already destroyed (e.g. IOC exit: PVA tears down after driver). */
+  if (NDArrayPool::isPoolDestroyed(pNDArrayPool)) {
+    pNDArrayPool = NULL;
+    return(ND_ERROR);
+  }
   return(pNDArrayPool->release(this));
 }
 
--- a/ADApp/ADSrc/NDArray.h
+++ b/ADApp/ADSrc/NDArray.h
@@ -187,6 +187,13 @@ public:
     size_t       getMemorySize();
     int          getNumFree();
     void         emptyFreeList();
+
+    /** Register a pool as being destroyed so late NDArray::release() calls no-op (avoids SIGSEGV
+     * when pvAccess tears down MonitorElements after the driver and pool are deleted). */
+    static void registerDestroyingPool(NDArrayPool *p);
+    /** Check if the given pool address was registered as destroyed (pointer not dereferenced). */
+    static bool isPoolDestroyed(NDArrayPool *p);
+
     static void  setDefaultFrameMemoryFunctions(MallocFunc_t newMalloc,
                                                 FreeFunc_t newFree);
     virtual void* frameMalloc(size_t size);
--- a/ADApp/ADSrc/asynNDArrayDriver.cpp
+++ b/ADApp/ADSrc/asynNDArrayDriver.cpp
@@ -866,8 +866,8 @@ asynNDArrayDriver::asynNDArrayDriver(const char *portName, int maxAddr, int maxB
                      interfaceMask | asynInt32Mask | asynFloat64Mask | asynOctetMask | asynInt32ArrayMask | asynGenericPointerMask | asynDrvUserMask,
                      interruptMask | asynInt32Mask | asynFloat64Mask | asynOctetMask | asynInt32ArrayMask | asynGenericPointerMask,
                      asynFlags, autoConnect, priority, stackSize),
-      pNDArrayPool(NULL), queuedArrayCountMutex_(NULL), queuedArrayCount_(0),
-      queuedArrayUpdateRun_(true)
+      pNDArrayPool(NULL), maxAddr_(maxAddr), queuedArrayCountMutex_(NULL),
+      queuedArrayCount_(0), queuedArrayUpdateRun_(true)
 {
     char versionString[20];
     static const char *functionName = "asynNDArrayDriver";
@@ -1029,6 +1029,17 @@ asynNDArrayDriver::~asynNDArrayDriver()
     epicsEventSignal(queuedArrayEvent_);
     epicsEventWait(queuedArrayUpdateDone_);
 
+    /* Register pool as destroying so any late NDArray::release() (e.g. from pvAccess
+     * MonitorElement teardown) no-ops instead of touching the pool we are about to delete. */
+    NDArrayPool::registerDestroyingPool(this->pNDArrayPoolPvt_);
+
+    /* Null pNDArrayPool on driver-owned arrays as well (belt and braces). */
+    if (this->pArrays) {
+        for (int i = 0; i < this->maxAddr_; i++) {
+            if (this->pArrays[i])
+                this->pArrays[i]->pNDArrayPool = NULL;
+        }
+    }
     delete this->pNDArrayPoolPvt_;
     free(this->pArrays);
     delete this->pAttributeList;
--- a/ADApp/ADSrc/asynNDArrayDriver.h
+++ b/ADApp/ADSrc/asynNDArrayDriver.h
@@ -241,6 +241,7 @@ protected:
 
 private:
     asynStatus preAllocateBuffers();
+    int maxAddr_;
     NDArrayPool *pNDArrayPoolPvt_;
     epicsMutex *queuedArrayCountMutex_;
     epicsEventId queuedArrayEvent_;
--- a/iocBoot/EXAMPLE_commonPlugins.cmd
+++ b/iocBoot/EXAMPLE_commonPlugins.cmd
@@ -204,11 +204,11 @@ startPVAServer
 
 # Optional: load scan records
 #dbLoadRecords("$(SSCAN)/db/scan.db", "P=$(PREFIX),MAXPTS1=2000,MAXPTS2=200,MAXPTS3=20,MAXPTS4=10,MAXPTSH=10")
-#set_requestfile_path("$(SSCAN)/db")
+#set_requestfile_path("$(SSCAN)/sscanApp/Db")
 
 # Optional: load sseq record for acquisition sequence
 #dbLoadRecords("$(CALC)/db/sseqRecord.db", "P=$(PREFIX), S=AcquireSequence")
-#set_requestfile_path("$(CALC)/db")
+#set_requestfile_path("$(CALC)/calcApp/Db")
 
 # Optional: load devIocStats records (requires DEVIOCSTATS module)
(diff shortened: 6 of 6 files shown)


ericonr:
Hi Kaz! Have you seen #572 ? It would be nice to determine how these two interact, since you're currently registering your [exit handler manually](https://github.com/areaDetector/ADTimePix3/blob/5d85d2de16b372dcff691f6edfb019f166f3970d/tpx3App/src/ADTimePix.cpp#L5676), and you could take advantage of `ASYN_DESTRUCTIBLE` in the future

exzombie:
You need the latest asyn, and the ADCore from the PR that Erico linked above. Then, follow [these guidelines](https://github.com/areaDetector/ADCore/blob/91d002c3afa482c6da827f3b11b95706d1d5028b/docs/ADCore/destructible.rst)

kgofron:
## destructible-drivers

I compiled with destructible-drivers branch, but unfortunatly exit after acquisition still results in segmentation fault.

```
epics> exit
PrvHst TCP disconnected
./st.cmd: line 5: 2228300 Segmentation fault      ../../bin/linux-x86_64/tpx3App st_base.cmd
```
Perhaps I missed something.

* git remote 
  * https://github.com/exzombie/ADCore
```
kg1@lap133454:/epics/support2/areaDetector/ADCore/iocBoot$ git branch -a
* destructible-drivers
```
* 
```
kg1@lap133454:/epics/support2/areaDetector/ADCore/iocBoot$ git log
commit 91d002c3afa482c6da827f3b11b95706d1d5028b (HEAD -> destructible-drivers, origin/destructible-drivers)
Author: Jure Varlec <jure.varlec@cosylab.com>
Date:   Fri Feb 27 09:25:06 2026 +0000

    Add a release note about port shutdown
```

* asyn/asyn/asynDriver/asynDriver.h 
```
/* Version number names similar to those provide by base
 * These macros are always numeric */
#define ASYN_VERSION       4
#define ASYN_REVISION     45
#define ASYN_MODIFICATION  0
```

### Summary

The destructible-drivers branch (ADCore PR 572) fixes shutdown order (asyn calls shutdownPortDriver() and then deletes the driver), but it does not include the pool-safety fix from ADCore PR 570. So the crash you see is still the same: after the driver (and its NDArrayPool) are destroyed, pvAccess (PVA) can later call NDArray::release() on arrays that belonged to that pool → use-after-free → SIGSEGV.

So ADTimePix3 needs both:

* Destructible drivers (PR 572) 
* Destroyed-pool safety (PR 570) – this is what prevents the SIGSEGV when PVA (or anything else) calls release() after the pool is gone.


#### What to do

Apply the ADCore PR 570 (destroyed-pool) changes on top of your current destructible-drivers ADCore. That PR adds:

* **NDArrayPool**: registerDestroyingPool(NDArrayPool*) and isPoolDestroyed(NDArrayPool*) (e.g. static set + mutex).
* **NDArray::release()**: at the start, if isPoolDestroyed(pNDArrayPool) then set pNDArrayPool = NULL and return (do not call the pool).
* **asynNDArrayDriver destructor**: call registerDestroyingPool(pNDArrayPoolPvt_), null pNDArrayPool on each pArrays[i], then delete pNDArrayPoolPvt_.

Ways to get that into your tree:

* **Option A**: In your ADCore repo (on destructible-drivers), merge or cherry-pick the commits from PR 570 (the “fix: prevent SIGSEGV on IOC exit when pvAccess holds NDArrays after driver/pool destroyed” commit and any dependencies). Then rebuild ADCore and your IOC.
* **Option B** : Manually apply the same code changes from PR 570’s diff into your current ADCore (same files and logic as above), then rebuild.

After ADCore has both destructible-drivers and the pool-safety logic, exit should no longer segfault. If you paste your ADCore branch/commit and the PR 570 patch or link, I can outline exact merge/cherry-pick steps or a minimal patch for your tree.

kgofron:
## Stack trace:

“SIGSEGV on exit with ADCore master. Backtrace shows crash in NDArrayPool::release (ADCore) called from PVA teardown (freeNDArray → NDArray::release) after the driver and its pool are already destroyed. Full bt full and info sharedlibrary attached.”

Attached: The full bt full output (and optionally info sharedlibrary) from your tpx3_SIGSEGV.md file, or the relevant frames (#0–#2, #37, #49, #65–#66, #71–#72, #79–#81) to see the PVA → NDArray release → pool release path and the exit-handler order.

## GDB analysis – SIGSEGV on exit (ADCore master)
**Build**: ADTimePix3 IOC, run with ADCore current master (no PR 570, no destructible PR 572 in this run).
**Repro**: Acquisition run, then exit in the IOC shell.
**Crash**: SIGSEGV in NDArrayPool::release() at NDArrayPool.cpp:373 (onReleaseArray(pArray)).

### Where it crashes
**#0** – NDArrayPool::release(this=0x555556ccc870, pArray=0x7fff3c001a00) at NDArrayPool.cpp:373 So the fault is inside the pool’s release() (use-after-free on the pool or its internals).

### Call chain (who called into the pool)
**#81–79** – User types exit → epicsExit(0) → C library runs atexit handlers.
**#72–71** – One of those handlers destroys the pvAccess ServerContext (static/atexit cleanup).
**#70–65** – PVA tears down transports and channels, then ServerMonitorRequesterImpl::destroy.
**#57–48** – MonitorElementQueue and its MonitorElements are destroyed.
**#37–36** – MonitorElement destructor runs (PVA held a structure that contained array data).
**#30–11** – Destructors for PVStructure → PVValueArray<unsigned short> → shared_vector; the shared storage uses a custom deleter.
**#2** – freeNDArray<unsigned short>::operator() (ADCore ntndArrayConverter.cpp:44) – deleter for the PVA copy of the NDArray data.
**#1** – NDArray::release() – called from that deleter.
**#0** – NDArrayPool::release() – crash.

So: PVA is shutting down (atexit), destroying MonitorElements that still hold NDArray-backed data. Their deleter calls NDArray::release(), which calls NDArrayPool::release() on a pool that no longer exists.

### Why the pool is gone
Shutdown order is:

* Earlier in exit, the ADTimePix driver is destroyed (either by epicsAtExit(exitCallbackC) or by asyn destructible teardown). That runs ~asynNDArrayDriver, which deletes the NDArrayPool (pNDArrayPoolPvt_). So the pool at 0x555556ccc870 is freed.
* Later, the pvAccess ServerContext is destroyed (another atexit). Its MonitorElements still hold NDArrays (or views) that point at that same pool. When those are freed, freeNDArray → NDArray::release() → NDArrayPool::release() runs on the already deleted pool → use-after-free → SIGSEGV.
So the trace matches the “destroyed pool” scenario: driver (and pool) destroyed first, PVA tears down later and calls release() on that pool.

### Conclusion
**Root cause:** Use-after-free in ADCore: NDArray::release() is called from PVA’s deleter after the driver’s NDArrayPool has already been destroyed. The crash is in NDArrayPool::release (ADCore), not in ADTimePix3.
**Fix:** This is exactly what ADCore PR 570 (destroyed-pool safety) addresses: make NDArray::release() (and optionally the pool) safe when the pool has already been marked destroyed / unregistered, so PVA’s late release no-ops instead of touching freed pool memory.

[tpx3_SIGSEGV.md](https://github.com/user-attachments/files/25981453/tpx3_SIGSEGV.md)
[tpx3_SIGSEGV_frames.md](https://github.com/user-attachments/files/25981456/tpx3_SIGSEGV_frames.md)

# Why fix ADCore

* ADCore / PVA lifetime bug, and ADTimePix3 can at best work around it, not truly fix it.

* The crash is in:
  * NDArrayPool::release() → NDArray::release() → freeNDArray (from ntndArrayConverter)
  * Called while pvAccess ServerContext is being destroyed at atexit.
* By that time, the driver’s NDArrayPool has already been deleted (driver shutdown/destructor ran earlier), but PVA still holds NDArrays whose deleter calls NDArray::release().

So the use‑after‑free is between PVA’s lifetime and ADCore’s pool, not inside ADTimePix3.


## What ADTimePix3 can and cannot do

* **What it cannot do:**
  * It **cannot intercept** NDArray::release() or freeNDArray calls coming from PVA. Once an NDArray is handed to PVA, only ADCore’s NDArray/NDArrayPool layer can protect against late release() calls (this is exactly what [PR 570](https://github.com/areaDetector/ADCore/pull/570) does with the “destroyed pool” registry and early return in NDArray::release()).

* **What it can do as a workaround (no ADCore change):**
  * **Do not destroy the driver/pool at exit.**
     * Don’t pass ASYN_DESTRUCTIBLE in the IOC (use ADTimePixConfigWithFlags(..., 0)), and
     * Don’t register epicsAtExit (or guard it behind a macro).
  * Then, when PVA shuts down at atexit, the NDArrayPool is still alive; NDArray::release() works and the process exits cleanly.
  * Trade‑off: the driver, its threads, and the pool leak until process exit (which is usually fine for IOC shutdown).

* If **we are willing to leak on exit:** yes, ADTimePix3 can avoid the crash by **not** being destructible (no ASYN_DESTRUCTIBLE, no epicsAtExit deletion) when using an unpatched ADCore.
* **If you want a correct, leak‑free, future‑proof fix: ADCore must be changed** (PR 570 or equivalent). That’s the only place that can safely handle “PVA calls NDArray::release() after the pool is gone.”



*

exzombie:
Strange, I've been working with pvxs lately and I thought that the server was stopped before the epicsAtExit hooks are run. Is it possible that the NDPluginPvxs is still holding SharedPVs referencing NDArrays? Is it destroyed at all? In that case, it seems fixing that would be the right thing to do.

Then again I don't know what happens when NDPluginPva is used. And I recently learned that NDPlugin keeps a reference to the last NDArray, which keeps hanging around unless the NDPlugin is destroyed. So, maybe fixing this in the pool is not a bad thing. The reason why it seems a bit icky to me is that it looks like a band aid for a deeper ownership problem. But I also realize that changing how ownership works when the pool class is used all over the place and referenced using raw pointers is probably not going to happen.

Sorry, I didn't have any time for this in the past few weeks, so I'm just throwing ideas out there. Ignore if stupid :)

jwlodek:
I suppose making an ADCore R4-* that leverages c++ smart pointers for this and other things (thus requiring c++ 11 and adjustments to all drivers) would probably require too much work to be worthwhile

kgofron:
## Summary
A reasonable engineering path: land 570-style safety in ADCore (or equivalent), combine with 572 where you use destructible drivers, and separately audit PVA/pvxs plugins for dangling NDArray use on exit—without betting on an ADCore R4 refcount rewrite.

ADApp/ADSrc/NDArrayPool.cpp
```diff
@@ -46,6 +47,33 @@ FreeFunc_t defaultFrameFree = free;
...
+static std::set<NDArrayPool *> *destroyedPools = NULL;
+static epicsMutexId destroyedPoolsMutex = NULL;
+
+void NDArrayPool::registerDestroyingPool(NDArrayPool *p)
+{
+    if (!p) return;
+    if (!destroyedPoolsMutex)
+        destroyedPoolsMutex = epicsMutexCreate();
+    if (!destroyedPools)
+        destroyedPools = new std::set<NDArrayPool *>();
```
ericonr, line 62: This is not the most thread-safe static initialization. If static objects are needed, it would definitely be best to use proper C++ classes so they could take advantage of the language guarantees around static objects inside of functions.

ericonr (commented):
Out of curiosity, couldn't we simply move the pool handling to a `std::shared_ptr`? If every NDArray holds a reference to it, then it will only be destroyed when the last NDArray is.

That seems to me like it would be a simpler implementation, with less global state.

Issue #57, opened 2015-02-19
time to delete file: NexusTemplate.xml


prjemian:
in ADCore/iocs/simDetectorIOC/iocBoot/iocSimDetector/
the file NexusTemplate.xml is legacy from AD 1.x and can now be deleted

MarkRivers:
Please explain.  AD 2-x still supports the NDFileNexus plugin, and we still want to have an example XML file of how to configure it.

Mark

prjemian:
agreed

PR #572, opened 2026-02-25
Destructible drivers

exzombie:
This is a followup to https://github.com/epics-modules/asyn/pull/171. It adds
support to the `asynNDArrayDriver` and `NDPluginDriver` base classes, and
`NDPluginROI`. No changes to `ADDriver` are needed.

The core idea is that destructability is opt-in to maintain compatiblity with
derived classes that have not been updated yet, and may only be opted in by the
leaf class. And because one doesn't know whether any particular class is a leaf
class as it's always possible to create a derived class, destructability should
in practice be declared in the iocsh command that instantiates a driver. This is
how it was done for `NDPluginROI`.

To maintain compatibility with older versions of asyn, use of new symbols is
gated by `#ifdef ASYN_DESTRUCTIBLE` where needed. It needs to be pointed out
that `virtual void shutdownPortDriver()` does not need to be gated because
there's no issue with adding it even with older versions of asyn as long as you
don't use the C++11 `override` keyword. Only the call into the base
`asynPortDriver::shutdownPortDriver()` needs to be gated.

Note that I did not change the existing destruction in any way, only moved
things around. I seems to me that `~asynNDArrayDriver()` is incomplete, but that
is a job for another time.

The recipe for making a driver destructible is as follows:

* If the driver does **not** need to be compatible with older versions of asyn
  (i.e., R44 or earlier), and is not (yet) further subclassed:

  1. Pass `ASYN_DESTRUCTIBLE` flag to the base constructor (i.e., `NDPluginDriver`
     or `ADDriver`).
  2. Override `shutdownPortDriver()` and put there code that needs to be
     executed with the driver intact. This is a good place to stop threads, for
     example. `shutdownPortDriver()` is a virtual function, so don't forget to
     call the base implementation.
  3. Implement the destructor. Do the cleanup as best you can. The destructors
     are actually called now, which didn't use to be the case; make use of it.
     Note that `shutdownPortDriver()` will only be called when the IOC shuts
     down, so, if the driver could be used outside an IOC (e.g. in unit tests),
     you should call `shutdownPortDriver()` from the destructor. To determine if
     it has already been run, call `shutdownNeeded()` which will return `false`
     if the shutdown has already happened.
 
 * If the driver needs to be backwards compatible:
 
  1. The constructor should not add `ASYN_DESTRUCTIBLE` to the flags. Instead,
     it should accept flags as an argument, and `ASYN_DESTRUCTIBLE` should be
     put there by the iocsh command that instantiates the driver. This allows
     the driver to be subclassed when the derived class is not destructible.
     Use of `ASYN_DESTRUCTIBLE` needs to be gated with an `#ifdef`.
  2. Override `shutdownPortDriver()` and put there code that needs to run on
     IOC shutdown. `shutdownPortDriver()` is a virtual function, so don't
     forget to call the base implementation.
  3. Implement the destructor. Note that newer versions of asyn will call it,
     but older versions **will not**. So, use it to release memory and such,
     but anything that needs to happen in order to disconnect from the device
     must go into `shutdownPortDriver()`.

     Note that `shutdownPortDriver()` will only be called when the IOC shuts
     down, so, if the driver could be used outside an IOC (e.g. in unit tests),
     you should call `shutdownPortDriver()` from the destructor. To determine if
     it has already been run, you will need to set a variable in
     `shudownPortDriver()` yourself because the `shutdownNeeded()` function is only
     available in newer asyn versions.

### DIFF
--- a/ADApp/ADSrc/asynNDArrayDriver.cpp
+++ b/ADApp/ADSrc/asynNDArrayDriver.cpp
@@ -1022,12 +1022,29 @@ asynNDArrayDriver::asynNDArrayDriver(const char *portName, int maxAddr, int maxB
 
 }
 
+// When the driver subclass is destructible, this function will be called at IOC
+// shutdown.
+void asynNDArrayDriver::shutdownPortDriver() {
+    asynPrint(pasynUserSelf, ASYN_TRACE_FLOW, "%s shutting down\n", driverName);
 
-asynNDArrayDriver::~asynNDArrayDriver()
-{
     queuedArrayUpdateRun_ = false;
     epicsEventSignal(queuedArrayEvent_);
     epicsEventWait(queuedArrayUpdateDone_);
+
+#ifdef ASYN_DESTRUCTIBLE
+    asynPortDriver::shutdownPortDriver();
+#endif
+}
+
+asynNDArrayDriver::~asynNDArrayDriver()
+{
+    // If the driver subclass is not destructible, or asyn is old, or we are not
+    // in an IOC (e.g. unit tests), we need to call shutdown ourselves.
+    // On newer versions of asyn, we could check with needsShutdown() to see if
+    // shutdown has already been done, be we don't want to rely on that.
+    if (queuedArrayUpdateRun_) {
+        shutdownPortDriver();
+    }
 
     delete this->pNDArrayPoolPvt_;
     free(this->pArrays);
--- a/ADApp/ADSrc/asynNDArrayDriver.h
+++ b/ADApp/ADSrc/asynNDArrayDriver.h
@@ -148,6 +148,7 @@ public:
     virtual asynStatus setIntegerParam(int index, int value);
     virtual asynStatus setIntegerParam(int list, int index, int value);
     virtual void report(FILE *fp, int details);
+    virtual void shutdownPortDriver();
 
     /* These are the methods that are new to this class */
     virtual asynStatus createFilePath(const char *path, int pathDepth);
--- a/ADApp/pluginSrc/NDPluginDriver.cpp
+++ b/ADApp/pluginSrc/NDPluginDriver.cpp
@@ -169,17 +169,33 @@ NDPluginDriver::NDPluginDriver(const char *portName, int queueSize, int blocking
     unlock();
 }
 
+// When the plugin subclass is destructible, this function will be called at IOC
+// shutdown.
+void NDPluginDriver::shutdownPortDriver() {
+    asynPrint(pasynUserSelf, ASYN_TRACE_FLOW, "%s shutting down\n", driverName);
+
+    // Most methods in NDPluginDriver expect to be called with the asynPortDriver mutex locked.
+    // Shutdown does not, the mutex should be unlocked before it is called.
+    // We lock the mutex because deleteCallbackThreads expects it to be held, but then
+    // unlocked it because the mutex is deleted in the asynPortDriver destructor and the
+    // mutex must be unlocked before deleting it.
+    this->lock();
+    deleteCallbackThreads();
+    this->unlock();
+
+    asynNDArrayDriver::shutdownPortDriver();
+}
+
 NDPluginDriver::~NDPluginDriver()
 {
-  // Most methods in NDPluginDriver expect to be called with the asynPortDriver mutex locked.
-  // The destructor does not, the mutex should be unlocked before calling the destructor.
-  // We lock the mutex because deleteCallbackThreads expects it to be held, but then
-  // unlocked it because the mutex is deleted in the asynPortDriver destructor and the
-  // mutex must be unlocked before deleting it.
-  delete throttler_;
-  this->lock();
-  deleteCallbackThreads();
-  this->unlock();
+    // If the driver subclass is not destructible, or asyn is old, or we are not
+    // in an IOC (e.g. unit tests), we need to call shutdown ourselves.
+    // On newer versions of asyn, we could check with needsShutdown() to see if
+    // shutdown has already been done, be we don't want to rely on that.
+    if (pToThreadMsgQ_)
+        shutdownPortDriver();
+
+    delete throttler_;
 }
 
 /** Method that is normally called at the beginning of the processCallbacks
@@ -1011,7 +1027,7 @@ asynStatus NDPluginDriver::createCallbackThreads()
 }
 
 /** Deletes the plugin threads.
-  * This method is called from the destructor and whenever QueueSize or NumThreads is changed. */
+  * This method is called on shutdown and whenever QueueSize or NumThreads is changed. */
 asynStatus NDPluginDriver::deleteCallbackThreads()
 {
     ToThreadMessage_t toMsg = {ToThreadMessageExit, 0};
--- a/ADApp/pluginSrc/NDPluginDriver.h
+++ b/ADApp/pluginSrc/NDPluginDriver.h
@@ -70,6 +70,7 @@ public:
                           size_t *nActual);
     virtual asynStatus readInt32Array(asynUser *pasynUser, epicsInt32 *value,
                                         size_t nElements, size_t *nIn);
+    virtual void shutdownPortDriver();
 
     /* These are the methods that are new to this class */
     virtual void driverCallback(asynUser *pasynUser, void *genericPointer);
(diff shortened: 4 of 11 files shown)


henriquesimoes:
> To maintain compatibility with older versions of asyn, use of new symbols is
> gated by #ifdef ASYN_DESTRUCTIBLE where needed. It needs to be pointed out
> that virtual void shutdownPortDriver() does not need to be gated because
> there's no issue with adding it even with older versions of asyn as long as you
> don't use the C++11 override keyword. Only the call into the base
> asynPortDriver::shutdownPortDriver() needs to be gated.

IMHO, this information should go in the commit message for d321ecf1 (Add port shutdown to asynNDArrayDriver, 2026-02-24) and

> The recipe for making a driver destructible is as follows:
> [...]

should go in the [driver guidelines page](https://areadetector.github.io/areaDetector/ADCore/guidelines.html) (in a dedicated section?).

In general, I'm in favor of documenting any relevant aspects in each commit its message (rather than **only** in the PR description). This was proposed [in a recent collaboration meeting](https://github.com/areaDetector/collaboration/blob/master/2026-02-05.md#meta-issues), so it not a widespread practice in areaDetector repositories (yet). Do you  mind rebasing your patches to add the motivation, attention points and any other relevant "why?" answers not clearly available in the patch content itself?

exzombie:
Done! What's the approach for updating release notes? Does that belong into the PR as well?

henriquesimoes:
> Done!

Thanks!

> What's the approach for updating release notes? Does that belong into the PR as well?

AFAIK, there isn't a documented procedure for this, but for most past releases @MarkRivers wrote down the notes afterwards in batches (for instance, 7ff36dc78f2f20f706e4efc784db9cfa31488899 and 516e91f1212419ee9955dbc877224abcae898c5d).

I find this process a bit sub-optimal, because we usually have a better context of the changes when they are proposed rather than when a release happens. It is also more tiresome to revisit all commits to choose which changes are relevant to be mentioned in the release notes than to decide that during the PR review.

In my opinion, we should move towards including the release notes updates in the corresponding PRs whenever possible. I'm not the one making releases, though.

prjemian:
GitHub web has a checkbox to collate concise release notes when making a new release.  Structure titles of pull requests and issues to provide content.  See https://github.com/bluesky/hklpy2/releases for examples.

prjemian:
This repo does not use GH Web for [releases](https://github.com/areaDetector/ADCore/releases).  Only git [tags](https://github.com/areaDetector/ADCore/tags):
image

MarkRivers:
> In my opinion, we should move towards including the release notes updates in the corresponding PRs whenever possible.

I agree.  And if the PR makes changes in the behavior or user interface then the documentation also should be updated in the PR as well.

kgofron:
## ADTimePix3

I compiled with destructible-drivers branch, but unfortunatly exit after acquisition still results in segmentation fault.

```
epics> exit
PrvHst TCP disconnected
./st.cmd: line 5: 2228300 Segmentation fault      ../../bin/linux-x86_64/tpx3App st_base.cmd
```
Perhaps I missed something.

* git remote 
  * https://github.com/exzombie/ADCore
```
kg1@lap133454:/epics/support2/areaDetector/ADCore/iocBoot$ git branch -a
* destructible-drivers
```
* 
```
kg1@lap133454:/epics/support2/areaDetector/ADCore/iocBoot$ git log
commit 91d002c3afa482c6da827f3b11b95706d1d5028b (HEAD -> destructible-drivers, origin/destructible-drivers)
Author: Jure Varlec <jure.varlec@cosylab.com>
Date:   Fri Feb 27 09:25:06 2026 +0000

    Add a release note about port shutdown
```

* asyn/asyn/asynDriver/asynDriver.h 
```
/* Version number names similar to those provide by base
 * These macros are always numeric */
#define ASYN_VERSION       4
#define ASYN_REVISION     45
#define ASYN_MODIFICATION  0
```

### Summary

The destructible-drivers branch (ADCore PR 572) fixes shutdown order (asyn calls shutdownPortDriver() and then deletes the driver), but it does not include the pool-safety fix from ADCore PR 570. So the crash you see is still the same: after the driver (and its NDArrayPool) are destroyed, pvAccess (PVA) can later call NDArray::release() on arrays that belonged to that pool → use-after-free → SIGSEGV.

So ADTimePix3 needs both:

* Destructible drivers (PR 572) 
* Destroyed-pool safety (PR 570) – this is what prevents the SIGSEGV when PVA (or anything else) calls release() after the pool is gone.

MarkRivers:
@kgofron can you post the stack trace when it crashes on exit?

kgofron:
## Stack trace:

“SIGSEGV on exit with ADCore master. Backtrace shows crash in NDArrayPool::release (ADCore) called from PVA teardown (freeNDArray → NDArray::release) after the driver and its pool are already destroyed. Full bt full and info sharedlibrary attached.”

Attached: The full bt full output (and optionally info sharedlibrary) from your tpx3_SIGSEGV.md file, or the relevant frames (#0–#2, #37, #49, #65–#66, #71–#72, #79–#81) to see the PVA → NDArray release → pool release path and the exit-handler order.

## GDB analysis – SIGSEGV on exit (ADCore master)
**Build**: ADTimePix3 IOC, run with ADCore current master (no PR 570, no destructible PR 572 in this run).
**Repro**: Acquisition run, then exit in the IOC shell.
**Crash**: SIGSEGV in NDArrayPool::release() at NDArrayPool.cpp:373 (onReleaseArray(pArray)).

### Where it crashes
**#0** – NDArrayPool::release(this=0x555556ccc870, pArray=0x7fff3c001a00) at NDArrayPool.cpp:373 So the fault is inside the pool’s release() (use-after-free on the pool or its internals).

### Call chain (who called into the pool)
**#81–79** – User types exit → epicsExit(0) → C library runs atexit handlers.
**#72–71** – One of those handlers destroys the pvAccess ServerContext (static/atexit cleanup).
**#70–65** – PVA tears down transports and channels, then ServerMonitorRequesterImpl::destroy.
**#57–48** – MonitorElementQueue and its MonitorElements are destroyed.
**#37–36** – MonitorElement destructor runs (PVA held a structure that contained array data).
**#30–11** – Destructors for PVStructure → PVValueArray<unsigned short> → shared_vector; the shared storage uses a custom deleter.
**#2** – freeNDArray<unsigned short>::operator() (ADCore ntndArrayConverter.cpp:44) – deleter for the PVA copy of the NDArray data.
**#1** – NDArray::release() – called from that deleter.
**#0** – NDArrayPool::release() – crash.

So: PVA is shutting down (atexit), destroying MonitorElements that still hold NDArray-backed data. Their deleter calls NDArray::release(), which calls NDArrayPool::release() on a pool that no longer exists.

### Why the pool is gone
Shutdown order is:

* Earlier in exit, the ADTimePix driver is destroyed (either by epicsAtExit(exitCallbackC) or by asyn destructible teardown). That runs ~asynNDArrayDriver, which deletes the NDArrayPool (pNDArrayPoolPvt_). So the pool at 0x555556ccc870 is freed.
* Later, the pvAccess ServerContext is destroyed (another atexit). Its MonitorElements still hold NDArrays (or views) that point at that same pool. When those are freed, freeNDArray → NDArray::release() → NDArrayPool::release() runs on the already deleted pool → use-after-free → SIGSEGV.
So the trace matches the “destroyed pool” scenario: driver (and pool) destroyed first, PVA tears down later and calls release() on that pool.

### Conclusion
**Root cause:** Use-after-free in ADCore: NDArray::release() is called from PVA’s deleter after the driver’s NDArrayPool has already been destroyed. The crash is in NDArrayPool::release (ADCore), not in ADTimePix3.
**Fix:** This is exactly what ADCore PR 570 (destroyed-pool safety) addresses: make NDArray::release() (and optionally the pool) safe when the pool has already been marked destroyed / unregistered, so PVA’s late release no-ops instead of touching freed pool memory.

[tpx3_SIGSEGV.md](https://github.com/user-attachments/files/25981453/tpx3_SIGSEGV.md)
[tpx3_SIGSEGV_frames.md](https://github.com/user-attachments/files/25981456/tpx3_SIGSEGV_frames.md)

See reasons why it must be resolved in ADCore, and can not be resolved in ADTimePix3 driver: https://github.com/areaDetector/ADCore/pull/570

ericonr (approved):
@MarkRivers this looks good to merge!

PR #576, opened 2026-04-03
Add support for lz4hdf5 codec

MarkRivers:
This adds support for the lz4hdf5 codec to NDPluginCodec and NDFileHDF5.   The codec itself is in ADSupport. There are also changes to ADEiger to use this new codec for Stream2 LZ4 compression, and to ADViewers to display arrays with this codec in ImageJ.

### DIFF
--- a/ADApp/Db/NDCodec.template
+++ b/ADApp/Db/NDCodec.template
@@ -197,6 +197,23 @@ record(longin, "$(P)$(R)BloscNumThreads_RBV")
     field(SCAN, "I/O Intr")
 }
 
+record(longout, "$(P)$(R)LZ4HDF5BlockSize")
+{
+    field(PINI, "YES")
+    field(DTYP, "asynInt32")
+    field(OUT,  "@asyn($(PORT),$(ADDR=0),$(TIMEOUT=1))LZ4HDF5_BLOCKSIZE")
+    field(VAL,  "0")
+    field(DRVL, "0")
+    info(autosaveFields, "VAL")
+}
+
+record(longin, "$(P)$(R)LZ4HDF5BlockSize_RBV")
+{
+    field(DTYP, "asynInt32")
+    field(INP,  "@asyn($(PORT),$(ADDR=0),$(TIMEOUT=1))LZ4HDF5_BLOCKSIZE")
+    field(SCAN, "I/O Intr")
+}
+
 record(mbbi, "$(P)$(R)CodecStatus")
 {
     field(DTYP, "asynInt32")
--- a/ADApp/Db/NDCodec.template
+++ b/ADApp/Db/NDCodec.template
@@ -42,8 +42,10 @@ record(mbbo, "$(P)$(R)Compressor")
     field(TWVL, "2")
     field(THST, "LZ4")
     field(THVL, "3")
-    field(FRST, "BSLZ4")
+    field(FRST, "LZ4HDF5")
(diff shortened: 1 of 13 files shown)


ADApp/pluginSrc/NDPluginCodec.cpp
```diff
@@ -531,6 +532,76 @@ NDArray *decompressLZ4(NDArray *input, NDCodecStatus_t *status, char *errorMessa
...
+{
+    if (!input->codec.empty()) {
+        sprintf(errorMessage, "Array is already compressed");
+        *status = NDCODEC_WARNING;
+        return NULL;
+    }
+
+    NDArrayInfo_t info;
+    input->getInfo(&info);
+    int outputSize = LZ4_compressBound((int)info.totalBytes);
```
ericonr: Should we add our own `compress_bound_lz4hdf5` function in ADSupport? The checking performed by `compress_lz4hdf5` is reasonably complex and I don't think it should be repeated across the code base.

As is, I believe this size would cause `compress_lz4hdf5` below to error out.

ericonr (commented):
Isn't this missing the changes to https://github.com/areaDetector/ADCore/blob/6c53844e72d467b86a03fe712ea1f69210e2cae0/ADApp/pluginSrc/NDFileHDF5Dataset.cpp#L299 to properly pass the data through to the HDF5 file without inserting any metadata?

ADApp/Db/NDCodec.template
```diff
@@ -42,8 +42,10 @@ record(mbbo, "$(P)$(R)Compressor")
     field(TWVL, "2")
     field(THST, "LZ4")
     field(THVL, "3")
-    field(FRST, "BSLZ4")
+    field(FRST, "LZ4HDF5")
     field(FRVL, "4")
+    field(FVST, "BSLZ4")
+    field(FVVL, "5")
```
ericonr, line 48: I think this change in order would be nice to at least document in release notes, since I believe it could break someone's autosave.

MarkRivers:
> Isn't this missing the changes to
 ADCore/ADApp/pluginSrc/NDFileHDF5Dataset.cpp to properly pass the data through to the HDF5 file without inserting any metadata?

No.  The codecName for lz4hdf5 is NDCODEC_LZ4HDF5, not NDCODEC_LZ4, and there is no special treatment required for that codec in NDFileHDF5Dataset.cpp.  I have tested this.  Here is the output of h5dump --header -p on a file saving NDArrays compressed with lz4hdf5 with a block size of 65000.  This was generated with ADSimDetector and NDPluginCodec.
```
            DATATYPE  H5T_STD_U8LE
            DATASPACE  SIMPLE { ( 100, 1024, 1024 ) / ( 100, 1024, 1024 ) }
            STORAGE_LAYOUT {
               CHUNKED ( 1, 1024, 1024 )
               SIZE 1545960 (67.827:1 COMPRESSION)
            }
            FILTERS {
               USER_DEFINED_FILTER {
                  FILTER_ID 32004
                  COMMENT HDF5 lz4 filter; see http://www.hdfgroup.org/services/contributions.html
                  PARAMS { 0 0 }
               }
            }
```

ADApp/Db/NDCodec.template
MarkRivers: I have added this to RELEASE.md

PR #581, opened 2026-04-27
Re-enable coveralls code coverage by adding coveralls action to github actions workflow

jwlodek:
- **Remove unused travis ci configuration file**
- **Update github actions CI configuration to re-enable coveralls code coverage metrics**


jwlodek:
This appears to have worked:

image

jwlodek:
I've moved the coverage check to the matrix job that builds against adsupport, to guarantee that most dependency libraries are available - I noticed that on the default ubuntu-latest/os/default the lz4 and bslz4 tests in the Codec tests did not get covered, presumably since there isn't a bitshuffle OS package.

jwlodek:
Hmm, now it picked up those branches in the Codec tests, but I don't seem to get `.gcda` files for all of the files I'd expect (e.g. `ADDriver.cpp` is not listed). The previous run listed 59 files, the new one only 37. Need to investigate further.

jwlodek:
Ok, figured out that this was caused by difference in gcda file generation between static/shared builds. I added a linux/adsupport/shared entry to the matrix, and have used that for the coverage. This should now be good to go.

MarkRivers (commented):
This looks good to me.  I am not an expert in this.

ericonr:
@jwlodek I've rewritten the commit history in this branch: https://github.com/ericonr/ADCore/commits/fix-coveralls/

What do you think about that organization? Would it be okay if we merged it instead?

jwlodek:
No strong opinions one way or another.

ericonr:
I'm closing this in favor of #587 , then.

PR #583, opened 2026-04-28
Make NdPluginAttribute compression aware, add unit tests for NDPluginAttribute

### LINKED ISSUE #582: NDPluginAttribute isn't compression aware
I don't think it does anything dealing with actual NDArray data, so it should be simple to make it safely compression aware.

I can do this when I have time, simply opening an issue to track it.

jwlodek:
Resolves #582. Was as simple as adding the boolean flag to the end of the superclass constructor call. Also added unit tests for the plugin that test with compressed and uncompressed arrays.

### DIFF
--- a/ADApp/pluginTests/test_NDPluginAttribute.cpp
+++ b/ADApp/pluginTests/test_NDPluginAttribute.cpp
@@ -17,6 +17,7 @@
 #include <asynDriver.h>
 #include <Codec.h>
 #include <NDPluginAttribute.h>
+#include <NDPluginCodec.h>
 
 #include "testingutilities.h"
 #include "AttributePluginWrapper.h"
@@ -55,7 +56,7 @@ struct AttributeTestFixture
         delete dummy_driver;
     }
 
-    NDArray *createTestArray()
+    NDArray *createTestArray(bool compress = false)
     {
         size_t dims[2] = {TEST_XSIZE, TEST_YSIZE};
         NDArray *arr = arrayPool->alloc(2, dims, NDUInt16, 0, NULL);
@@ -63,6 +64,17 @@ struct AttributeTestFixture
         for (int i = 0; i < TEST_NELEMENTS; i++) {
             pData[i] = (epicsUInt16)(i % 256);
         }
+#ifdef HAVE_ZLIB
+        if (compress) {
+            NDCodecStatus_t status;
+            char errorMessage[256] = "";
+            NDArray *compressed = compressZlib(arr, 6, &status, errorMessage);
+            arr->release();
+            return compressed;
+        }
+#else
+        (void)compress;
+#endif
         return arr;
     }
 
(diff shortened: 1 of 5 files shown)


ADApp/pluginTests/test_NDPluginAttribute.cpp
```diff
@@ -0,0 +1,147 @@
...
+        attr->unlock();
+    }
+};
+
+BOOST_FIXTURE_TEST_SUITE(AttributePluginTests, AttributeTestFixture)
+
+/* Test that NDPluginAttribute reads attributes from an uncompressed array */
+BOOST_AUTO_TEST_CASE(test_attribute_uncompressed)
+{
+    NDArray *input = createTestArray();
```
gustavosr8, line 94: Since you are using in the other tests, missing `BOOST_REQUIRE(input != NULL)`

ADApp/pluginTests/test_NDPluginAttribute.cpp
```diff
@@ -0,0 +1,147 @@
...
+    double testVal = 42.5;
+    input->pAttributeList->add("TestAttr", "", NDAttrFloat64, &testVal);
+
+    attr->write(NDPluginAttributeAttrNameString, std::string("TestAttr"), 0);
+
+    processArray(input);
+
+    double readVal = attr->readDouble(NDPluginAttributeValString, 0);
+    BOOST_CHECK_CLOSE(readVal, 42.5, 0.001);
```
gustavosr8, line 103:
```suggestion
    BOOST_CHECK_CLOSE(readVal, testVal, 0.001);
```

ADApp/pluginTests/test_NDPluginAttribute.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+    double testVal = 99.0;
+    input->pAttributeList->add("CompAttr", "", NDAttrFloat64, &testVal);
+
+    attr->write(NDPluginAttributeAttrNameString, std::string("CompAttr"), 0);
+
+    processArray(input);
+
+    double readVal = attr->readDouble(NDPluginAttributeValString, 0);
+    BOOST_CHECK_CLOSE(readVal, 99.0, 0.001);
```
gustavosr8, line 124:
```suggestion
    BOOST_CHECK_CLOSE(readVal, testVal, 0.001);
```

ADApp/pluginTests/test_NDPluginAttribute.cpp
```diff
@@ -0,0 +1,147 @@
...
+    NDArray *input = createTestArray(true);
+    BOOST_REQUIRE(input != NULL);
+    input->uniqueId = 12345;
+
+    attr->write(NDPluginAttributeAttrNameString, std::string("NDArrayUniqueId"), 0);
+
+    processArray(input);
+
+    double readVal = attr->readDouble(NDPluginAttributeValString, 0);
+    BOOST_CHECK_CLOSE(readVal, 12345.0, 0.001);
```
gustavosr8, line 141: Why not `BOOST_CHECK_EQUAL`?

jwlodek: I think `CHECK_EQUAL` should be OK. I usually use `CHECK_CLOSE` or the framework equivalent for floats due to precision issues, but since we're not doing any calculations or the like here it should match exactly.

ADApp/pluginTests/test_NDPluginAttribute.cpp
```diff
@@ -0,0 +1,139 @@
...
+    NDArray *createTestArray()
+    {
+        size_t dims[2] = {TEST_XSIZE, TEST_YSIZE};
+        NDArray *arr = arrayPool->alloc(2, dims, NDUInt16, 0, NULL);
+        epicsUInt16 *pData = (epicsUInt16 *)arr->pData;
+        for (int i = 0; i < TEST_NELEMENTS; i++) {
+            pData[i] = (epicsUInt16)(i % 256);
+        }
+        return arr;
+    }
```
gustavosr8, line 67: Why not using [`fillNDArraysFromPool`](https://github.com/areaDetector/ADCore/blob/3f74bad281f91f819c5b9eef45e8785ebabe0600/ADApp/pluginTests/testingutilities.cpp#L45-L63)?

PR #586, opened 2026-05-20
Fix NDArrayInfo setting in ColorConvert

gustavosr8:
This PR was motivated after seeing different values of compressed and uncompressed size after performing a mono to RGB1 conversion.

Tested with ADSim converting its mono image to RGB1, 2 and 3. Inspecting the data end metadata with PVA plugin. The dimensions and the data size (compressed and uncompressed) are correct.

### DIFF
--- a/ADApp/pluginTests/test_NDPluginColorConvert.cpp
+++ b/ADApp/pluginTests/test_NDPluginColorConvert.cpp
@@ -1,0 +1,145 @@
+#include <boost/test/tools/old/interface.hpp>
+#include <string.h>
+#include <stdlib.h>
+
+#include "boost/test/unit_test.hpp"
+
+#include <NDPluginDriver.h>
+#include <NDArray.h>
+#include <asynDriver.h>
+#include <NDPluginColorConvert.h>
+
+#include "testingutilities.h"
+#include "ColorConvertPluginWrapper.h"
+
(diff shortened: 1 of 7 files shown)


ericonr (approved):

jwlodek:
Could we rebase this on master now that CI is fixed? Also, it would be good if we could have a test with a compressed array as an input (like in the Attribute plugin tests).

ADApp/pluginTests/test_NDPluginColorConvert.cpp
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    void processArray(NDArray *pArray)
+    {
+        cc->lock();
+        cc->processCallbacks(pArray);
+        cc->unlock();
+    }
+
+    /* set array metadata and convert it to desider color mode*/
```
jwlodek:
```suggestion
    /* set array metadata and convert it to desired color mode*/
```

gustavosr8:
> Also, it would be good if we could have a test with a compressed array as an input (like in the Attribute plugin tests).

@jwlodek do you think this still relevant since the plugin isn't compression aware?

jwlodek:
Ah OK, at first glance I thought it supported compressed data since the PR was adjusting compressedSize variable behavior. In that case I think this is good to merge.

Issue #604, opened 2026-09-02
CB plugin does not correctly clear `Trigger` busy record on final soft trigger


jwlodek:
Currently, if you soft trigger the CB plugin, on the last preset trigger the busy record does not clear.  This is because the asyn param is not set back to zero in the final loop iteration. As a result, a caput to the soft trigger PV with put completion will timeout.

PR #605, opened 2026-09-02
Set Soft trigger back to zero on last loop iteration in CB plugin, to clear busy record. Add unit test to cover this case

jwlodek:


jwlodek:
Resolves #604

